PR #37, opened 2024-04-02
Include README

mirhamasala:


CharlyMartin:
@mirhamasala The README looks good!

I've updated the shared prettier config to match the current format of the code and added a pointer to it for VSCode users (both of us, I believe).

Here are some suggestions/questions below. Let me know what you think :)

### Prerequisites
Should we add the prerequisites under ["engines"](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines) in the package.json?

On the same note, it seems like [Node v14](https://nodejs.org/en/about/previous-releases) is no longer supported / actively maintained. If we do decide to enforce node and npm versions, it would be best to pick more recent versions.

### Linting
Should we set up a pre-commit hook to make sure linting rules are respected? Or is it overkill?
A tool like [Husky](https://github.com/typicode/husky) does the job for instance.

Maybe it's redundant given the existing CI check on PRs.

### Pre-build
Should the pre-build step always be run before the build? If so, could we include it in the build step?
```bash
 "build": "node scripts/convertYamlToJson.js && next build",
```

### Updating the sitemap
It seems like the `sitemap()` function already includes dynamic content?
If not, can you give me an example of content that isn't automatically included in the sitemap?

mirhamasala:
@CharlyMartin - Looking good, Charly. I only removed the `trailingComma` rule - It was a leftover and shouldn't be there. Let's get this PR merged, and pick up some of your suggestions in a separate PR. 🙏🏼

---

- Let's talk Node versions once we update Next.js Runtime
- Yes to Husky
- About `prebuild` - Not sure if I understand the question. The `prebuild` script is designed to automatically run before the build script when you execute `npm run build`. This behavior is built into npm as a convention, where scripts named with a `pre` prefix run before their corresponding script without the prefix, and those with a `post` prefix run afterward.
- About the sitemap, any collection pages have to be added manually at this point, like so:

```javascript
  const blogPosts = getBlogPostsData()
  const dynamicBlogRoutes = generateDynamicRoutes(blogPosts, PATHS.BLOG.path)
```

While all other pages are automatically included through the `PATHS` object, like so:

```javascript
  const staticRoutes = Object.values(PATHS).map((path) => ({
    url: `${BASE_URL}${path.path}`,
    lastModified: new Date(),
  }))
```

CharlyMartin:
Thank you @mirhamasala for the info 👌 

I didn't know that `prebuild` automatically ran before `build` as a convention, really cool :) Scratch my comment about it then!

PR #43, opened 2024-04-05
Blog Post Pagination

CharlyMartin:
This PR adds pagination to the blog page index.

It doesn't match the Figma designs perfectly yet, but I want to get feedback on how the feature works before polishing the UI.

- [x]  Display 20 posts at a time while keeping the other posts accessible on the page.
- [x] There will be no progressive loading, all blog posts will be on the page for accessibility and SEO purposes
- [x]  Keep the current page in sync with the URL so that it’s shareable / persisted across reloads
- [x]  Keep the search query in sync with the URL so that it’s shareable / persisted across reloads
- [x]  Fix the blog card’s height to avoid layout shifts

https://www.notion.so/filecoin/FF-V4-Pagination-09ddd17d7f2b429f89982c052f74a923?pvs=4

### DIFF
--- a/src/app/_components/BlogPostCard.tsx
+++ b/src/app/_components/BlogPostCard.tsx
@@ -1,0 +1,46 @@
+import Image from 'next/image'
+
+import clsx from 'clsx'
+
+import { Heading } from '@/components/Heading'
+import { TextLink } from '@/components/TextLink'
+
+import type { BlogPostData } from '@/types/blogPostTypes'
+
+import { formatDate } from '@/utils/formatDate'
+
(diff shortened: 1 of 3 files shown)


CharlyMartin:
The blog page on the deploy preview is panicking; it seems to be stuck in an infinite loop or re-renders.

Will investigate after lunch 🧑‍🍳

CharlyMartin:
It's fixed and ready for review :)

CharlyMartin:
Feedback

- [ ] Use `Card` and `CardLayout` for the list of blogs

PR #53, opened 2024-04-10
Improve accessibility of Socials tap target

mirhamasala:
To comply with accessibility guidelines and enhance usability on touch devices, the touch targets for the social icons have been increased. This is achieved by applying `p-2.5` padding within the anchor tags, effectively enlarging the clickable area without altering the visual size of the icons.

- [Source](https://web.dev/articles/accessible-tap-targets)
- https://developer.chrome.com/docs/lighthouse/seo/tap-targets

## Note: Need to discuss design implications with Filipa


mirhamasala:
Closing in favor of #52

PR #55, opened 2024-04-11
Style blog posts section

mirhamasala:
This PR styles the Blog Posts list section using the `Card` component and includes the`Meta` and `NoResultsMessage` components ~ per [Figma Blog Design](https://www.figma.com/file/0pBfQs5tKI6bx5ypPIfVRA/FIL.org-V4-Prototypes?type=design&node-id=547-11816&mode=design&t=gAuzm182CmWC3SQb-0)

Screenshot 2024-04-11 at 10 43 24

Screenshot 2024-04-11 at 10 43 53

### DIFF
--- a/src/app/_components/Card.tsx
+++ b/src/app/_components/Card.tsx
@@ -1,3 +1,5 @@
+import { Fragment } from 'react'
+
 import Image from 'next/image'
 
 import { BookOpen } from '@phosphor-icons/react/dist/ssr'
@@ -8,6 +10,7 @@ import { Heading } from '@/components/Heading'
 
 export type CardProps = {
   title: string | React.ReactNode
+  metaData?: Array<string | null | undefined>
   description?: string
   cta?: {
     href: string
@@ -21,6 +24,7 @@ export type CardProps = {
   }
   borderColor?: 'brand-300' | 'brand-500' | 'brand-600'
   textIsClamped?: boolean
+
   as?: React.ElementType
   children?: React.ReactNode
 }
@@ -40,6 +44,7 @@ const imageSizes = {
 
 export function Card({
   title,
+  metaData,
   description,
   cta,
   entryType = 'blogPost',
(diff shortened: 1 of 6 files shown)


mirhamasala:
@CharlyMartin Interested in whether you think we should handle this part differently - Would prefer not to use a `className` prop.

```
 <div
    key={post.slug}
    className={determineVisibilityClass(index)}
  >
```

CharlyMartin (commented):
Looks great @mirhamasala 🤘

It can be good to merge as is, depending on what you think about these points.

### Accessibility of blog cards
In the current `Card` implementation, it is NOT possible to select text inside a given card - when copying/pasting for instance. Because the link tag is `absolute inset-0`, it blocks the elements underneath and might even trigger navigation.

I feel like we should make this possible. What do you think?

In the same vein, if we wanted to make the category badge in each card clickable for filtering, it wouldn't work.

### Card Component
You already mentioned this point. The `Card` component is starting to stretch itself to do too many things.

We could break it down into subcomponents and then create wrappers for each entity: `Blog`, `Event`, `People`, etc. It would probably be for a different PR.

It's how the Tailwind people do it in the personal website template we both use. Here's how, for reference:

```tsx
// Card.tsx
export function Card<T extends React.ElementType = 'div'>({ as, className, children}) {
  let Component = as ?? 'div'

  return (
    <Component
      className={clsx(className, 'group relative flex flex-col items-start')}
    >
      {children}
    </Component>
  )
}

Card.Link = function CardLink({ ... )}
Card.Title = function CardTitle({ ... )}
Card.Description = function CardDescription({ ... )}
Card.Cta = function CardCta({ ... )}

// ...

```
```tsx
// Article.tsx
import { Card } from "Card"

function Article({ article }) {
  return (
    <Card className="md:col-span-3">
      <Card.Title href={`/articles/${article.slug}`}>
        {article.title}
      </Card.Title>
      <Card.Eyebrow
        as="time"
        dateTime={article.date}
        className="md:hidden"
        decorate
      >
        {formatDate(article.date)}
      </Card.Eyebrow>
      <Card.Description>{article.description}</Card.Description>
      <Card.Cta>Read article</Card.Cta>
    </Card>
    <Card.Eyebrow
      as="time"
      dateTime={article.date}
      className="mt-1 hidden md:block"
    >
      {formatDate(article.date)}
    </Card.Eyebrow>
  )
}
```

CharlyMartin:
About your https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/55#issuecomment-2049324120 - I don't mind using a `className` prop, as long as it is used as a last resort.

We can even use `twMerge` to make sure the classes coming from outside the component do not override the existing ones.

mirhamasala:
@CharlyMartin All great points!

>Accessibility of blog cards

Yes. Oversight from my side - Forgot to add some `pointer-events` classes - Fixed.

>Card Component

Totally. Waiting for the designs to be more stable and to have a few more pages developed so we have a better overview of what we need.

>About your https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/55#issuecomment-2049324120 - I don't mind using a className prop, as long as it is used as a last resort.

Ugh. That "last resort" worries me. Let's hold off for now. This made me rework the logic so it would be needed. Let me know what you think.

CharlyMartin (approved):

PR #60, opened 2024-04-15
Add ScreenSizeIndicator component

CharlyMartin:
Add `ScreenSizeIndicator` component to see the current breakpoint in development mode.

image


src/app/layout.tsx
```diff
@@ -43,6 +44,7 @@ export default function RootLayout({
         </main>
 
         <Footer />
+        <ScreenSizeIndicator />
```
mirhamasala, line 47: Could we perform this check earlier - before component definition?

```suggestion
                {process.env.NODE_ENV === 'development' && <ScreenSizeIndicator />}
```

src/app/_components/ScreenSizeIndicator.tsx
```diff
@@ -0,0 +1,14 @@
+export function ScreenSizeIndicator() {
```
mirhamasala, line 1: What about `ResponsiveDebugger`. I feel it something like that might indicate it's used for Development purposes.

I think we can also prefix this component with an `_` - What do you think?

CharlyMartin: Yes, good points.

I renamed the component `BreakpointDebbuger` and changed its path to `_dev_BreakpointDebbuger.tsx`, so that it's clear that it's for dev only.

Let me know what you think :)

src/app/_components/_dev_BreakpointDebbuger.tsx
```diff
@@ -0,0 +1,12 @@
+export function BreakpointDebbuger() {
```
mirhamasala, line 1:
```suggestion
export function BreakpointDebugger() {
```

mirhamasala:
@CharlyMartin Minor spelling mistake left - other than that - ready to be merged. 🚀

src/app/_components/_dev_BreakpointDebbuger.tsx
mirhamasala: @CharlyMartin Minor spelling mistake left - other than that - ready to be merged. 🚀

PR #61, opened 2024-04-15
Change LinkedIn Link placement

mirhamasala:


mirhamasala:
@filipagr Let me know what you think [>>](https://deploy-preview-61--filecoin-foundation.netlify.app/about)

Screenshot 2024-04-15 at 10 20 20
Screenshot 2024-04-15 at 10 20 07

filipagr:
I like it very much! I understand and agree with Julia's perspective of the LinkedIn not being the most important thing on the card, but visually, as we now have it at the bottom, I prefer it with text. I feel it's more balanced.

barbaraperic:
Looks good 👌 I'm still getting familiar with the structure you follow but one thing I was wondering if the `reportsData`, `advisoryData` and similar `data` arrays could be extracted in its own file since they take out some space.

barbaraperic (approved):

mirhamasala:
Yes - @barbaraperic - that's a great point. We can extract it into the `_data` folder. I'll work on it. 🙏🏼

PR #59, opened 2024-04-15
REF: Badge component and optional prop in PageHeader

barbaraperic:
## Changes

- `PageHeader` component takes an optional prop `isFeatured`
- `isFeatured` is true for `PageHeader` component on `events` and `blog` pages

PR #64, opened 2024-04-16
CR #59 - Badge component and optional prop in PageHeader

mirhamasala:
CR #59 - Badge component and optional prop in PageHeader


mirhamasala:
FYI @barbaraperic - My CR didn't end up on remote so here it is.

PR #71, opened 2024-04-18
STYLE: grants and opportunities section

barbaraperic:
## Changes

- reused `CardLayout` and `Card` component in the `grants` page
- created new `_icons` folder to store the `svg`s for the cards - let me know if you prefer to have it all in the `public`. I made them as components so they can be modified easily. 

## Note

- I've noticed there's a margin bottom class in the `Card` which adds extra spacing below and doesn't match the design - let me know if you want to adjust that as well. It might need a bit of extra refactoring.

### Current
Screenshot 2024-04-18 at 08 33 17

### Figma

Screenshot 2024-04-18 at 08 33 54

## Visuals

Screenshot 2024-04-18 at 08 27 37
Screenshot 2024-04-18 at 08 27 52


mirhamasala:
@barbaraperic We're using `@phosphor-icons` as the React icon library. Please check the `Explore` section on the homepage for implementation details - `/src/app/page.tsx`. And yes, we need to fix the `margin-bottom` issue but we can pick this up in another PR as design is currently working on restyling the `Card` component. 🙏🏼

src/app/grants/page.tsx
```diff
@@ -23,6 +29,27 @@ const grantsPageBaseData = generateWebPageStructuredData({
...
+    Icon: MicrograntsIconComponent,
+  },
+  {
+    title: 'Requests for Proposal',
+    description:
+      'Request for Proposals (RFP) grants have clearly scoped deliverables, milestones, and funding limits. Refer to GitHub for any ongoing RFPs.',
+    Icon: RequestForProposalIconComponent,
+  },
+]
+
```
mirhamasala, line 52: Happy to move this into `/_data` folder within` /grants` as per your own suggestion for the About page.

src/app/grants/page.tsx
```diff
@@ -33,55 +60,36 @@ export default function Grants() {
...
+            return (
+              <Card
+                key={title}
+                title={
+                  <div className="flex items-center gap-3">
+                    <Icon className="h-6 w-6"></Icon>
+                    <Heading
+                      tag="h3"
+                      variant="lg"
+                      className="line-clamp-2 text-ellipsis"
```
mirhamasala, line 81: `Card` doesn't accept a `className` prop so this has no effect. Also, note that there are some components that accept a `className` prop but you should only use it if absolutely necessary. In fact, thinking of removing those in the near future as well. The `Card` component does have a `textIsClamped` option - but, in this case, I don't think we want any of the text to be clamped.

mirhamasala: Apologies - Oversight here. Wrong component. Scratch previous comment. In this case, though, we don't want the text to be clamped. That's useful, for example, for long blog post headings. In this case, we want all this copy to be visible - since the user can't open the card anywhere else either.

mirhamasala:
@barbaraperic Instead of reusing `Card` - Can you create a separate component for this use case? We're trying to map the different cards we have, and this would be useful. We can refactor late. (Also, have a look at `HomeExploreSectionCard` - it might be useful - and we might want to abstract `HeadingWithIcon` into its own component.)

src/app/_components/GrantsSectionCard.tsx
```diff
@@ -0,0 +1,21 @@
+import { HeadingWithIcon } from './HeadingWithIcon'
```
mirhamasala, line 1:
```suggestion
import { HeadingWithIcon } from '@/components/HeadingWithIcon'
```

src/app/_components/HeadingWithIcon.tsx
```diff
@@ -0,0 +1,20 @@
+import { Heading } from '@/components/Heading'
+
+export function HeadingWithIcon({
+  icon,
+  children,
+}: {
+  icon: React.ReactNode
+  children: React.ReactNode
```
mirhamasala, line 8:
```suggestion
  children: string
```

src/app/grants/page.tsx
```diff
@@ -33,55 +38,29 @@ export default function Grants() {
...
+      >
+        <CardLayout>
+          {grantsAndOpportunitiesData.map((card) => {
+            const { title, description, Icon } = card
+
+            return (
+              <GrantsSectionCard
+                key={title}
+                heading={{
+                  icon: <Icon className="h-6 w-6 text-brand-300"></Icon>,
```
mirhamasala, line 54: Styling can happen in the component.

src/app/_components/HeadingWithIcon.tsx
mirhamasala: Don't hesitate to change what was previously written if you think it should be something else.

src/app/_components/HeadingWithIcon.tsx
```diff
@@ -0,0 +1,20 @@
...
+}) {
+  return (
+    <div className="inline-flex items-center gap-3">
+      <span className="text-brand-300" aria-hidden={true}>
+        {icon}
+      </span>
+      <Heading tag="h3" variant="lg">
```
mirhamasala, line 15: At the moment, this would limit us to `h3` and `lg` at all times. This was arguably possible within `HomeExploreSectionCard` but now that we're abstracting it, we have a chance to change it.

src/app/_components/GrantsSectionCard.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+type GrantsSectionCardProps = {
+  heading: {
+    icon: React.ReactNode
+    title: string
+  }
+  children: React.ReactNode
```
mirhamasala, line 8:
```suggestion
  children: string
```

src/app/grants/page.tsx
```diff
@@ -11,9 +11,14 @@ import { generateWebPageStructuredData } from '@/utils/structuredData'
 
 import { attributes } from '@/content/pages/grants.md'
 
+import { CardLayout } from '@/_components/CardLayout'
+import { GrantsSectionCard } from '@/_components/GrantsSectionCard'
```
mirhamasala, line 15:
```suggestion
import { GrantsSectionCard } from '@/components/GrantsSectionCard'
```

src/app/_components/HeadingWithIcon.tsx
barbaraperic: Since it's being used only in cards for now, would it be better to limit it for consistency? Or we could give it, for example, 3 sizes as have that as a prop? It would be more explicit.

barbaraperic: Also Heading component would need modifications as well as it has required props.

src/app/grants/page.tsx
barbaraperic: Ah I see now that the color is not necessary, but sizes need to be set intrinsically, which I can't do in the component. it will not work if the wrapper element (span) has h-6 w-6.

Talking about this part in HeadingWithIcon component
```
<div className="inline-flex items-center gap-3">
  <span className=" text-brand-300" aria-hidden={true}>
     {icon}
   </span>
   <Heading tag="h3" variant="lg">
     {children}
   </Heading>
</div>
```

PR #76, opened 2024-04-19
STYLE: apply section on grants page

barbaraperic:
## Changes
- updated the Apply now section in the grants page

## Visuals

Screenshot 2024-04-19 at 09 48 18
Screenshot 2024-04-19 at 09 48 12

### DIFF
--- a/src/app/_components/CallToActionSection.tsx
+++ b/src/app/_components/CallToActionSection.tsx
@@ -1,0 +1,20 @@
+import { Heading } from '@/components/Heading'
+
+type CallToActionSectionProps = {
+  title: string
+  children?: React.ReactNode
+}
+
+export function CallToActionSection({
+  title,
+  children,
+}: CallToActionSectionProps) {
+  return (
+    <section className="m-auto space-y-6 text-left sm:text-center [&>p]:max-w-[60ch]">
+      <Heading tag="h2" variant="3xl">
+        {title}
+      </Heading>
+      {children}
+    </section>
+  )
+}
--- a/src/app/grants/page.tsx
+++ b/src/app/grants/page.tsx
@@ -11,6 +11,7 @@ import { generateWebPageStructuredData } from '@/utils/structuredData'
 
 import { attributes } from '@/content/pages/grants.md'
 
+import { Button } from '@/_components/Button'
 import { PATHS } from '@/constants/paths'
 import { FILECOIN_URLS } from '@/constants/siteMetadata'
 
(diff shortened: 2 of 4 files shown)


src/app/grants/page.tsx
```diff
@@ -171,16 +172,25 @@ export default function Grants() {
...
-        <TextLink href={FILECOIN_URLS.social.slack.href}>Learn More</TextLink>
-        <GetInvolvedList />
+        <Button
+          className="sm:self-start"
+          variant="primary"
+          href={FILECOIN_URLS.grants.email}
+        >
+          Apply Now
+        </Button>
       </section>
```
mirhamasala, line 194: @barbaraperic Can you please have a look at the Figma designs, and confirm that we'll need to this component in a few pages? If so, let's create a component for it!

barbaraperic: You're right, there is a similar section. I've extracted the section styles and heading, left paragraph and button as children as they can vary with icons and inner text links.

Screenshot 2024-04-22 at 09 51 18

src/app/_components/CallToActionSection.tsx
```diff
@@ -0,0 +1,20 @@
+import { Heading } from '@/components/Heading'
+
+type CallToActionSectionProps = {
+  title: string
+  children?: React.ReactNode
+}
+
+export function CallToActionSection({
```
mirhamasala, line 8:
```suggestion
export function CTASection({
```

src/app/grants/page.tsx
mirhamasala: I understand the thinking. Still, we should try and move the text and button inside of the component. Not unlike we do for the optional `cta` in  `Card` (please note that `Card` needs some refactoring so asses yourself what can be reused!)

```
        {cta && (
          <CustomLink
            href={cta.href}
            aria-label={cta.ariaLabel}
            className="absolute inset-0 rounded-lg focus:outline-2 focus:outline-brand-100"
          >
            <span className="absolute bottom-4 left-4 inline-flex items-center gap-2 text-brand-300">
              {cta.icon || <BookOpen size={24} aria-hidden={true} />}
              <span>{cta.text || 'Learn More'}</span>
            </span>
          </CustomLink>
        )}
```        

As for the up arrow icon with outgoing links we should add that as the permanent icon under `Button` for external links.

mirhamasala: Also - You don't need to take all options into account in this PR - Make the component work for this PR - and then we can get to the variants once we need to.

PR #77, opened 2024-04-22
Make pagination SSR

CharlyMartin:
This PR updates the implementation of the pagination component on the blog page. It changes the network boundary: the blog page is now a server component and only the pagination is a client component. The search functionality has been set aside for now but will also be a client component.

This follows Next's principle to [move client components down the tree](https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#moving-client-components-down-the-tree) and only "use client" where interactivity is needed, which means we don't send all the component Javascript of the blog page to the client.

_Before_
Before this PR, the pagination state was stored in `useState`, and synced with the URL.

_After_
Now, the URL holds the pagination state - via the `?page=` search param - and is the only source of truth. Hence, we don't need to sync states anymore.

Here's the breakdown of the changes:

1. `<Pagination>` now updates the URL directly thanks to `useRouter`, instead of relying on a setter function via `prop`,
2. The blog page receives server-side search params via `props` instead of using `useSearchParams` client-side,
3. `<BlogClient>` is now `<BlogList>`, as it’s not a client component anymore,
4. `<Card>` does not accept a pagination `prop` anymore. The blog filtering is handled upper in the tree and keep the data and UI concerns separate,
5. `usePagination()` does not set state anymore; it computes `pageCount` and sanitizes `currentPage`. Its return values are the `props` of `<Pagination>`, as indicated in the updated type definition.
6. `useUpdateHistory()` no longer exists as we don’t sync history.

_Resources_
https://nextjs.org/docs/app/building-your-application/rendering/server-components#dynamic-functions
https://nextjs.org/docs/app/api-reference/file-conventions/page#searchparams-optional


CharlyMartin:
@mirhamasala, as per our discussion this morning:

1. I added the search functionality back. I created a wrapper component, `<BlogSearchInput>` for this purpose.
2. I moved the code previously in `<BlogList>` in the parent page.

Now we use server state for the pagination and search features, the click/type interactions feel less snappy. Once we fix deploys, we can test whether they are fast enough in production.

Issue #82, opened 2024-04-23
Include Filecoin Slack section on homepage


mirhamasala:
Per Figma design [>>](https://www.figma.com/file/0pBfQs5tKI6bx5ypPIfVRA/FIL.org-V4-Prototypes?type=design&node-id=2147-50981&mode=design&t=Z1U0umQ1WL20yQjr-4)

<img width="632" alt="Screenshot 2024-04-23 at 08 24 44" src="https://github.com/FilecoinFoundationWeb/filecoin-foundation/assets/26620750/e9dc354a-e002-44fe-ab8e-73124584f19e">

PR #83, opened 2024-04-23
STYLE: home page slack section

barbaraperic:
## Changes
- This PR addresses issue #82 
- I refactored the `CTASection` a bit, sorry I rushed it the first time. Issue was with the mobile view, the button was not taking the full width

### Before

Screenshot 2024-04-23 at 14 13 40

### After

Screenshot 2024-04-23 at 14 13 54

## Note

- In the designs, this component has the following spacing: 32px between `Header` and `Text`, and 24px between `Text` and `Button`. I kept it 24px and 24px, since this was the design of the [Ready to Apply section in /grants 
](https://www.figma.com/file/0pBfQs5tKI6bx5ypPIfVRA/FIL.org-V4-Prototypes?type=design&node-id=1442-28546&mode=design&t=gnr0VjzuzA8lDRU3-4) - component which we are reusing. @filipagr just wanted to confirm if this is ok? Or you would like to stick with 32px and 24px?

### DIFF
--- a/src/app/_components/CTASection.tsx
+++ b/src/app/_components/CTASection.tsx
@@ -14,11 +14,11 @@ export function CTASection({ title, description, cta }: CTASectionProps) {
   return (
     <section className="flex flex-col items-center gap-16">
       <div className="h-px w-full bg-brand-100" />
-      <div className="flex flex-col space-y-6 whitespace-pre-line sm:block sm:text-center">
-        <Heading tag="h2" variant="3xl" className="max-w-72 sm:m-auto">
+      <div className="xs:block xs:max-w-md xs:text-center flex flex-col space-y-6">
+        <Heading tag="h2" variant="3xl">
           {title}
         </Heading>
-        <p className="max-w-[60ch]">{description}</p>
+        <p className="max-w-[75ch]">{description}</p>
         {cta && <Button href={cta.href}>{cta.text}</Button>}
       </div>
     </section>
--- a/src/app/_components/CTASection.tsx
+++ b/src/app/_components/CTASection.tsx
@@ -15,7 +15,7 @@ export function CTASection({ title, description, cta }: CTASectionProps) {
     <section className="flex flex-col items-center gap-16">
       <div className="h-px w-full bg-brand-100" />
       <div className="flex flex-col space-y-6 whitespace-pre-line sm:block sm:text-center">
-        <Heading tag="h2" variant="3xl">
+        <Heading tag="h2" variant="3xl" className="max-w-72 sm:m-auto">
           {title}
         </Heading>
         <p className="max-w-[60ch]">{description}</p>
--- a/src/app/_components/CTASection.tsx
+++ b/src/app/_components/CTASection.tsx
@@ -14,7 +14,7 @@ export function CTASection({ title, description, cta }: CTASectionProps) {
   return (
     <section className="flex flex-col items-center gap-16">
       <div className="h-px w-full bg-brand-100" />
-      <div className="space-y-6 text-left sm:text-center">
+      <div className="flex flex-col space-y-6 whitespace-pre-line sm:block sm:text-center">
         <Heading tag="h2" variant="3xl">
           {title}
         </Heading>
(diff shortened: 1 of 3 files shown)


src/app/page.tsx
```diff
@@ -186,6 +187,14 @@ export default function Home() {
           View All
         </Button>
       </PageSection>
+      <CTASection
+        title={`Become Part of Our\n Vibrant Community`}
```
mirhamasala, line 191: Can we use some sort of `max-width` here?

src/app/page.tsx
```diff
@@ -186,6 +187,14 @@ export default function Home() {
...
+        description="Join Filecoin's Slack to engage with the community and stay updated on the latest&nbsp;developments."
+        cta={{
+          href: FILECOIN_URLS.social.slack.href,
+          text: 'Join Filecoin Slack',
+        }}
+      ></CTASection>
```
mirhamasala, line 197: If there are no `children`, components can be self-closing. 🙏🏼

src/app/page.tsx
mirhamasala, line 191: If we are using `max-width` now, do we still need `\n`?

src/app/_components/CTASection.tsx
```diff
@@ -14,8 +14,8 @@ export function CTASection({ title, description, cta }: CTASectionProps) {
   return (
     <section className="flex flex-col items-center gap-16">
       <div className="h-px w-full bg-brand-100" />
-      <div className="space-y-6 text-left sm:text-center">
-        <Heading tag="h2" variant="3xl">
+      <div className="flex flex-col space-y-6 whitespace-pre-line sm:block sm:text-center">
+        <Heading tag="h2" variant="3xl" className="max-w-72 sm:m-auto">
           {title}
         </Heading>
         <p className="max-w-[60ch]">{description}</p>
```
mirhamasala, line 21: Agreed with @filipagr to lock at `75` characters

src/app/page.tsx
```diff
@@ -186,6 +187,14 @@ export default function Home() {
           View All
         </Button>
       </PageSection>
+      <CTASection
+        title="Become Part of Our Vibrant&nbsp;Community"
```
mirhamasala, line 191: Why are we adding `&nbsp` - Do we need to break this intentionally? Asking because it effects responsiveness.

barbaraperic: I added it to avoid having [orphan text](https://gomakethings.com/a-css-native-way-to-prevent-orphan-text/) - poor "Community" hanging by itself. Could remove if you're not a fan

mirhamasala: This is a good idea. I tried `text-wrap: pretty` but it didn't seem to do much. `text-wrap: balance` did work - which was a suggestion that @CharlyMartin made earlier. I've now added as a default `baseStyle` to the `Heading` component. Please do keep in mind that it's not fully supported yet [>>](https://caniuse.com/css-text-wrap-balance)

mirhamasala:
@barbaraperic - Have a look at the `CR` commits - And let me know if that looks alright. To me, some of the `xs` breakpoint use didn't seem right. (And if I didn't miss something, I wonder if we need it.)

barbaraperic:
> @barbaraperic - Have a look at the `CR` commits - And let me know if that looks alright. To me, some of the `xs` breakpoint use didn't seem right. (And if I didn't miss something, I wonder if we need it. If we don't need it - I prefer we revert [this](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/86) commit.)

I don't mind if you think it looks better - maybe we should then just make sure that the design matches the implementation, and let Filipa know because we have `xs` screens in Figma that are different from `sm` screens.

mirhamasala:
> We have xs screens in Figma that are different from sm screens.

The breakpoint before `sm` is effectively `xs` - right?

barbaraperic:
> > We have xs screens in Figma that are different from sm screens.
> 
> The breakpoint before `sm` is effectively `xs` - right?

Exactly, `xs` is 376px, `sm` is 640px

barbaraperic:
@mirhamasala removed the `xs` breakpoint - you're right, not necessary!

mirhamasala (approved):

PR #84, opened 2024-04-24
STYLE: 404 page

barbaraperic:
https://github.com/FilecoinFoundationWeb/filecoin-foundation/assets/50910606/d83113a2-1096-4c05-82e7-f18fa1ced42b

### DIFF
--- a/src/app/_components/BreadCrumbsWrapper.tsx
+++ b/src/app/_components/BreadCrumbsWrapper.tsx
@@ -7,9 +7,15 @@ import { BreadCrumbs } from '@/components/BreadCrumbs'
 import { PATHS } from '@/constants/paths'
 
 export function BreadCrumbsWrapper() {
-  const pathname = usePathname()
+  const pathname = usePathname() as string
+  const allPaths = Object.values(PATHS).map((route) => route.path) as string[]
 
   if (pathname === PATHS.HOME.path) {
+    return null
+  }
+
+  // this is to exclude breadcrumbs in not-found
+  if (!allPaths.includes(pathname)) {
     return null
   }
 
--- a/src/app/not-found.tsx
+++ b/src/app/not-found.tsx
@@ -1,0 +1,22 @@
+import { baseButtonStyles, variantButtonStyles } from '@/components/Button'
+import { CustomLink } from '@/components/CustomLink'
+
+export default function NotFound() {
+  return (
+    <div className="xs:items-center xs:justify-center flex flex-col gap-6">
+      <div className="h-[102px]"></div>
+      <span className="text-8xl">404</span>
+      <h2 className="text-3xl">Page Not Found</h2>
+      <p className="xs:max-w-[35ch] xs:text-center">
+        We&apos;re sorry, but the page you were looking for could not be found
+      </p>
+      <CustomLink
+        href="/"
+        className={`${baseButtonStyles} ${variantButtonStyles.primary}`}
+      >
+        Go to Home Page
+      </CustomLink>
+      <div className="h-[102px]"></div>
+    </div>
+  )
+}
--- a/src/app/_components/Button.tsx
+++ b/src/app/_components/Button.tsx
@@ -17,12 +17,15 @@ type ButtonInnerProps = {
   children: React.ReactNode
 }
 
-const variantStyles = {
+export const variantButtonStyles = {
   primary:
     'border-brand-300 bg-brand-300 text-brand-800 hover:border-brand-400 hover:bg-brand-400',
   ghost:
     'border-white bg-brand-800 text-brand-100 hover:border-brand-400 hover:text-brand-400 focus:text-brand-400',
 }
+
+export const baseButtonStyles =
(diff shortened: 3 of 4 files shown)


src/app/not-found.tsx
```diff
@@ -0,0 +1,16 @@
...
+export default function NotFound() {
+  return (
+    <div className="xs:items-center xs:text-center m-auto flex max-w-xs flex-col gap-6">
+      <div className="h-[104px]"></div>
+      <span className="text-8xl">404</span>
+      <h2 className="text-3xl">Page Not Found</h2>
+      <p>
+        We&apos;re sorry, but the page you were looking for could not be found
+      </p>
+      <Button href="/">Go to Home Page</Button>
```
mirhamasala, line 12: Let's use the `PATHS` object

src/app/not-found.tsx
```diff
@@ -0,0 +1,16 @@
+import { Button } from '@/components/Button'
+
+export default function NotFound() {
+  return (
+    <div className="xs:items-center xs:text-center m-auto flex max-w-xs flex-col gap-6">
+      <div className="h-[104px]"></div>
+      <span className="text-8xl">404</span>
+      <h2 className="text-3xl">Page Not Found</h2>
```
mirhamasala, line 8: Perhaps we should talk to @filipagr about taking these font sizes down a notch on `xs`

src/app/_components/BreadCrumbsWrapper.tsx
```diff
@@ -8,10 +8,16 @@ import { PATHS } from '@/constants/paths'
 
 export function BreadCrumbsWrapper() {
   const pathname = usePathname()
+  const allPaths = Object.values(PATHS).map((route) => route.path) as string[]
 
   if (pathname === PATHS.HOME.path) {
     return null
   }
 
+  // this is to exclude breadcrumbs in not-found.tsx
```
mirhamasala, line 17: I started a discussion on Slack on whether we should only include `Breadcrumbs` on individual blog posts, ecosystem projects and events pages. That is, pages that aren't standalone. This implementation will change pending on that outcome.

barbaraperic: @mirhamasala based on Friday's discussion - just to confirm, we agreed on not having `Breadcrumbs` on 404?

src/app/not-found.tsx
```diff
@@ -0,0 +1,16 @@
+import { Button } from '@/components/Button'
+
+import { PATHS } from './_constants/paths'
+
+export default function NotFound() {
+  return (
+    <div className="m-auto flex max-w-xs flex-col gap-6 py-[104px] sm:items-center sm:text-center">
```
mirhamasala, line 7: Are you following up with Filipa regarding [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/84#discussion_r1580669681)? 🙏🏼

src/app/_components/BreadCrumbsWrapper.tsx
```diff
@@ -8,10 +8,16 @@ import { PATHS } from '@/constants/paths'
 
 export function BreadCrumbsWrapper() {
   const pathname = usePathname()
+  const allPaths = Object.values(PATHS).map((route) => route.path) as string[]
```
mirhamasala, line 11: This wouldn't catch dynamic paths.

src/app/_components/BreadCrumbsWrapper.tsx
```diff
@@ -8,10 +8,16 @@ import { PATHS } from '@/constants/paths'
...
 
   if (pathname === PATHS.HOME.path) {
     return null
   }
 
+  // this is to exclude breadcrumbs in not-found.tsx
+  if (!allPaths.includes(pathname)) {
+    return null
+  }
+
```
mirhamasala, line 21: Can be consolidated.

src/app/_components/BreadCrumbsWrapper.tsx
barbaraperic: good catch!

src/app/not-found.tsx
filipagr: Hey! Already discussed this with Barbara and if we're taking the Page not Found font size down a notch on mobile, then we should also do it on the other screens as well, because we're using the same font sizes. The 404 font size was already taken down to 7xl, instead of 8xl. The same on the 500 error page.

src/app/_components/BreadCrumbsWrapper.tsx
barbaraperic: For blog posts and case studies - could we add a condition in blog and case study pages to render the 404 if blog/case study not found? At the moment we have:

```
 if (!featuredPost) {
    throw new Error('Featured post not found')
  } 
```

src/app/_components/BreadCrumbsWrapper.tsx
barbaraperic: what do you mean by that? 🤔

src/app/not-found.tsx
mirhamasala: @filipagr We can look into that another time, thank you.

PR #93, opened 2024-04-26
Revert "CONFIG: new xs breakpoint"

mirhamasala:
Reverts FilecoinFoundationWeb/filecoin-foundation#86 - We don't need the `xs` breakpoint after all.


mirhamasala:
@barbaraperic PRs, especially when isolated like this one, can be reverted directly from GitHub.

Screenshot 2024-04-26 at 15 25 54

PR #94, opened 2024-04-26
Include SortListbox component

mirhamasala:
Include `SortListbox` component

Screenshot 2024-04-29 at 09 42 53


src/app/_components/SortListbox.tsx
```diff
@@ -0,0 +1,78 @@
...
+type SortSetting = {
+  id: SortType
+  name: string
+}
+
+type SortListboxProps = {
+  onSortTypeChange: (selectedOption: SortType) => void
+}
+
+export const sortSettings: SortSetting[] = [
```
CharlyMartin, line 24: `sortSettings` is exported but never used outside of this file.

src/app/_components/SortListbox.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const [selectedSortType, setSelectedSortType] = useState(sortSettings[0])
+
+  function handleChange(newSortType: SortSetting) {
+    setSelectedSortType(newSortType)
+    onSortTypeChange(newSortType.id)
+  }
+
+  return (
+    <Listbox value={selectedSortType} onChange={handleChange}>
+      {({ open }) => (
```
CharlyMartin, line 39: `open` is declared but never used in the function body.

src/app/_components/SortListbox.tsx
```diff
@@ -0,0 +1,78 @@
...
+              />
+            </span>
+          </Listbox.Button>
+          <Listbox.Options className="absolute z-10 mt-14 min-w-40 overflow-hidden rounded-lg border border-brand-100 bg-brand-800 py-1 text-brand-100 focus-within:outline-2 focus:outline-2 focus:outline-brand-100">
+            {sortSettings.map((option) => (
+              <Listbox.Option key={option.id} value={option} as={Fragment}>
+                {({ active, selected }) => (
+                  <li
+                    className={clsx(
+                      'flex items-center justify-between gap-12 px-5 py-2',
```
CharlyMartin, line 61:
```suggestion
                      'flex cursor-default items-center justify-between gap-12 px-5 py-2',
```

On Firefox, the `li` options have a text cursor. The default one would be best in this case?

src/app/_utils/sortEntriesByDate.ts
```diff
@@ -0,0 +1,24 @@
...
+
+interface SortableByDate {
+  startDate: string
+}
+
+export function sortEntriesByDate<T extends SortableByDate>(
+  entries: T[],
+  sortType: SortType,
+): T[] {
+  return entries.slice().sort((a, b) => {
```
CharlyMartin, line 11:
```suggestion
  return [...entries].sort((a, b) => {
```

I get what you are trying to do here - create a copy to avoid mutating the original array - but `slice` got me confused for a second. I wondered if you forgot to add indexes.

What do you think about my suggestion?

src/app/_components/SortListbox.tsx
```diff
@@ -0,0 +1,78 @@
...
+type SortListboxProps = {
+  onSortTypeChange: (selectedOption: SortType) => void
+}
+
+export const sortSettings: SortSetting[] = [
+  { id: SortType.Newest, name: 'Newest' },
+  { id: SortType.Oldest, name: 'Oldest' },
+]
+
+export function SortListbox({ onSortTypeChange }: SortListboxProps) {
```
CharlyMartin, line 29: TS warned me that the `onSortTypeChange` prop is not serializable.

If we were to pass this prop from a server component, it would break. But I'm not sure if we are planning to.

CharlyMartin (commented):
Good one @mirhamasala, I left a few code comments, but it looks good 👌

Two more general comments:

1. The designs contain several `ListBox` instances: Sort, Category, Region, etc. This component should receive its list of options via prop to make it more generic and reusable. What do you think?
2. Pagination and Search rely on URL params for their state. Should we consider implementing a similar solution for consistency's sake?

mirhamasala:
@CharlyMartin 

>The designs contain several ListBox instances: Sort, Category, Region, etc. This component should receive its list of options via prop to make it more generic and reusable. What do you think?

Not sure that will be the exact implementation but yes, I've already taken this into account. I was thinking of extending

```

mirhamasala:
`

PR #102, opened 2024-04-30
Governance Page - About section

barbaraperic:
## Changes
- new `PageSectionWithImage` component
- the `PageSectionWithImage` component handles the case where we don't have an image, it applies only the max paragraph length (maybe we should include this in some global css at some point if design agrees)
- extracted `setDescriptionParagraph` in its own file, since we now use it in 2 places (removed the 
fixed spacing between paragraphs so now it inherits from the parent)
- we use `ImageProps` in 2 places - for now I've exported it from the `PageHeader`, let me know if you'd prefer to have it under types folder. 

### With image

https://github.com/FilecoinFoundationWeb/filecoin-foundation/assets/50910606/d6266c8a-1272-4e93-beec-32c2b7791d5e

### Without image

Screenshot 2024-04-30 at 08 06 06


src/app/_components/PageHeader.tsx
```diff
@@ -12,7 +14,7 @@ type CTAProps = {
   icon?: React.ReactNode
 }
 
-type ImageProps = {
+export type ImageProps = {
```
mirhamasala, line 17: I think we can centralize some of these under `_types` - Maybe under a file called `SharedProps` - the `cta` type could also go here, for example.

src/app/governance/page.tsx
```diff
@@ -97,39 +98,14 @@ export default function Governance() {
...
-        </p>
-      </section>
+      <PageSectionWithImage
+        kicker="About"
+        title="Understanding Governance"
+        description={[
+          'In the Filecoin ecosystem – which is decentralized and open source – governance refers to the rights, rules, processes, and norms that empower community members to collectively determine the development of the protocol.',
+          'The Foundation’s role includes developing and implementing processes and policies for decision-making and managing community engagement and feedback – all to support the development of the ecosystem at large. ',
+        ]}
+      ></PageSectionWithImage>
```
mirhamasala, line 108: Self-closing.

src/app/_utils/setDescriptionParagraph.tsx
```diff
@@ -0,0 +1,14 @@
...
+    return (
+      <>
+        {description.map((paragraph, index) => (
+          <p key={index}>{paragraph}</p>
+        ))}
+      </>
+    )
+  }
+  return <p>{description}</p>
+}
```
mirhamasala, line 14: I find it slightly odd to have this bit under `utils` - This feels more like a component if anything.

src/app/_components/PageSectionWithImage.tsx
```diff
@@ -0,0 +1,48 @@
+import Image from 'next/image'
```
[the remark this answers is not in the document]
mirhamasala: Maybe this helps?

```

const containerClasses = (hasImage: Boolean) => {
  return hasImage
    ? 'grid grid-cols-1 gap-6 sm:grid-cols-2'
    : 'mb-6 max-w-[60ch]'
}
```

mirhamasala (approved):
@barbaraperic - Have a look at the CR - And let me know if it works. 🙏🏼

src/app/_components/PageSection.tsx
```diff
@@ -1,26 +1,56 @@
...
-  description?: string
+  description?: DescriptionTextType
+  image?: ImageProps
   children?: React.ReactNode
 }
 
+const containerClasses = (hasImage: Boolean) => {
+  return hasImage
+    ? 'grid grid-cols-1 gap-6 sm:grid-cols-2'
+    : 'mb-6 max-w-[60ch]'
```
mirhamasala, line 23: I believe we settled on `max-w-75` - We should standardize this by adding it to the config. Something like:

```
module.exports = {
  theme: {
    extend: {
      maxWidth: {
        readable: '60ch'
      }
    }
  }
}
```

This way, we can use `max-w-readable` allover.

barbaraperic:
> @barbaraperic - Have a look at the CR - And let me know if it works. 🙏🏼

Looks good! 👌 Added some tiny fixes to handle some edge cases. For ex, if we have an image but no description and added `mb` in case we have an image and children present.

PR #106, opened 2024-05-01
Showcase featured ecosystem projects on homepage

mirhamasala:
Per Figma Design [>>](https://www.figma.com/file/0pBfQs5tKI6bx5ypPIfVRA/FIL.org-V4-Prototypes?type=design&node-id=538-12336&mode=design&t=LYXpIONPhTbFU9wC-4)

Screenshot 2024-05-01 at 11 11 10

## Notes:

- Category tags will be picked up in a different PR
- Image aspect ratios still need work with design team
- Lot of Ecosystem projects still need to have their images imported


src/app/_components/FeaturedEcosystemProjects.tsx
```diff
@@ -1,34 +1,44 @@
...
+const FEATURED_PROJETS_LIMIT = 6
 
 export function FeaturedEcosystemProjects() {
-  if (featuredEcosystemProjects.length === 0) {
+  const featuredEcosystemProjects = shuffleArray(ecosystemProjects).slice(
+    0,
+    FEATURED_PROJETS_LIMIT,
+  )
+
+  if (featuredEcosystemProjects.length < FEATURED_PROJETS_LIMIT) {
```
barbaraperic, line 20: If there are 5 `featuredEcosytemProjects` and the` FEATURED_PROJECTS_LIMIT` is 6, we show No "featured projects available"?

src/app/_components/FeaturedEcosystemProjects.tsx
```diff
@@ -1,34 +1,44 @@
...
-      {featuredEcosystemProjects.map(({ slug, title, description }) => (
-        <li key={slug} className="ml-0">
-          <Badge>Featured Project</Badge>
-          <Heading tag="h3" variant="lg">
-            {title}
-          </Heading>
-          <p>{description}</p>
-          <TextLink href={`${PATHS.ECOSYSTEM.path}/${slug}`}>See More</TextLink>
-        </li>
+    <CardLayout>
```
barbaraperic, line 25: gap is 24px in the designs.

src/app/page.tsx
```diff
@@ -159,9 +160,11 @@ export default function Home() {
 
       <PageSection
         kicker="Learn"
-        title="Use Cases"
-        description="Learn about projects building on and using Filecoin."
-      />
+        title="Filecoin Use Cases"
+        description="Navigate the Filecoin Ecosystem Explorer, a crowd-sourced and open database to showcase projects powering the Filecoin network."
+      >
+        <FeaturedEcosystemProjects />
```
barbaraperic, line 166: Should it have View All button? 
Screenshot 2024-05-01 at 12 48 24

src/app/_utils/shuffleArray.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Randomly shuffles the elements of an array and returns a new array using the Fisher-Yates (Durstenfeld) shuffle algorithm.
```
barbaraperic, line 2: Why are we doing this? Could it happen that the user clicks on the card, and on the return to the Ecosystem page, the card is not shown because of the shuffle?

barbaraperic: Also this happens - clicking on the card navigates to /ecosystem/slug, but the breadcrumbs don't navigate to the same place. 🤔 not sure if it's a big deal, but it would be tricky to find that card again if you don't remember which page you were on.

https://github.com/FilecoinFoundationWeb/filecoin-foundation/assets/50910606/2ddff16d-eb13-44d6-ad5c-ba098a6634a4

src/app/_components/Card.tsx
```diff
@@ -59,13 +66,16 @@ export function Card({
       )}
     >
       {image?.url && (
-        <div className="relative block h-52">
+        <div className="relative block aspect-video h-52">
```
barbaraperic, line 69: is the h-52 needed? Btw responsiveness looks really nice, super smooth

src/app/_utils/shuffleArray.ts
mirhamasala: > Could it happen that the user clicks on the card, and on the return to the page, the card is not shown because of the shuffle?

This is a good point. Need to align with @filipagr and Heath on expectations here. Maybe it would better to have a button that users can click to shuffle them if they want. Or something like that.

mirhamasala: > Also this happens - clicking on the card navigates to /ecosystem/slug, but the breadcrumbs don't navigate to the same place. 🤔 not sure if it's a big deal, but it would be tricky to find that card again if you don't remember which page you were on.

Not sure I understand. What would be the expected behaviour? This is a section of featured Ecosystem Projects on the homepage. Clicking on a single project brings you to the relevant single Ecosystem project page. Following `/ecosystem` goes to the Ecosystem page.

src/app/page.tsx
mirhamasala: Totally. 🙏🏼

src/app/_components/FeaturedEcosystemProjects.tsx
mirhamasala: Yes. This will never happen but I understand the confusion. I will revert back to `=== 0`

src/app/_components/Card.tsx
mirhamasala: Yes. Needed.

WITH
Screenshot 2024-05-02 at 15 03 58

WITHOUT
Screenshot 2024-05-02 at 15 04 05

src/app/_utils/shuffleArray.ts
mirhamasala: @barbaraperic After raising your concerns, I talked to design and marketing - And we decided that these will be featured projects that the content team can set. New CR reflects this.

Let me know if this is still something you want to discuss issue? [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/106#discussion_r1586273635)

src/app/_components/FeaturedEcosystemProjects.tsx
mirhamasala: When you notice inconsistencies like these - It's probably an oversight on design sides. And it's a good idea to check [>>](https://filecoinfoundation.slack.com/archives/C06MTEQ3SP6/p1714658929642879)

barbaraperic (approved):
Looks good 👌

PR #108, opened 2024-05-01
Governance Page - Upcoming Events section

barbaraperic:
## Changes
- new `GovernanceCalendarCard` component
- new `convertToISOTime` utils file

## How to test
- test if clicking on "Add Event to Calendar" allows you to add the event to your calendar. Also check if the time matches the time stated in the `GovernanceCalendarCards`

## Note
1.  you can find the` NEXT_PUBLIC_CALENDAR_ID` and `NEXT_PUBLIC_CALENDAR_API_KEY` for `.env.local` in 1password, under UXIT team
2. Why `GovernanceCalendarCards` takes `currentDate` as a prop?
- in order to get all future dates I've used query param `timeMin` which takes ISO date as a value. Since I wanted to use today's date I've used `new Date()` method , however that caused some issues with the API call since the Date object hasn't been fully initialised by the time the URL is constructed, leading to this error:
`Uncaught (in promise) Error: A listener indicated an asynchronous response by returning true, but the message channel closed before a response was received`
To prevent that I passed the `currentDate` as a prop, and that seemed to work. If you think it could be done in a better way, let me know!

## Visuals

Screenshot 2024-05-14 at 15 50 36
Screenshot 2024-05-14 at 15 50 29
Screenshot 2024-05-14 at 15 50 23


mirhamasala:
@barbaraperic - Requested @CharlyMartin review. 🧐

mirhamasala:
@barbaraperic Please check [>>](https://filecoinfoundation.slack.com/archives/C06MTEQ3SP6/p1714663574474389?thread_ts=1714464163.788199&cid=C06MTEQ3SP6) for Lucky's answer on when these calls happen.

src/app/_components/GovernanceCalendarCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+type CalendarProps = {
+  startDate: string
+}
+
+function Calendar({ startDate }: CalendarProps) {
+  const { day, month } = extractTimeFromISO(startDate)
+  return (
+    <div className="min-w-xs grid grid-rows-[_40px,_96px] rounded-md border border-blue-400 bg-blue-500 sm:min-w-32">
+      <span className="flex items-center justify-center rounded-tl-[5px] rounded-tr-[5px] text-base font-normal uppercase">
```
CharlyMartin, line 28: I don't understand the reason behind `rounded-tl-[5px] rounded-tr-[5px]`. Why 5px?

Also, why not use `rounded-t-[value]` in one go?

src/app/_components/GovernanceCalendarCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+type CalendarProps = {
+  startDate: string
+}
+
+function Calendar({ startDate }: CalendarProps) {
+  const { day, month } = extractTimeFromISO(startDate)
+  return (
+    <div className="min-w-xs grid grid-rows-[_40px,_96px] rounded-md border border-blue-400 bg-blue-500 sm:min-w-32">
```
CharlyMartin, line 27: Cool use of grid 😎

src/app/_components/GovernanceCalendarCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <span className="flex items-center justify-center rounded-tl-[5px] rounded-tr-[5px] text-base font-normal uppercase">
+        {month}
+      </span>
+      <span className="flex items-center justify-center rounded-bl-[5px]  rounded-br-[5px] bg-blue-100 text-5xl font-bold text-blue-400">
```
CharlyMartin, line 31: Same comment as above.

src/app/_components/GovernanceCalendarCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+  durationInMinutes,
+  cta,
+  as: Tag = 'li',
+}: GovernanceCalendarCardProps) {
+  const { time } = extractTimeFromISO(startDate)
+  const endTime = addMinutesToISO(startDate, durationInMinutes)
+
+  if (!startDate) {
+    return
+  }
```
CharlyMartin, line 50: Based on the type definition `GovernanceCalendarCardProps`, it looks like `startDate` is mandatory, so perhaps this early return is not necessary? Or maybe I'm missing something?

src/app/_components/GovernanceCalendarCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+    <Tag className="flex flex-col gap-6 rounded-md border border-blue-500 p-4 sm:flex-row">
+      <Calendar startDate={startDate} />
+
+      <div className="flex flex-1 flex-col items-start justify-between gap-6">
+        <div className="flex flex-col gap-3">
+          <Heading tag="h3" variant="lg">
+            {title}
+          </Heading>
+
+          <div className="flex w-max items-center gap-1 rounded-lg border border-blue-100 px-1 py-[2px] text-xs uppercase">
```
CharlyMartin, line 62: There's a `0.5` value for spacing in Taliwind 🙂

```suggestion
          <div className="flex w-max items-center gap-1 rounded-lg border border-blue-100 px-1 py-0.5 text-xs uppercase">
```

src/app/_components/GovernanceCalendarCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+            <Clock size={13} />
+            <span className="mb-[-1px]">
```
CharlyMartin, line 64: Same visual result as your code, but the Tailwind convention for negative margin is:

```suggestion
            <span className="-mb-[1px]">
```

src/app/_components/GovernanceCalendarCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+    </div>
+  )
+}
+
+export function GovernanceCalendarCard({
+  title,
+  startDate,
+  durationInMinutes,
+  cta,
+  as: Tag = 'li',
```
CharlyMartin, line 43: Just curious, can `as` be anything else than `li`?

src/app/_utils/convertISOTime.ts
```diff
@@ -0,0 +1,46 @@
+type extractTimeFromISOType = {
+  day: number
+  month: string
+  time: string
+}
+
+export function extractTimeFromISO(
+  isoDateString: string,
+): extractTimeFromISOType {
+  const date = new Date(isoDateString)
```
CharlyMartin, line 10: How do we check that this date is in the right ISO format? TypeScript only checks that a string is passed.

Maybe we should add an extra check to these 2 functions? What do you think? Maybe not critical for now.

CharlyMartin (commented):
Aside from my comments, I'm sharing this [Figma conversation](https://www.figma.com/file/0pBfQs5tKI6bx5ypPIfVRA?type=design&node-id=1442-35842&mode=design#793451765) about adding another tag to `GovernanceCalendarCard`

I might have a second look once these are resolved, thank you 🙂

src/app/_components/GovernanceCalendarCard.tsx
CharlyMartin: Oh I didn't know about the `px` value, cool one!

src/app/_components/GovernanceCalendarCard.tsx
barbaraperic: Ah you're right! Think `rounded-t-md` will do, for some reason it seemed off when I put the same value for outer and inner border, but on the second look it looks fine.

src/app/_components/GovernanceCalendarCard.tsx
barbaraperic: Put it in the case we don't get the data from the API - we don't want to show the card?

src/app/_components/GovernanceCalendarCard.tsx
barbaraperic: Good point. Since CardLayout is a `ul`, I can just set the component to be `li,` guess there's no cases where' we'd need to modify it.

src/app/_components/Badge.tsx
```diff
@@ -10,12 +13,17 @@ const borderStyles = {
...
 
-export function Badge({ borderColor = 'brand-500', children }: BadgeProps) {
+export function Badge({
+  borderColor = 'brand-500',
+  icon,
+  children,
+}: BadgeProps) {
   const baseStyles =
-    'inline-block max-w-fit rounded-lg border bg-brand-800 px-2 py-1 text-xs font-semibold capitalize text-brand-100'
+    'inline-block max-w-fit rounded-lg border bg-brand-800 px-2 py-1 text-xs font-semibold capitalize text-brand-100 flex gap-1'
```
mirhamasala, line 22: This has two different display styles. Also needs to be sorted. 🙏🏼

src/app/_components/GovernanceCalendarCards.tsx
```diff
@@ -0,0 +1,108 @@
...
+                    borderColor="brand-100"
+                    icon={Clock}
+                  >{`UTC ${startTime} - ${endTime}`}</Badge>
+                  <Badge>Zoom</Badge>
+                </div>
+                <Heading tag="h3" variant="lg">
+                  {summary}
+                </Heading>
+              </div>
+              <TextLink className="flex gap-2 text-base" href={htmlLink}>
```
mirhamasala, line 98: We don't want to use the `className` prop. We should actually create a ticket to remove it so we're not tempted.

src/app/_components/GovernanceCalendarCards.tsx
```diff
@@ -0,0 +1,108 @@
...
+    <div className="grid min-h-40 min-w-36 grid-rows-[_40px,auto] rounded-md border border-blue-400 bg-blue-500">
+      <span className="flex items-center justify-center rounded-t text-base font-normal uppercase">
+        {month}
+      </span>
+      <span className="flex flex-1 items-center justify-center rounded-b bg-blue-100 text-5xl font-bold text-blue-400">
+        {day}
+      </span>
+    </div>
+  )
+}
```
mirhamasala, line 35: Think this has to be moved down, to right above `export function GovernanceCalendarCards()`.

src/app/_utils/convertISOTime.ts
```diff
@@ -0,0 +1,25 @@
+type extractTimeFromISOType = {
+  day: number
+  month: string
+  time: string
+}
```
CharlyMartin, line 5: When return types are used only once, like here, I believe they are unnecessary. We can let TS infer the return type and simplify the code.

```suggestion
```

src/app/_utils/convertISOTime.ts
CharlyMartin: Update: The string passed comes from the Google API, so we can trust that it's safe. We can ignore this check for now.

But in general, I like to validate each piece of data that comes from outside our code :)

CharlyMartin (changes requested):

src/app/_components/GovernanceCalendarCards.tsx
```diff
@@ -0,0 +1,108 @@
...
+export function GovernanceCalendarCards() {
+  const [currentDate] = useState(new Date())
+  const timeMin = encodeURIComponent(currentDate.toISOString())
+
+  const url = `${GOOGLE_CALENDAR_API_URL}${process.env.NEXT_PUBLIC_CALENDAR_ID}/events?maxResults=6&singleEvents=true&timeMin=${timeMin}&key=${process.env.NEXT_PUBLIC_CALENDAR_API_KEY}`
+
+  const { data: events, error } = useSWR(url, getEvents)
+
+  if (error) return <div>Failed to load events</div>
+  if (!events) return <div>Loading events...</div>
```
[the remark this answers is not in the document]
barbaraperic: Good point for the loading state! I would still keep the error in case of server error or rate limit exceeded. Doesn't hurt 🤷‍♀️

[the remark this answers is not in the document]
barbaraperic: Good thinking!

barbaraperic:
> Hey @barbaraperic
> 
> I added one commit ([9bf2f5b](https://github.com/FilecoinFoundationWeb/filecoin-foundation/commit/9bf2f5b8fa962fa062ff2d06c77a0e73428bca31)) to your PR to add type safety to external data - the events coming from Google API.
> 
> I wanted to add suggestions to the PR, but I needed to install [zod](https://zod.dev/?id=parse) to show you, so I ended up doing it... I hope it's okay! Now, `events.items` is automatically typed thanks to `zod`, no need to use `any` anymore.
> 
> Also, I fixed the issue you had with `new Date` by "locking it" in a `useState`.
> 
> Let me know what you think / if it makes sense :)
> 
> => I haven't been able to test it since I couldn't access the API keys. But I tested the error state, and it works 😄 I'll have a second look once I have the credentials.
> 
> Thank you!

Nice! The only thing I added is a try catch in case there's a validation error 

```
 try {
    return eventsSchema.parse(data)
  } catch (error) {
    throw new Error(`getEvents: Data validation failed - ${error}`)
  }
```

If you and @mirhamasala agree, I would love to use jest to write a few test, for the peace of mind.

src/app/_components/GovernanceCalendarCards.tsx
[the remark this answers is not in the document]
CharlyMartin: Fine by me :)

CharlyMartin (approved):

src/app/_components/GovernanceCalendarCards.tsx
mirhamasala: We create a separate one! Thank you.

mirhamasala (approved):
Well done, @barbaraperic and @CharlyMartin. 🙌🏼✨ #fisi-today! 

~Tests yes - can be in another PR!

PR #109, opened 2024-05-01
Events Page - Get Involved Section

mirhamasala:
Style Events Page - Get Involved Section per Figma design [>>](https://www.figma.com/file/0pBfQs5tKI6bx5ypPIfVRA/FIL.org-V4-Prototypes?type=design&node-id=1231-33806&mode=design&t=Bzqh6wru2l2AsJtp-4)

Screenshot 2024-05-03 at 07 20 58

*Note:* Graphics to be added when design gets them.

### DIFF
--- a/src/app/events/data/getInvolvedData.ts
+++ b/src/app/events/data/getInvolvedData.ts
@@ -5,10 +5,9 @@ import { FILECOIN_FOUNDATION_URLS } from '@/constants/siteMetadata'
 export const getInvolvedData = [
   {
     title: 'Sponsor an Event',
-    description:
-      'To be featured partner or sponsor an upcoming Foundation-hosted event, reach out to sponsorships@fil.org.',
+    description: `To be featured partner or sponsor an upcoming Foundation-hosted event, reach out to ${FILECOIN_FOUNDATION_URLS.events.sponsorshipsEmail.replace('mailto:', '')}.`,
     cta: {
-      href: FILECOIN_FOUNDATION_URLS.events.sponsorShipsEmail,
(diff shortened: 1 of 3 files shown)


src/app/events/page.tsx
```diff
@@ -93,6 +96,22 @@ export default function Events() {
...
+        kicker="Get Involved"
+        title="Get in Touch With the Events Team"
+      >
+        <CardLayout>
+          {getInvolvedData.map(({ title, description, cta }) => (
+            <Card
+              key={title}
+              title={title}
+              description={description}
+              cta={{ ...cta }}
```
barbaraperic, line 110: why spread in an object? it's already an object?

src/app/events/data/getInvolvedData.ts
```diff
@@ -0,0 +1,35 @@
...
+    cta: {
+      href: '#',
+      text: 'Submit Form',
+      icon: Clipboard,
+    },
+  },
+  {
+    title: 'Join Filecoin Orbit',
+    description:
+      'The Filecoin Orbit program promotes Filecoin technology through community events that raise network awareness. Apply to be an Orbit Ambassador.',
```
barbaraperic, line 28: Orbit Ambassador should be a link?

mirhamasala: The entire card is clickable - We shouldn't then also have links within the card. Something I already alerted design to. 🙏🏼

src/app/events/page.tsx
mirhamasala: I guess some refactor gone wrong. 🙏🏼

PR #112, opened 2024-05-02
Use Icon component

mirhamasala:
- Refactor to use `Icon` component everywhere
- Abstract `CTAProps` into shared types
- Refactor to use `ImageProps` and `CTAProps` types


barbaraperic (approved):

mirhamasala:
> Nice 👏 just 2 small things - think in SortListbox Icon components have color="inherit" prop, don't think it's needed.

Yes. You are right. It's a leftover. Will fix.

> Also wondering why you removed the ArrowUpRight icon in app/page? It's in the designs?||

We want to handle these globally for all external links, not just on a per case basis. Thus, this logic should be refactored into `Button`.

PR #116, opened 2024-05-06
Style navigation menu mobile

CharlyMartin:
This PR adds a sliding navigation menu for screen sizes up to `md` (mobiles and tablets):

1. I've added a generic `<SlideOver />` component to separate concerns and keep things cleaner.
2. I've reworked the `<Social />` component to make it reusable in the `<MobileNavigation />` component
3. I've implemented the sliding navigation on mobile as per the [Figma design](https://www.figma.com/file/0pBfQs5tKI6bx5ypPIfVRA/FIL.org-V4-Prototypes?type=design&node-id=1988-78176&mode=design&t=UvnAlLmcodPFI5Js-11)

About the `<MobileNavigation />` implementation, I've played with negative margins to make sure the link's touch targets are big enough. This is why I added `inline-block px-5 py-2.5` to `<TextLink />`, and `-mr-4 size-12` to buttons.

Unlike the Figma design, both the sliding menu and website have a `brand.800` background, which is not ideal for contrast. Adding the gradient background should solve this issue. If not, we can find other ways to better delimit the menu.

Voila!

PS: [The design team seems happy](https://filecoinfoundation.slack.com/archives/C06MTEQ3SP6/p1715009223575449) with the current implementation!

### DIFF
--- a/src/app/_components/Footer.tsx
+++ b/src/app/_components/Footer.tsx
@@ -72,7 +72,11 @@ export function Footer() {
       </div>
 
       <hr />
-      <Social />
+
+      <div className="sm:px-8">
+        <Social />
+      </div>
+
       <hr />
 
       <div className="flex flex-wrap gap-x-20 gap-y-6">
--- a/src/app/_components/MobileNavigation.tsx
+++ b/src/app/_components/MobileNavigation.tsx
@@ -1,0 +1,125 @@
+'use client'
+
+import { Fragment, useState, Dispatch, SetStateAction } from 'react'
+
+import Link from 'next/link'
+
+import { List, X } from '@phosphor-icons/react'
+import clsx from 'clsx'
+
+import { Icon } from '@/components/Icon'
+import { Logo } from '@/components/Logo'
+import { SlideOver } from '@/components/SlideOver'
+import { Social } from '@/components/Social'
+import { TextLink } from '@/components/TextLink'
+
+import { PATHS } from '@/constants/paths'
+
+const getInvolvedItems = [PATHS.EVENTS, PATHS.GRANTS]
+const communityItems = [PATHS.ECOSYSTEM, PATHS.GOVERNANCE]
+
+export function MobileNavigation() {
+  const [open, setOpen] = useState(false)
+
+  return (
+    <Fragment>
+      <button
+        className="-mr-4 flex size-12 items-center justify-center lg:hidden"
+        aria-label="Open mobile navigation"
+        onClick={() => setOpen(!open)}
+      >
+        <Icon component={List} color="brand-300" />
+      </button>
+
+      <SlideOver open={open} setOpen={setOpen}>
+        <div className="flex items-center justify-between px-6 pb-12 pt-8 sm:pb-16 md:pb-24">
+          <Link
+            className="flex-shrink-0 outline-white focus:outline-2"
+            href={PATHS.HOME.path}
+            aria-label="Go to homepage"
+            onClick={() => setOpen(false)}
+          >
+            <Logo />
+          </Link>
+
+          <button
+            className="-mr-4 flex size-12 items-center justify-center"
+            aria-label="Close mobile navigation"
+            onClick={() => setOpen(false)}
+          >
+            <Icon component={X} color="brand-300" />
+          </button>
+        </div>
+
+        <ul
+          className="px-6 pb-12 sm:pb-16 md:pb-24"
+          aria-label="Navigation options"
+        >
+          <MobileLink
+            label={PATHS.ABOUT.label}
+            path={PATHS.ABOUT.path}
+            setOpen={setOpen}
+          />
+
+          <li>
+            <p className="my-2.5 text-brand-300">Get Involved</p>
+          </li>
+          {getInvolvedItems.map((item) => (
+            <MobileLink
+              key={item.path}
+              nested
+              label={item.label}
+              path={item.path}
+              setOpen={setOpen}
+            />
+          ))}
+
+          <li>
+            <p className="my-2.5 text-brand-300">Community</p>
+          </li>
+
+          {communityItems.map((item) => (
+            <MobileLink
+              key={item.path}
+              nested
+              label={item.label}
+              path={item.path}
+              setOpen={setOpen}
+            />
+          ))}
+
+          <MobileLink
+            label={PATHS.BLOG.label}
+            path={PATHS.BLOG.path}
+            setOpen={setOpen}
+          />
+        </ul>
+
+        <div className="px-4">
+          <Social justify="left" />
+        </div>
+      </SlideOver>
+    </Fragment>
+  )
+}
+
+interface MobileLinkProps {
+  label: string
+  path: string
+  nested?: boolean
+  setOpen: Dispatch<SetStateAction<boolean>>
+}
+
+function MobileLink({ label, path, nested, setOpen }: MobileLinkProps) {
+  return (
+    <li className={clsx(nested ? 'border-l border-brand-400' : '-ml-5')}>
+      <TextLink
+        href={path}
+        className="inline-block px-5 py-2.5"
+        onClick={() => setOpen(false)}
+      >
+        {label}
+      </TextLink>
+    </li>
+  )
+}
--- a/src/app/_components/MobileNavigation.tsx
+++ b/src/app/_components/MobileNavigation.tsx
@@ -149,7 +149,7 @@ export function MobileNavigation() {
         </ul>
 
         <div className="px-4">
-          <Social justify="start" />
+          <Social />
         </div>
       </SlideOver>
     </Fragment>
--- a/src/app/_components/MobileNavigation.tsx
+++ b/src/app/_components/MobileNavigation.tsx
@@ -49,11 +49,6 @@ function MobileLink({ label, path, nested, setOpen }: MobileLinkProps) {
   )
 }
 
-const buttonTouchTarget = {
-  class: 'size-12',
-  offsetClass: '-mr-4',
-}
-
 type IconButtonProps = {
   icon: IconProps['component']
   label: string
@@ -65,9 +60,7 @@ function IconButton({ icon: IconComponent, label, onClick }: IconButtonProps) {
     <button
       aria-label={label}
       className={clsx(
-        buttonTouchTarget.class,
-        buttonTouchTarget.offsetClass,
-        'flex items-center justify-center focus:outline-2 focus:outline-brand-100 lg:hidden',
+        'flex size-12 items-center justify-center rounded-lg border border-brand-300 focus:outline-2 focus:outline-brand-100 lg:hidden',
       )}
       onClick={onClick}
     >
--- a/src/app/_components/MobileNavigation.tsx
+++ b/src/app/_components/MobileNavigation.tsx
@@ -81,7 +81,7 @@ export function MobileNavigation() {
       />
 
       <SlideOver open={open} setOpen={setOpen}>
-        <div className="flex items-center justify-between px-6 pb-12 pt-8 sm:pb-16 md:pb-24">
+        <div className="flex items-center justify-between px-6 pb-12 pt-8 sm:pb-16">
           <Link
             className="flex-shrink-0 outline-white focus:outline-2"
             href={PATHS.HOME.path}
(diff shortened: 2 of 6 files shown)


src/app/_components/MobileNavigation.tsx
```diff
@@ -0,0 +1,125 @@
...
+              label={item.label}
+              path={item.path}
+              setOpen={setOpen}
+            />
+          ))}
+
+          <li>
+            <p className="my-2.5 text-brand-300">Community</p>
+          </li>
+
```
mirhamasala, line 80:
```suggestion
```

src/app/_components/MobileNavigation.tsx
```diff
@@ -0,0 +1,125 @@
...
+          <Link
+            className="flex-shrink-0 outline-white focus:outline-2"
+            href={PATHS.HOME.path}
+            aria-label="Go to homepage"
+            onClick={() => setOpen(false)}
+          >
+            <Logo />
+          </Link>
+
+          <button
```
mirhamasala, line 45: Missing focus styles.

src/app/_components/MobileNavigation.tsx
```diff
@@ -0,0 +1,125 @@
...
+      <TextLink
+        href={path}
+        className="inline-block px-5 py-2.5"
+        onClick={() => setOpen(false)}
+      >
+        {label}
+      </TextLink>
+    </li>
+  )
+}
```
mirhamasala, line 125: We should keep this up.

src/app/_components/MobileNavigation.tsx
```diff
@@ -0,0 +1,125 @@
...
+        </div>
+      </SlideOver>
+    </Fragment>
+  )
+}
+
+interface MobileLinkProps {
+  label: string
+  path: string
+  nested?: boolean
```
mirhamasala, line 109: Should we add a default?

src/app/_components/Social.tsx
```diff
@@ -8,11 +8,16 @@ const touchTargetSpacing = 2
 const touchTargetClass = `p-${touchTargetSpacing}`
 const touchTargetOffsetClass = `-m-${touchTargetSpacing} sm:mx-0`
 
-export function Social() {
+interface SocialProps {
+  justify?: 'left' | 'between'
```
mirhamasala, line 12: Where's `left` used below?

mirhamasala (changes requested):
Looking beautiful, @CharlyMartin. 🤩

Left a couple of minor comments. And about the following:

>About the <MobileNavigation /> implementation, I've played with negative margins to make sure the link's touch targets are big enough. This is why I added inline-block px-5 py-2.5 to <TextLink />, and -mr-4 size-12 to buttons.

If there's a relationship between these, it should be clear from the code. See `Social` component:

```
const touchTargetSpacing = 2
const touchTargetClass = `p-${touchTargetSpacing}`
const touchTargetOffsetClass = `-m-${touchTargetSpacing} sm:mx-0`
``` 

This way any developer will know that these are dependencies.

src/app/_components/MobileNavigation.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  return (
+    <Fragment>
+      <button
+        className="-mr-4 flex size-12 items-center justify-center lg:hidden"
+        aria-label="Open mobile navigation"
+        onClick={() => setOpen(!open)}
+      >
+        <Icon component={List} color="brand-300" />
+      </button>
```
mirhamasala, line 32: What about refactoring this? Seems like open and close button have the same styles?

src/app/_components/MobileNavigation.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+          <MobileLink
+            label={PATHS.BLOG.label}
+            path={PATHS.BLOG.path}
+            setOpen={setOpen}
+          />
+        </ul>
+
+        <div className="px-4">
+          <Social justify="left" />
```
mirhamasala, line 99: Should this be `start` instead of `left`?

src/app/_components/Footer.tsx
```diff
@@ -72,7 +72,11 @@ export function Footer() {
       </div>
 
       <hr />
-      <Social />
+
+      <div className="sm:px-8">
```
mirhamasala, line 76: This wrapper makes me happy.

src/app/_components/MobileNavigation.tsx
```diff
@@ -0,0 +1,125 @@
...
+  label: string
+  path: string
+  nested?: boolean
+  setOpen: Dispatch<SetStateAction<boolean>>
+}
+
+function MobileLink({ label, path, nested, setOpen }: MobileLinkProps) {
+  return (
+    <li className={clsx(nested ? 'border-l border-brand-400' : '-ml-5')}>
+      <TextLink
```
mirhamasala, line 116: Instead of using `TextLink` and passing down `className` (which is something we want to move away from) - What if we just used `Link` and added the styles directly? This is quite a standalone component either way..

mirhamasala: Another thing I could think of is to add a prop to the `TextLink` - something for increasing the touch target that we could use allover the project?

src/app/_components/MobileNavigation.tsx
CharlyMartin: Ah yes! Old habits die hard 🫠

src/app/_components/MobileNavigation.tsx
CharlyMartin: To `nested`? For booleans, I don't believe it's necessary, `undefined` will always get coerced into `false`.

src/app/_components/MobileNavigation.tsx
```diff
@@ -0,0 +1,151 @@
...
+      className={clsx(
+        nested ? 'border-l border-brand-400' : touchTargetOffsetClass,
+      )}
+    >
+      <Link
+        href={path}
+        className={clsx(
+          baseStyles,
+          touchTargetClass,
+          'inline-block px-5 py-2.5',
```
mirhamasala, line 45: Is `px-5` already the `touchTargetClass`?

```suggestion
          'inline-block py-2.5',
```

src/app/_components/MobileNavigation.tsx
```diff
@@ -0,0 +1,151 @@
...
+  return (
+    <li
+      className={clsx(
+        nested ? 'border-l border-brand-400' : touchTargetOffsetClass,
+      )}
+    >
+      <Link
+        href={path}
+        className={clsx(
+          baseStyles,
```
mirhamasala, line 43: Where are `baseStyles` coming from?

src/app/_components/MobileNavigation.tsx
```diff
@@ -0,0 +1,151 @@
...
+type IconButtonProps = {
+  icon: IconType
+  label: string
+  onClick: React.ComponentPropsWithoutRef<'button'>['onClick']
+}
+
+function IconButton({ icon: IconComponent, label, onClick }: IconButtonProps) {
+  return (
+    <button
+      className="-mr-4 flex size-12 items-center justify-center focus:outline-2 focus:outline-brand-100 lg:hidden"
```
mirhamasala, line 64: Is `-mr-4` some sort of dependency?

src/app/_components/MobileNavigation.tsx
```diff
@@ -0,0 +1,151 @@
...
+        onClick={() => setOpen(false)}
+      >
+        {label}
+      </Link>
+    </li>
+  )
+}
+
+type IconButtonProps = {
+  icon: IconType
```
mirhamasala, line 56: What about using `IconProps['component']` - We could also refactor it into a type and export it `  icon?: IconProps['component']` - Since we already have this centralized?

src/app/_components/MobileNavigation.tsx
CharlyMartin: Yes, it is; I forgot to add it!

About that, **I noticed that Tailwind classes coming from variables don't work**, as in they are present in the HTML but with no corresponding CSS, as shown in the screenshot below.

My guess is that because Tailwind strips out unused classes from the CSS file at build time, classes coming from variables are not counted. I `npm run build` the app locally and inspected the final CSS output, `-ml-5` is not part of it.

Tailwind probably only keeps the classes explicitly listed in `className`? I would have to research this more, maybe there's a config to make this work.

By the way, I also checked the negative margins in `<Social />`, `-m-2` coming from `touchTargetOffsetClass`, and it's not working either. `p-2` from `touchTargetClass` does work because we must use `p-2` explicitly somewhere else, so it gets picked up by Tailwind.

CleanShot 2024-05-07 at 18 42 35

src/app/_components/MobileNavigation.tsx
CharlyMartin: `-mr-4` and `size-12` are related, but I'm not sure there's a clear correlation between them, as in, we can't just multiply by 3 to compute one from the other.

src/app/_components/MobileNavigation.tsx
CharlyMartin: Yes I like that! Let's derive from `IconProps` for now :)

src/app/_components/MobileNavigation.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+import { List, X, Icon as IconType } from '@phosphor-icons/react'
+import clsx from 'clsx'
+import { Route } from 'next'
+
+import { Icon } from '@/components/Icon'
+import { Logo } from '@/components/Logo'
+import { SlideOver } from '@/components/SlideOver'
+import { Social } from '@/components/Social'
+import { baseStyles } from '@/components/TextLink'
```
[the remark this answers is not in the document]
CharlyMartin: Yes, we can, but TBD based on my [comment above](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/116#discussion_r1592857558).

src/app/_components/MobileNavigation.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+          <MobileLink
+            label={PATHS.BLOG.label}
+            path={PATHS.BLOG.path}
+            setOpen={setOpen}
+          />
+        </ul>
+
+        <div className="px-4">
+          <Social justify="start" />
```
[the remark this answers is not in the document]
CharlyMartin: Indeed, on `xs` it's different.

We might then have to expose the `className` prop on `Social` to give more flexibility to the component's consumer.

Unless you know of a better solution?

src/app/_components/MobileNavigation.tsx
mirhamasala: Isn't `-mr-4` there because of the touchtarget offset?

src/app/_components/MobileNavigation.tsx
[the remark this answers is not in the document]
mirhamasala: I asked if she wants to mimic the styles for the Footer [>>](https://filecoinfoundation.slack.com/archives/C06MTEQ3SP6/p1715160446823629)

If not, I think we can use a `layout` prop instead and compute the styles - Maybe with options like `default` `navigation` `footer` - Whatever you think is appropriate. I prefer that over exposing the className prop (for now). 🙏🏼

src/app/_components/MobileNavigation.tsx
mirhamasala: Related >> Asking Filipa if we can make the open and close button look like buttons [>>](https://filecoinfoundation.slack.com/archives/C06MTEQ3SP6/p1715161658918759)

src/app/_components/MobileNavigation.tsx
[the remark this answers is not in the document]
mirhamasala: Filipa answered that the styles for navigation and footer can remain the same so this becomes irrelevant. 🙏🏼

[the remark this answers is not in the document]
CharlyMartin: Good to know thank you! I'm gonna revert it.

src/app/_components/MobileNavigation.tsx
CharlyMartin: Done -> https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/116/commits/32b5c87f57b5c1e3aad6cb9ce396fe2e9d1c05da

src/app/_components/MobileNavigation.tsx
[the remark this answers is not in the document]
CharlyMartin: Done -> https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/116/commits/b7ba1854b25665edfcb87edfcbcfcf1aea461100

mirhamasala (approved):
Pure Blue 💙 - Can you please check the spacing between the nav and social before merging with Figma? 🙏🏼

PR #117, opened 2024-05-07
Blog page sort

mirhamasala:
- Included `SORT_KEY`
- Validated and normalized the `sortQuery` parameter
- Refactored `sortEntriesByDate` function to support dynamic date fields
- Implemented dynamic sorting control for blog posts

## Notes:

- I'm aware that we'll have to split and extend the `SortSettings` when we include options like sorting by name, etc.
- I did not refactor `Events` page to use the `searchParams` - We can pick it up in a different PR

Screenshot 2024-05-08 at 15 43 41

### DIFF
--- a/src/app/_components/BlogSortController.tsx
+++ b/src/app/_components/BlogSortController.tsx
@@ -1,0 +1,25 @@
+'use client'
+
+import { useState, useEffect } from 'react'
+
+import { useUpdateSearchParams } from '@/hooks/useUpdateSearchParams'
+
+import {
+  type SortOptionType,
+  DEFAULT_SORT_OPTION,
+  SortListbox,
+} from '@/components/SortListbox'
+
+export function BlogSortController() {
+  const [sortOption, setSortOption] =
+    useState<SortOptionType>(DEFAULT_SORT_OPTION)
+  const updateSearchParams = useUpdateSearchParams()
+
+  useEffect(() => {
+    updateSearchParams({ sort: sortOption })
+  }, [sortOption, updateSearchParams])
+
+  return (
+    <SortListbox sortOption={sortOption} onSortOptionChange={setSortOption} />
+  )
+}
--- a/src/app/_components/BlogSortController.tsx
+++ b/src/app/_components/BlogSortController.tsx
@@ -4,11 +4,12 @@ import { useState, useEffect } from 'react'
 
 import { useUpdateSearchParams } from '@/hooks/useUpdateSearchParams'
 
-import {
-  type SortOptionType,
-  DEFAULT_SORT_OPTION,
-  SortListbox,
-} from '@/components/SortListbox'
+import { SortListbox } from '@/components/SortListbox'
+
+import { type SortOptionType } from '@/types/sortTypes'
+
+import { SORT_KEY } from '@/constants/searchParams'
+import { DEFAULT_SORT_OPTION } from '@/constants/sortConstants'
 
 export function BlogSortController() {
   const [sortOption, setSortOption] =
@@ -16,7 +17,7 @@ export function BlogSortController() {
   const updateSearchParams = useUpdateSearchParams()
 
   useEffect(() => {
-    updateSearchParams({ sort: sortOption })
+    updateSearchParams({ [SORT_KEY]: sortOption })
   }, [sortOption, updateSearchParams])
 
   return (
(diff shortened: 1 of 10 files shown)


src/app/_components/BlogSortController.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { DEFAULT_SORT_OPTION } from '@/constants/sortConstants'
+
+export function BlogSortController() {
+  const [sortOption, setSortOption] =
+    useState<SortOptionType>(DEFAULT_SORT_OPTION)
+  const updateSearchParams = useUpdateSearchParams()
+
+  useEffect(() => {
+    updateSearchParams({ [SORT_KEY]: sortOption })
+  }, [sortOption, updateSearchParams])
```
CharlyMartin, line 21: This sets the `sort` search params on page load. `Pagination` only updates the search params on the first user interaction. 

I'm not sure which one is best, but we should agree on one behaviour and normalise it for all components updating the URL.

We don't have to implement the changes now; just make sure to include them in a ticket.

PS: Search is different as there's no default value.

src/app/_components/SortListbox.tsx
```diff
@@ -1,55 +1,46 @@
 'use client'
```
CharlyMartin, line 1:
```suggestion
'use client'

```

src/app/_components/SortListbox.tsx
```diff
@@ -1,55 +1,46 @@
...
+  onSortOptionChange,
+}: SortListboxProps) {
+  const selectedSortSetting =
+    sortSettings.find((option) => option.id === sortOption) || sortSettings[0]
 
   return (
-    <Listbox value={selectedSortType} onChange={handleChange}>
+    <Listbox value={sortOption} onChange={onSortOptionChange}>
       {() => (
         <>
```
CharlyMartin, line 29:
```suggestion
```

This seems unnecessary; can we remove it?

src/app/_types/sortTypes.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+export type SortableByDate = {
+  publishedOn?: string
+  startDate?: string
+}
+
+export type SortEntriesParams<T> = {
+  entries: T[]
+  dateField: 'startDate' | 'publishedOn'
```
CharlyMartin, line 15: What about this instead?

```suggestion
  dateField: keyof SortableByDate
```

CharlyMartin (commented):
This is a solid implementation @mirhamasala, good job 👏 There's a lot going on but I'm sure I'll have a second look. It might be best to merge and refactor once we see patterns emerge.

As usual, I left a few comments, but some of them are related to minor implementation differences between pagination and sort. Whatever we choose to do, we should make it consistent across this class of components.

One of these differences lies in the component names: `BlogSearchInput` vs `BlogSortController`.

I'm not used to naming wrapper components "controllers", but why not! Or we can simplify the names as such:

`BlogSearchInput` -> `BlogSearch`
`BlogSortController` -> `BlogSort`

src/app/_components/SortListbox.tsx
mirhamasala: Yes! Leftover. 🙏🏼

src/app/_components/SortListbox.tsx
```diff
@@ -1,55 +1,46 @@
...
           <Listbox.Button className="border-1 inline-flex items-center justify-between gap-2 rounded-lg border border-brand-300 p-3 text-brand-300 hover:border-current hover:text-brand-400 focus:outline-2 focus:outline-brand-100 sm:min-w-40">
```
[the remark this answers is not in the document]
mirhamasala: Great catch! Adding the following:

```
aria-haspopup="listbox"
aria-expanded={open}
aria-label="Sort options"
```

mirhamasala:
@CharlyMartin Thank you for the review. ❤️ All changes implemented in [ad3c804](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/117/commits/ad3c8043c03a9c356ba8c260a15a9740f1409a75)

>I'm not used to naming wrapper components "controllers", but why not!

Changed to `BlogSort` in latest commit - [6ddb6eb](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/117/commits/6ddb6ebd255dd9487308f3bbdfdca3cf6a6bcef4).

PR #118, opened 2024-05-09
Desktop Navigation

CharlyMartin:
This PR adds popover menus to the navigation component on larger screens.

It relies on Headless UI's [popover](https://headlessui.com/v1/react/popover), as well as [@headlessui-tailwindcss](https://github.com/tailwindlabs/headlessui/tree/main/packages/%40headlessui-tailwindcss) for styling elements with the `ui-open` modifiers, as shown [here](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/118/commits/b5baa90ff5eeaecbae929fcf6df96883e29c8966#diff-d17350a5f2a1c03d1dcc0d3b9c8cb3e82c1ad6a3ba6a6b545f4163f13dab26a1R27)

The popover menus are keyboard accessible, meaning they can be opened/closed using tabs and Enter.

This PR also changes the label of the blog path - from `News & Blog` to `Blog` - to match the designs.

---

_Folded_
CleanShot 2024-05-13 at 11 32 58

_Unfolded_
CleanShot 2024-05-13 at 11 33 21

### DIFF
--- a/src/app/_components/DesktopNavigation.tsx
+++ b/src/app/_components/DesktopNavigation.tsx
@@ -15,7 +15,7 @@ import { NavigationPopover } from '@/components/NavigationPopover'
 import { PATHS } from '@/constants/paths'
 import { desktopNavigationItems } from '@/data/components/navigationData'
 
-export type SubNavItemProps = {
+export type SubMainNavItemProps = {
   href: string | Route
   label: string
   description?: string
@@ -28,17 +28,17 @@ type InternalLinkProps = {
   isActive?: boolean
 }
 
-function SubNavItem({
+function SubMainNavItem({
   href,
   label,
   description,
   linkType = 'internal',
-}: SubNavItemProps) {
+}: SubMainNavItemProps) {
   const external = linkType !== 'internal'
 
-  const baseClasses =
+  const baseStyles =
     'group w-full rounded-lg focus:outline focus:outline-2 focus:outline-brand-100'
-  const styleClasses = {
+  const extendedStyles = {
     internal: 'inline-block bg-brand-800 p-4 hover:bg-brand-700',
     externalPrimary:
       'inline-block border border-brand-500 bg-brand-700 p-4 hover:border-brand-400 focus:border-transparent',
@@ -46,7 +46,7 @@ function SubNavItem({
       'inline-flex items-center justify-center gap-1 bg-brand-800 px-3 py-5 text-brand-300 hover:bg-brand-700',
   }
 
-  const linkClasses = clsx(baseClasses, styleClasses[linkType])
+  const linkClasses = clsx(baseStyles, extendedStyles[linkType])
   const commonProps = {
     className: linkClasses,
     'aria-label': `${label} page (${external ? 'external link' : 'internal link'})`,
@@ -70,19 +70,22 @@ function SubNavItem({
   )
 }
 
-function getNavItemBaseStyles(isActive: boolean) {
+function getMainNavItemBaseStyles(isActive: boolean) {
   return clsx(
     'rounded-xl py-1.5 text-base hover:bg-brand-700 focus:outline focus:outline-2 focus:outline-brand-100',
     isActive ? 'text-brand-400' : 'text-brand-300',
   )
 }
 
-function NavItem({ label, href, isActive = false }: InternalLinkProps) {
+function MainNavItem({ label, href, isActive = false }: InternalLinkProps) {
   return (
     <li>
       <Link
         href={href}
-        className={clsx(getNavItemBaseStyles(isActive), 'inline-block px-4')}
+        className={clsx(
+          getMainNavItemBaseStyles(isActive),
+          'inline-block px-4',
+        )}
       >
         {label}
       </Link>
@@ -109,7 +112,7 @@ export function DesktopNavigation() {
       className="relative z-10 hidden lg:flex lg:items-center lg:gap-0.5"
       aria-label="Navigation items"
     >
-      <NavItem
+      <MainNavItem
         label={PATHS.ABOUT.label}
         href={PATHS.ABOUT.path}
         isActive={pathname === PATHS.ABOUT.path}
@@ -118,23 +121,23 @@ export function DesktopNavigation() {
       <NavigationPopover
         as="li"
         label="Get Involved"
-        navItemBaseStyles={getNavItemBaseStyles(isGetInvolvedActive)}
+        mainNavItemBaseStyles={getMainNavItemBaseStyles(isGetInvolvedActive)}
       >
         <div className="grid w-screen max-w-2xl grid-cols-2 gap-4">
           <div className="space-y-4">
             {getInvolvedInternalItems.map((item) => (
-              <SubNavItem key={item.href} {...item} linkType="internal" />
+              <SubMainNavItem key={item.href} {...item} linkType="internal" />
             ))}
           </div>
           <div className="space-y-4">
             {getInvolvedExternalItems.map((item) => (
-              <SubNavItem
+              <SubMainNavItem
                 key={item.href}
                 {...item}
                 linkType="externalPrimary"
               />
             ))}
-            <SubNavItem {...learnMoreItem} linkType="externalSecondary" />
+            <SubMainNavItem {...learnMoreItem} linkType="externalSecondary" />
           </div>
         </div>
       </NavigationPopover>
@@ -142,16 +145,16 @@ export function DesktopNavigation() {
       <NavigationPopover
         as="li"
         label="Community"
-        navItemBaseStyles={getNavItemBaseStyles(isCommunityActive)}
+        mainNavItemBaseStyles={getMainNavItemBaseStyles(isCommunityActive)}
       >
         <div className="w-80 space-y-4">
           {communityItems.map((item) => (
-            <SubNavItem key={item.label} {...item} linkType="internal" />
+            <SubMainNavItem key={item.label} {...item} linkType="internal" />
           ))}
         </div>
       </NavigationPopover>
 
-      <NavItem
+      <MainNavItem
         label={PATHS.BLOG.label}
         href={PATHS.BLOG.path}
         isActive={pathname === PATHS.BLOG.path}
--- a/src/app/_components/DesktopNavigation.tsx
+++ b/src/app/_components/DesktopNavigation.tsx
@@ -1,0 +1,181 @@
+'use client'
+
+import Link from 'next/link'
+import { usePathname } from 'next/navigation'
+
+import { ArrowUpRight } from '@phosphor-icons/react'
+import clsx from 'clsx'
+import { Route } from 'next'
+
+import { Icon } from '@/components/Icon'
+import { NavigationPopover } from '@/components/NavigationPopover'
+
+import { PATHS } from '@/constants/paths'
+
+const getInvolvedItems = [
+  {
+    label: PATHS.EVENTS.label,
+    href: PATHS.EVENTS.path,
+    description: 'Overview of upcoming and past events',
+  },
+  {
+    label: PATHS.GRANTS.label,
+    href: PATHS.GRANTS.path,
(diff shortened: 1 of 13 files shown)


src/app/_components/DesktopNavigation.tsx
```diff
@@ -0,0 +1,181 @@
...
+      <p className="text-brand-300">{description}</p>
+    </Link>
+  )
+}
+
+function ExternalLink({ label, description, href }: LinkProps) {
+  return (
+    <a
+      href={href}
+      target="_blank"
```
mirhamasala, line 82: We won't be opening external links in new windows. Will adapt `Social` to reflect this as well. 🙏🏼

```suggestion
```

src/app/_components/DesktopNavigation.tsx
```diff
@@ -0,0 +1,181 @@
...
+      rel="noopener noreferrer"
+      className="group inline-block w-full rounded-lg border border-brand-500 bg-brand-700 p-4 hover:border-brand-400 focus:border-transparent focus:outline focus:outline-2 focus:outline-brand-100"
+      aria-label={`${label} (opens a new window)`}
```
mirhamasala, line 85:
```suggestion
```

src/app/_components/NavigationPopover.tsx
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export function NavigationPopover({ label, children, as }: PopOverProps) {
+  return (
+    <Popover as={as}>
+      <Popover.Button
+        aria-label={`${label} (opens a navigation menu)`}
+        className={clsx(
+          'inline-flex items-center gap-2 rounded-xl py-1.5 pl-4 pr-3 text-base text-brand-300 hover:bg-brand-700 focus:outline focus:outline-2 focus:outline-brand-100 ui-open:bg-brand-700 ui-open:text-brand-400',
+        )}
```
mirhamasala, line 24:
```suggestion
        className="inline-flex items-center gap-2 rounded-xl py-1.5 pl-4 pr-3 text-base text-brand-300 hover:bg-brand-700 focus:outline focus:outline-2 focus:outline-brand-100 ui-open:bg-brand-700 ui-open:text-brand-400"
```

src/app/_components/DesktopNavigation.tsx
```diff
@@ -0,0 +1,181 @@
...
+            <InternalLink key={item.label} {...item} />
+          ))}
+        </div>
+      </NavigationPopover>
+
+      <li>
+        <Link
+          href={PATHS.BLOG.path}
+          className={clsx(
+            'inline-block rounded-xl px-4 py-1.5 text-brand-300 hover:bg-brand-700 focus:outline focus:outline-2 focus:outline-brand-100',
```
mirhamasala, line 170:
```suggestion
            'inline-block rounded-xl px-4 py-1.5 hover:bg-brand-700 focus:outline focus:outline-2 focus:outline-brand-100',
```

src/app/_components/DesktopNavigation.tsx
```diff
@@ -0,0 +1,181 @@
...
+  description: string
+  href: string
+}
+
+function InternalLink({ label, description, href }: LinkProps) {
+  return (
+    <Link
+      href={href as Route}
+      className="inline-block w-full rounded-lg bg-brand-800 p-4 hover:bg-brand-700 focus:outline focus:outline-2 focus:outline-brand-100"
+      aria-label={`${label} page (internal link)`}
```
mirhamasala, line 70: Is an `aria-label` necessary when the link text is clear?

src/app/_components/DesktopNavigation.tsx
```diff
@@ -0,0 +1,181 @@
...
+  )
+}
+
+const getInvolvedInternalItems = getInvolvedItems.filter((item) =>
+  item.href.startsWith('/'),
+)
+
+const getInvolvedExternalItems = getInvolvedItems.filter(
+  (item) => item.href.startsWith('https://') || item.href.startsWith('http://'),
+)
```
mirhamasala, line 105: in `CustomLink` we have:

```
  const isInternalLink = href.startsWith('/') || href.startsWith('#')
```

src/app/_components/DesktopNavigation.tsx
```diff
@@ -0,0 +1,181 @@
...
+  return (
+    <a
+      href={href}
+      target="_blank"
+      rel="noopener noreferrer"
+      className="group inline-block w-full rounded-lg border border-brand-500 bg-brand-700 p-4 hover:border-brand-400 focus:border-transparent focus:outline focus:outline-2 focus:outline-brand-100"
+      aria-label={`${label} (opens a new window)`}
+    >
+      <div className="flex gap-1">
+        <p className="pb-2 font-bold">{label}</p>
```
mirhamasala, line 88: Nit: We tend to use margin instead of padding. Also - Looking at the designs - it should be `mb-4`

mirhamasala: Also I think we need to move this to the parent - Otherwise, the alignment of the icon isn't right.

src/app/_components/DesktopNavigation.tsx
```diff
@@ -0,0 +1,181 @@
...
+    label: 'Contribute to the Filecoin Project',
+    href: 'https://docs.filecoin.io/basics/project-and-community/ways-to-contribute',
+    description:
+      'Shape the future of Filecoin by contributing to its code, research, or docs.',
+  },
+]
+
+const communityItems = [
+  {
+    label: PATHS.ECOSYSTEM.label,
```
mirhamasala, line 42: Should be `Ecosystem Explorer`

src/app/_components/DesktopNavigation.tsx
mirhamasala: And on that note - Did you look into using `CustomLink`?

src/app/_components/DesktopNavigation.tsx
```diff
@@ -0,0 +1,181 @@
...
+const getInvolvedExternalItems = getInvolvedItems.filter(
+  (item) => item.href.startsWith('https://') || item.href.startsWith('http://'),
+)
+
+export function DesktopNavigation() {
+  const pathname = usePathname()
+
+  return (
+    <ul
+      className="relative z-10 hidden items-center gap-0.5 lg:flex"
```
mirhamasala, line 112: Could we use just `flex`?

src/app/_components/DesktopNavigation.tsx
```diff
@@ -0,0 +1,181 @@
...
+    >
+      <li>
+        <Link
+          href={PATHS.ABOUT.path}
+          className={clsx(
+            'text-hover:bg-brand-700 inline-block rounded-xl px-4 py-1.5 focus:outline focus:outline-2 focus:outline-brand-100',
+            pathname.startsWith(PATHS.ABOUT.path)
+              ? 'text-brand-400'
+              : 'text-brand-300',
+          )}
```
mirhamasala, line 123: Why do we need this check here?

mirhamasala: Same for blog below.

src/app/_components/DesktopNavigation.tsx
CharlyMartin: I think so because the `Link` has two descendants, label and description, which get concatenated into one string in the accessibility tree. So the label makes it clearer hopefully.

src/app/_components/NavigationPopover.tsx
```diff
@@ -0,0 +1,59 @@
...
+  children: React.ReactElement
+  as: React.ElementType
+}
+
+export function NavigationPopover({ label, children, as }: PopOverProps) {
+  return (
+    <Popover as={as}>
+      <Popover.Button
+        aria-label={`${label} (opens a navigation menu)`}
+        className={clsx(
```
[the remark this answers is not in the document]
CharlyMartin: It is not! I removed it.

src/app/_components/DesktopNavigation.tsx
CharlyMartin: I did look into `CustomLink`, but I feel that the markup here is a bit too different to merge `InternalLink` and `ExternalLink` into one component.

src/app/_components/DesktopNavigation.tsx
mirhamasala: Because we're setting it below:

```
pathname.startsWith(PATHS.BLOG.path)
              ? 'text-brand-400'
              : 'text-brand-300',
```

But this may become irrelevant [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/118/files#r1599913830)

src/app/_components/DesktopNavigation.tsx
CharlyMartin: Should we update it in the `PATHS` object?

src/app/_components/DesktopNavigation.tsx
CharlyMartin: But we need `lg:flex` for `hidden` to kick in on smaller screens.

src/app/_components/DesktopNavigation.tsx
```diff
@@ -0,0 +1,181 @@
...
+  return (
+    <ul
+      className="relative z-10 hidden items-center gap-0.5 lg:flex"
+      aria-label="Navigation items"
+    >
+      <li>
+        <Link
+          href={PATHS.ABOUT.path}
+          className={clsx(
+            'text-hover:bg-brand-700 inline-block rounded-xl px-4 py-1.5 focus:outline focus:outline-2 focus:outline-brand-100',
```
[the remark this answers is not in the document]
CharlyMartin: It definitely does NOT 🫠

src/app/_components/DesktopNavigation.tsx
mirhamasala: Ah, I see! I was thrown off by `items-center gap-0.5` which aren't prefixed.

src/app/_components/DesktopNavigation.tsx
```diff
@@ -95,11 +97,11 @@ function ExternalLink({ label, description, href }: LinkProps) {
...
 const getInvolvedInternalItems = getInvolvedItems.filter((item) =>
-  item.href.startsWith('/'),
+  isInternalLink(item.href),
 )
 
-const getInvolvedExternalItems = getInvolvedItems.filter(
-  (item) => item.href.startsWith('https://') || item.href.startsWith('http://'),
+const getInvolvedExternalItems = getInvolvedItems.filter((item) =>
+  isExternalLink(item.href),
 )
```
mirhamasala: Wouldn't this be `!isInternLink` ?

CharlyMartin: Would you prefer `!isInternLink` over `isExternalLink`?

src/app/_components/DesktopNavigation.tsx
CharlyMartin: This is to highlight the text if the link is active, as per the [Figma designs](https://www.figma.com/design/0pBfQs5tKI6bx5ypPIfVRA/FIL.org-V4-Prototypes?node-id=1960-37512&t=FlmiDdbhUC1ODsgK-4)

CharlyMartin: I've abstracted the active styles

src/app/_components/DesktopNavigation.tsx
mirhamasala: The current implementation of  `isExternalLink` seems a bit basic and limited? What if a URL starts with `www`?

```
export function isExternalLink(href: Link) {
  return href.startsWith('http://') || href.startsWith('https://')
}
```

If `isInternalLink` already exists - but you still wanted to be more explicit - You could always do:

```
export function isExternalLink(href: string) {
  return !isInternalLink(href)
}
```

Perhaps it helps with readability - I'm fine with both.

PR #127, opened 2024-05-14
Events Search

CharlyMartin:
This PR adds search capabilities to the events page. Go to http://localhost:3000/events to see it live.

### About Search
My first instinct was to create an `EventsSearchInput` wrapper component for the events page, as we have `BlogSearchInput` for the blog.

However, I quickly noticed that `BlogSearchInput` can be used on the events page as is. Since we use URL params as a global state, the component can be used on any page out of the box. It only needs to receive a prop from `useSearch`, that I created for that purpose.

`useSearch` also encapsulates filtering data based on the search query so we avoid duplication in each page component. It looks like this:
```tsx
// Hook
const { searchQuery, searchResults } = useSearch({
  searchParams,
  data: events,
})

// Search input
<BlogSearchInput query={searchQuery} /> // To be renamed

// Card list
{searchResults.map((event) => (
  <Card key={event.slug} title={event.title} />
))}
```

`BlogSearchInput` will need to be renamed to make it more agnostic; how about just `Search`? (I haven't changed the name yet to keep the diff clean)

### About Sort

I've deprecated `EventsClient` and left the sort component out of this PR, for now. I can either add it as part of this PR or in a subsequent one.

But basically, `BlogSort` can receive the same treatment. It can be renamed as `Sort` and receive only the relevant `SortOptionType` as a prop/type.


src/app/_components/BlogSearchInput.tsx
```diff
@@ -2,30 +2,48 @@
...
+import { PAGE_KEY, SEARCH_KEY } from '@/constants/searchParams'
+
 type BlogSearchInputProps = {
-  searchQuery: string
+  query: ReturnType<typeof useSearch>['searchQuery']
 }
 
-export function BlogSearchInput({ searchQuery }: BlogSearchInputProps) {
-  const [value, setValue] = useState(searchQuery)
+const TYPE_DEBOUNCE_DELAY = 400
```
mirhamasala, line 20: Why is this prefixed with `TYPE` - What does that mean in this context?

src/app/_hooks/useSearch.ts
```diff
@@ -0,0 +1,36 @@
+import { useMemo } from 'react'
+
+import { NextServerSearchParams } from '@/types/searchParams'
+import { SearchableByTitle } from '@/types/searchTypes'
```
mirhamasala, line 4: Do we want to only to search by `title` - What about the `description` text, for example?

src/app/_hooks/useSearch.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import { SEARCH_KEY } from '@/constants/searchParams'
+
+function validateSearchQuery(searchQuery: NextServerSearchParams['search']) {
+  if (!searchQuery) {
+    return ''
+  }
+
+  const rawQuery = Array.isArray(searchQuery) ? searchQuery[0] : searchQuery
```
mirhamasala, line 13: Nit: If `rawQuery` is already `const rawQuery = searchParams[SEARCH_KEY]` - Then maybe this can be `rawQueryValue`?

src/app/_hooks/useSearch.ts
```diff
@@ -0,0 +1,36 @@
...
+  return rawQuery.toLowerCase()
+}
+
+type UseSearchProps<Item extends SearchableByTitle> = {
+  searchParams: NextServerSearchParams
+  data: Item[]
+}
```
mirhamasala, line 20: Shall we move it up? I know it's closer like this - Just trying to be consistent for now.

src/app/_hooks/useSearch.ts
```diff
@@ -0,0 +1,36 @@
+import { useMemo } from 'react'
+
+import { NextServerSearchParams } from '@/types/searchParams'
+import { SearchableByTitle } from '@/types/searchTypes'
+
+import { SEARCH_KEY } from '@/constants/searchParams'
+
+function validateSearchQuery(searchQuery: NextServerSearchParams['search']) {
```
mirhamasala, line 8: Can we have a look at `validateSortOptionQuery` - and see if we can abstract some of the logic?

src/app/_hooks/useSearch.ts
mirhamasala: Also curious - Is there a preference for `Item` over `T`?

src/app/_components/BlogSearchInput.tsx
```diff
@@ -2,30 +2,48 @@
...
   )
 
   function handleSearchChange(event: React.ChangeEvent<HTMLInputElement>) {
     const newValue = event.target.value
     setValue(newValue)
-    debouncedUpdateSearchParams({ search: newValue, page: '1' })
+
+    const existingPageParam = params.get(PAGE_KEY)
+
+    if (existingPageParam && existingPageParam !== '1') {
```
mirhamasala, line 39: What is `1` the default, the first page?

src/app/_components/BlogSearchInput.tsx
```diff
@@ -2,30 +2,48 @@
...
+      debouncedUpdateSearchParams({
+        [SEARCH_KEY]: newValue,
+        [PAGE_KEY]: '1',
+      })
+    } else {
+      debouncedUpdateSearchParams({ [SEARCH_KEY]: newValue })
+    }
   }
```
mirhamasala, line 47: I'm guessing we're going to want to refactor this into a custom `useDebounce` hook? Not a blocker.

src/app/_components/BlogSearchInput.tsx
CharlyMartin: It's type as in typing on a keyboard, but I'm realising it's not exactly right. Maybe `DEBOUNCE_DELAY` alone would be better?

src/app/_hooks/useSearch.ts
CharlyMartin: For now, it's only by title, but we can change `SearchableByTitle` to the following if necessary.

```
export type SearchableByTitle = {
  title: string
  description: string
}
```

(And also rename it to match the changes)

src/app/_hooks/useSearch.ts
```diff
@@ -0,0 +1,36 @@
...
+  searchParams: NextServerSearchParams
+  data: Item[]
+}
+
+export function useSearch<Item extends SearchableByTitle>({
+  searchParams,
+  data,
+}: UseSearchProps<Item>) {
+  const rawQuery = searchParams[SEARCH_KEY]
+  const cleanQuery = validateSearchQuery(rawQuery)
```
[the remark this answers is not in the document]
CharlyMartin: Yes, I agree. This will be part of the refactoring process.

src/app/_hooks/useSearch.ts
CharlyMartin: `Item` feels slightly less cryptic than `T`?

src/app/_hooks/useSearch.ts
CharlyMartin: Hmm, maybe we can abstract the following? Not sure if it's adding a lot of value.

```ts
const rawQuery = Array.isArray(searchQuery) ? searchQuery[0] : searchQuery
return rawQuery.toLowerCase()
```

src/app/_components/BlogSearchInput.tsx
CharlyMartin: Ah yes! `1` is the first page. Basically, we want to reset the pagination to the first page when users search to avoid weird behaviours.

Maybe we can `const FIRST_PAGE = 1`

src/app/_components/BlogSearchInput.tsx
CharlyMartin: How would you want to refactor this?

src/app/_components/BlogSearchInput.tsx
mirhamasala: Agree! `DEBOUNCE_DELAY` sounds good to me.

src/app/_components/BlogSearchInput.tsx
mirhamasala: Yeah - We have `DEFAULT_PAGE_NUMBER` in `validatePageNumber` - Could that be used?

This is what I did for `sortConstants`

```
import { type SortOptionType, type SortSetting } from '@/types/sortTypes'

export const DEFAULT_SORT_OPTION: SortOptionType = 'newest'

export const sortSettings: SortSetting[] = [
  { id: 'newest', name: 'Newest' },
  { id: 'oldest', name: 'Oldest' },
] as const

export const VALID_SORT_OPTIONS = sortSettings.map((setting) => setting.id)
```

I know it's unlikely we will ever change the default page to anything but `1` - and yet, it feels good to have this in one place - What do you think?

src/app/_components/BlogSearchInput.tsx
mirhamasala: Ah, I think I falsely assumed we were going to need something similar for sort and categories. But we're not going to need to debounce those. 🙏🏼

src/app/_hooks/useSearch.ts
mirhamasala: What about something like `queryUtils.ts` - This is a rough draft - We can create types instead of having `key: string` etc.

```
const getFirstQueryParam = (params: NextServerSearchParams, key: string): string | undefined => {
  const param = params[key];
  return Array.isArray(param) ? param[0] : param;
};

const normalizeQueryParam = (params: NextServerSearchParams, key: string): string | undefined => {
  const queryValue = getFirstQueryParam(params, key);
  return queryValue ? queryValue.toLowerCase() : undefined;
};

```

I'm saying this because we actually also use this for `usePagination`

```
  if (Array.isArray(pageQuery)) {
    currentPage = validatePageNumber(pageQuery[0], pageCount)
  } else {
    currentPage = validatePageNumber(pageQuery, pageCount)
  }
```

And we will be using it categories as well, I assume. That means, at least, 4 places for now. Seems worth abstracting?

src/app/_hooks/useSearch.ts
mirhamasala: Yes! I think it would make sense to have `SearchableByTitle` for Events (and later on perhaps location?); `SearchableByTitleAndDescription` for Blog Posts; and `SearchableByTitleAndDescriptionAndTopic` for Ecosystem Projects. Not sure about those names - but you get the gist :-)

src/app/_hooks/useSearch.ts
mirhamasala: It is - and I think that's why it makes me stop and think - "What `Item` - and `Item` of what?" ... which isn't necessarily bad. Can we agree on `Entry`?

src/app/_hooks/useSearch.ts
CharlyMartin: Yes I see!

src/app/_components/BlogSearchInput.tsx
CharlyMartin: I see :) We just need debounce for search and pagination.

src/app/_components/BlogSearchInput.tsx
CharlyMartin: Yes let's use the existing `DEFAULT_PAGE_NUMBER`. We will never change it to anything other than "1" but it reads better and feels cleaner :)

src/app/_components/BlogSearchInput.tsx
CharlyMartin: ✅

src/app/_hooks/useSearch.ts
CharlyMartin: Instead of using `SearchableByTitle` and the likes, I've added a `searchBy` prop to `useSearch` so that we can list the searchable fields there. And it's type-safe 😎

```ts
// Blog
const { searchQuery, searchResults } = useSearch({
  searchParams,
  entries: posts,
  searchBy: ['title', 'description'],
})

// Events
const { searchQuery, searchResults } = useSearch({
  searchParams,
  entries: events,
  searchBy: ['title', 'location'],
})
```

src/app/_hooks/useSearch.ts
CharlyMartin: ✅

src/app/_hooks/useSearch.ts
CharlyMartin: ✅

src/app/_components/Search.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+import { useSearch } from '@/hooks/useSearch'
+import { useUpdateSearchParams } from '@/hooks/useUpdateSearchParams'
+
+import { SearchInput } from '@/components/SearchInput'
+
+import { DEFAULT_PAGE_NUMBER } from '@/constants/paginationConstants'
+import { PAGE_KEY, SEARCH_KEY } from '@/constants/searchParams'
+
+type BlogSearchInputProps = {
```
mirhamasala, line 17:
```suggestion
type SearchProps = {
```

src/app/_components/Search.tsx
```diff
@@ -0,0 +1,59 @@
...
+  query: ReturnType<typeof useSearch>['searchQuery']
+}
+
+const DEBOUNCE_DELAY = 400
+
+export function Search({ query }: BlogSearchInputProps) {
```
mirhamasala, line 23:
```suggestion
export function Search({ query }: SearchProps) {
```

src/app/_hooks/useSearch.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import { SEARCH_KEY } from '@/constants/searchParams'
+
+type UseSearchProps<Entry extends Record<string, unknown>> = {
+  searchParams: NextServerSearchParams
+  entries: Entry[]
+  searchBy: keyof Entry | Array<keyof Entry>
+}
+
+function search<Entry extends Record<string, unknown>>(
```
mirhamasala, line 15: Would this function be more explicit as `matchesQuery` or something like that?

src/app/_utils/queryUtils.ts
```diff
@@ -0,0 +1,14 @@
...
+  const param = params[key]
+  return Array.isArray(param) ? param[0] : param
+}
+
+export function normalizeQueryParam(
+  params: NextServerSearchParams,
+  key: string,
+) {
+  const param = getFirstQueryParam(params, key)
+  return param?.toLowerCase()
```
mirhamasala, line 13: Maybe >>

```suggestion
  return param?.trim().toLowerCase()
```

mirhamasala (approved):
Lovely! ~minor leftover naming refactor and a `trim()` suggestion. 🙏🏼 #FISI

src/app/_hooks/useSearch.ts
CharlyMartin: Yes, it would :)

PR #129, opened 2024-05-15
Ecosystem Entry Page - TextLink with Icons

barbaraperic:
## Changes
- `TextLink` component refactored to accept 2 new props: `iconType` and `isExternalPage`
- this also includes a FIX on having the icons displayed in 1 column on mobile

## Visuals
Screenshot 2024-05-15 at 11 08 51


src/app/_components/TextLink.tsx
```diff
@@ -1,26 +1,48 @@
...
 import { CustomLink } from '@/components/CustomLink'
 
+import { Icon } from './Icon'
+
 type TextLinkProps = {
   href: string
   className?: string
   children: React.ReactNode
+  iconType?: 'caseStudy' | 'repo' | 'twitter' | 'website'
+  isExternalPage?: boolean
```
mirhamasala, line 19: This means we would have to "manually" pass this prop for every `externalPage`

barbaraperic: Exactly. Is that an issue?

mirhamasala: It would be nicer if we could think of a solution where that is not necessary. Perhaps we can leverage the logic in `CustomLink` somehow ..

mirhamasala: Also happy if you just make this work for the Ecosystem Page without thinking of the wider refactor.

mirhamasala:
Closing: Not relevant at the moment. We will pick up the general handling of external links with an arrow out in another PR.

PR #130, opened 2024-05-16
Include Ecosystem Projects Categories and Subcategories

mirhamasala:
This PR sets up the new categories and subcategories structure for Ecosystem Projects:

- Rename old `topics` to `categories` to align with same fields for blog posts and events
- Include `subcategories`
- Move ecosystem projects under `ecosystem/projects` for better organization

**Extra:**

- Include `customEntriesContentPath` for `createPathObject` 
- Refactor to use object for parameters for readability

**Note:**

- When setting `subcategories` for a project, we would like to only show the `subcategories` pertaining to chosen parent category. Unfortunately, Decap CMS doesn't allow filtering in the same "form". We'll need to look into creating a custom widget.


mirhamasala:
@barbaraperic @CharlyMartin - Tagging for visibility. No need for a review, but please have a look at `paths.ts` and `config.yml`. 🙏🏼

PR #134, opened 2024-05-16
Events Sort

CharlyMartin:
It only works with dates as of now, it will need to be tweaked for strings if we need to (alphabetical order)

To be renamed:
- `BlogSort` -> `Sort`
- `useSortQuery` -> `useSort`

To be deleted:
- `_[deprecated]_EventsClient.tsx`


src/app/_components/SearchInput.tsx
```diff
@@ -1,20 +1,23 @@
...
 }
 
 export function SearchInput({
   searchQuery = '',
   onSearchChange,
+  width = 'xs',
 }: SearchInputProps) {
   return (
-    <div className="w-full max-w-lg lg:max-w-xs">
+    <div className={clsx('w-full', width === 'xs' && 'max-w-lg lg:max-w-xs')}>
```
mirhamasala, line 20: To me, this feels like "reaching into the component" - Layout styles usually can be handled from the parent. I'm going to remove this for now. It will still be available as a solution in the PR history, so we can always pull it back if needed - though I think it might not be.

mirhamasala:
@CharlyMartin This is looking really awesome. 🤩 🙌🏼 

I finished the review and made two changes before getting this merged.

- Removed the layout styles for now [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/134#discussion_r1604408776)
- Fixed the `useEffect` [a7e6967](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/134/commits/a7e69673d6ebdd91f295ec45224aa56c7956a3c6)


~Let me know if you don't get warnings about missing dependencies - We should look into it if that's the case.

About the fix:

- Memoize `updateSearchParams` with `useCallback` to maintain function stability
- Include necessary dependencies in `useEffect` to ensure correct re-runs

### The Problem

- **Dependencies in `useEffect`**: `useEffect` needs all variables and functions it depends on. Missing dependencies can cause React to not re-run the effect when necessary.
- **Function Stability**: Functions like `updateSearchParams` are re-created on every render, making them unstable and causing `useEffect` to re-run unnecessarily.

### Specific Issues

1. **Unstable `updateSearchParams`**: The function was re-created on every render, causing `useEffect` to think it needed to re-run.
2. **Missing Dependencies Warning**: React’s ESLint plugin warned that `query` and `updateSearchParams` should be included in the dependency array.

### The Solution

1. **Memoizing `updateSearchParams`**:
   ```javascript
   const updateSearchParams = useCallback((updatedParams: Record<string, any>) => {
     const newParams = buildSearchParams(updatedParams, params);
     const newRoute = `${pathname}?${newParams}` as Route;
     router.push(newRoute, { scroll: false });
   }, [router, pathname, params]);
   ```
   This ensures `updateSearchParams` is stable and doesn’t change on every render.

2. **Including Dependencies**:
   ```javascript
   useEffect(() => {
     updateSearchParams({ [SORT_KEY]: query });
   }, [query, updateSearchParams]);
   ```
   This ensures `useEffect` runs correctly whenever `query` or `updateSearchParams` changes.

### Summary

- **Stability**: Using `useCallback` for `updateSearchParams` prevents unnecessary re-renders.
- **Complete Dependencies**: Including all necessary dependencies in `useEffect` ensures it behaves correctly.

src/app/_components/SearchInput.tsx
CharlyMartin: In that case, we need to remove `max-w-lg lg:max-w-xs` from this component so that it takes the full width of the parent's container.

PR #143, opened 2024-05-21
Blog Post - remove inline style from markdown

barbaraperic:
## Changes
- removed inline styles from markdown files 
- added 3 components (`figure`, `figcaption`, `iframe`) in `MarkdownContent` in order to pass `className` values that correspond to removed inline styles

## Note
- I've noticed that page `blog/what’s-new-with-the-filecoin-plus-notary-election-and-filecoin-day-highlights-from-labweek23` returns 404 - haven't looked much into it, happy to open a new ticket for this


mirhamasala:
@barbaraperic We're going to need to remove inline styles from all blog posts. And make sure that whatever needs styling is done through `MarkdownContent`

mirhamasala (changes requested):
@barbaraperic We're going to need to remove inline styles from all blog posts. And make sure that whatever needs styling is done through MarkdownContent

src/app/_components/MarkdownContent.tsx
```diff
@@ -16,6 +16,22 @@ export function MarkdownContent({ children }: MarkdownContentProps) {
...
+          <figcaption
+            {...props}
+            className="overflow-hidden overflow-ellipsis whitespace-nowrap break-words text-left font-sans text-[10px] font-thin"
+          >
+            {props.children}
+          </figcaption>
+        ),
+        iframe: ({ node, ...props }) => (
+          <iframe {...props} className="border-0" />
+        ),
```
CharlyMartin, line 34: Why not add the style of these 3 elements in `tailwind.config.js`, to stay consistent with this existing setup?

In the case of `img` just above, we have to render `CustomImage` (a custom component), so we need this extra configuration. With `figure`, `figcaption` and `iframe`, we're only adding style to the existing HTML markup so we don't need to recreate it. Unless I'm missing something? Does that make sense?

CharlyMartin (commented):
I would move all the style to `tailwind.config.js`, like you're already doing below :)

```
figcaption: {
  a: {
    color: '#cccccc',
    textDecoration: 'none',
  },
},
```

src/app/_components/MarkdownContent.tsx
barbaraperic: Reason why I included it is to make it self-contained, so the styles only apply when these elements are used in the markdown context. But since we already have `figcaption` in tailwind, it makes sense to move all of those there.  👍

src/content/blog/what’s-new-with-the-filecoin-plus-notary-election-and-filecoin-day-highlights-from-labweek23.md
```diff
@@ -84,7 +83,7 @@ FF Co-Founding Officer Clara Tsao discussed why the future of AI and enterprise
 
 **Crossing the Streams: A Ghostbusters-themed Approach to Scaling the Filecoin Ecosystem**
 
-<div style="text-align: center;">
+<figure>
```
mirhamasala, line 86: If there's not `figcaption` - There's no need to wrap the `iframe` in such a tag.

tailwind.config.js
```diff
@@ -63,6 +63,13 @@ module.exports = {
               height: 'auto',
               aspectRatio: '16/9',
             },
+
+            figcaption: {
+              a: {
+                color: '#cccccc',
+                textDecoration: 'none',
```
mirhamasala, line 70: is this necessary? Seems like we're already defining it here:

```
            a: {
              textDecoration: 'none',
              '&:hover': {
                textDecoration: 'underline',
              },
              '&:focus, &:focus-visible': {
                outlineWidth: 2,
                outlineStyle: 'solid',
                outlineColor: theme('colors.brand.100'),
              },
            },
            ```

File: tailwind.config.js [hunk 4]
```diff
@@ -63,6 +63,13 @@ module.exports = {
               height: 'auto',
               aspectRatio: '16/9',
             },
+
+            figcaption: {
+              a: {
+                color: '#cccccc',
```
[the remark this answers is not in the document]
mirhamasala: Also do we need to change this? [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/143/files#r1615770850)

mirhamasala:
@barbaraperic @CharlyMartin - The current setup would require the content team to know to wrap an iframe into a `figure` tag whenever using a `figcaption`. We won't be able to get away without any sort of training .. but there is this other option that I wanted to talk to you about. This is a rough example. Let me know what you think.


```
import React from 'react';
import ReactMarkdown from 'react-markdown';
import rehypeRaw from 'rehype-raw';
import { CustomImage } from '@/components/CustomImage';

type MarkdownContentProps = {
  children: string;
};

const extractCaption = (node: any) => {
  const nextSibling = node?.parent?.children?.[node?.parent?.children?.indexOf(node) + 1];
  if (nextSibling && nextSibling.type === 'comment' && nextSibling.value.startsWith('caption:')) {
    return nextSibling.value.replace('caption:', '').trim();
  }
  return null;
};

export function MarkdownContent({ children }: MarkdownContentProps) {
  return (
    <ReactMarkdown
      rehypePlugins={[rehypeRaw]}
      className="prose"
      components={{
        img: ({ node, ...props }) => (
          <CustomImage src={props.src as string} alt={props.alt as string} />
        ),
        iframe: ({ node, ...props }) => {
          const caption = extractCaption(node);
          return caption ? (
            <figure>
              <iframe {...props} />
              <figcaption>{caption}</figcaption>
            </figure>
          ) : (
            <iframe {...props} />
          );
        },
      }}
    >
      {children}
    </ReactMarkdown>
  );
}
```

Used liked this in the markdown:

```
 <iframe
    width="560"
    height="315"
    src="https://www.youtube.com/embed/57Ua4AEWe_g?si=nFb8apksN6EgLB_w"
    title="YouTube video player"
    allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture"
    allowfullscreen
  ></iframe>
  <!--caption:This is a caption describing the embedded YouTube video.-->
`;
```

tailwind.config.js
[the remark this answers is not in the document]
barbaraperic: We don't have `#cccccc` as our brand color?

src/app/_components/MarkdownContent.tsx
CharlyMartin: I believe the styles we define under `typography` in the `tailwind.config` are only applied when `className="prose"` is applied, which is the case on `MarkdownContent`. So we're safe on that front :)

https://github.com/tailwindlabs/tailwindcss-typography?tab=readme-ov-file#customizing-the-css

tailwind.config.js
```diff
@@ -56,13 +56,34 @@ module.exports = {
 
             iframe: {
               width: '100%',
+              borderWidth: '0',
```
mirhamasala, line 59: For all - Or just YouTube?

tailwind.config.js
```diff
@@ -56,13 +56,34 @@ module.exports = {
...
+            },
+
+            figcaption: {
+              lineBreak: 'anywhere',
+              wordBreak: 'normal',
+              overflow: 'hidden',
+              whiteSpace: 'nowrap',
+              textOverflow: 'ellipsis',
+              textAlign: 'left',
+              fontFamily:
```
mirhamasala, line 79: Let's not change the `fontFamily`

tailwind.config.js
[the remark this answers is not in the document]
mirhamasala: Remember that these inline styles were randomly added by the content team. We want to refactor and streamline these to align better with the current design of the site. We can change anything and everything we need.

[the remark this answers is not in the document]
mirhamasala: Also - why don't we just remove it all together and it inherit the `a` styles we set?

barbaraperic:
@mirhamasala @CharlyMartin let me know what you think about the new video component for Decap. There's a test blog post named "This is a test" where you can test how it works. 


https://github.com/FilecoinFoundationWeb/filecoin-foundation/assets/50910606/7a945a40-8a19-4b13-867f-2728fe8f74d8

barbaraperic:
> @mirhamasala @CharlyMartin let me know what you think about the new video component for Decap. There's a test blog post named "This is a test" where you can test how it works.
> 
>  CleanShot.2024-05-28.at.12.24.49.mp4

Sorry realised that we might want to have figcaptions as links, I'll amend.

mirhamasala:
>I've noticed that page blog/what’s-new-with-the-filecoin-plus-notary-election-and-filecoin-day-highlights-from-labweek23 returns 404 - haven't looked much into it, happy to open a new ticket for this

@barbaraperic I've fixed all the slugs with an `’` in them in #164  - That was the reason they were breaking.

CharlyMartin:
@barbaraperic I love the work you did on the video template, super cool 👏 

I've refactored it a bit to make it clearer, to me at least, with this commit https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/143/commits/8fbc73c346ce156f56a082954fa8c39555806fc4. Does it make sense to you too?

Also two things:
1. The `required: true` validation on `Embedded Video URL` doesn't work, meaning I can publish the article with an empty video URL. So I included this edge case in `renderVideoComponent`
2. I moved the `captionTexts` above the `captionLinks` as I believe it makes more sense for users. You can have texts without links but not the other way around, right?

barbaraperic:
> @barbaraperic I love the work you did on the video template, super cool 👏
> 
> I've refactored it a bit to make it clearer, to me at least, with this commit [8fbc73c](https://github.com/FilecoinFoundationWeb/filecoin-foundation/commit/8fbc73c346ce156f56a082954fa8c39555806fc4). Does it make sense to you too?
> 
> Also two things:
> 
> 1. The `required: true` validation on `Embedded Video URL` doesn't work, meaning I can publish the article with an empty video URL. So I included this edge case in `renderVideoComponent`
> 2. I moved the `captionTexts` above the `captionLinks` as I believe it makes more sense for users. You can have texts without links but not the other way around, right?

Love the refactor! Works well for me 👍

CharlyMartin (approved):

mirhamasala (approved):

tailwind.config.js
```diff
@@ -63,6 +64,20 @@ module.exports = {
...
+            figure: {
+              textAlign: 'center',
+            },
+
+            figcaption: {
+              lineBreak: 'anywhere',
+              wordBreak: 'normal',
+              overflow: 'hidden',
+              whiteSpace: 'nowrap',
+              textOverflow: 'ellipsis',
```
mirhamasala, line 77: Why do we use an ellipsis to hide the `figcaption` partially and is there a way to see it completely?

barbaraperic: If we don't want the size of the container to cut the paragraph, we can remove the `whiteSpace` property. Otherwise it's truncated and `ellipsis` property will add "...". Perhaps the better option would be, as you suggested, to have it fully readable across all screen devices.

mirhamasala (approved):
Stellar @barbaraperic 🙌🏼🔥 - LGTM!

PR #145, opened 2024-05-21
Update navigation design

CharlyMartin:
This PR updates the style of mobile navigation sub-items, as well as fixes the alignment with the social icons.

CleanShot 2024-05-21 at 15 45 44@2x

Regarding the desktop alignment, I only updated the breakpoint to match the mobile / desktop navigation elements, but I didn't change the current alignment as it looks correct. What did you have in mind?

CleanShot 2024-05-21 at 15 52 45@2x


mirhamasala (changes requested):
- Please check the spacing of elements
Screenshot 2024-05-21 at 16 31 35

- As for the alignment of desktop nav. The last nav item, in this case `Blog` has a `padding-right` - That offset isn't accounted for, making it look like the end of the nav item isn't aligned with the main page layout. In the picture below, notice how `Featured Content` aligns to the end of the main page layout while the last nav item doesn't.

Screenshot 2024-05-21 at 16 35 45

CharlyMartin:
**About desktop** - thanks for clarifying, I fixed it with this commit https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/145/commits/a77d11f22740ef7ca2b7fb4c865e1666a963ed76

CleanShot 2024-05-21 at 18 23 00@2x

---

**About mobile** - the gap does look smaller on Figma, but we have to be mindful of the touch target area, which is currently `44px` high for each link and sublink (as opposed to the standard `48px`).

At the moment, there's `10px` (`.625rem`) vertical padding/margin around elements, which simulates a `20px` gap between them. If we reduce the gap to match the design, then we also reduce their touch target area.

We could only reduce the vertical padding of the text elements (_Get Involved_ & _Community_), but it looks a bit weird, in my opinion. I would leave the spacing between elements even.

CleanShot 2024-05-21 at 18 36 18@2x

mirhamasala:
> At the moment, there's 10px (.625rem) vertical padding/margin around elements, which simulates a 20px gap between them. If we reduce the gap to match the design, then we also reduce their touch target area.

The spacing after `Get Involved` - per Figma design - Should be `16px` - That's `py-4` (or `pb-4` - whichever you prefer) - Currently, it's only `2.5`

src/app/_components/DesktopNavigation.tsx
```diff
@@ -70,13 +70,18 @@ function SubNavItem({
   )
 }
 
+const mainNavItem = {
+  width: 'px-4',
```
mirhamasala, line 74:
```suggestion
  spacingX: 'px-4',
```

src/app/_components/DesktopNavigation.tsx
```diff
@@ -70,13 +70,18 @@ function SubNavItem({
...
+  offsetWidth: '-mr-4',
```
mirhamasala, line 75:
```suggestion
  offsetSpacingX: '-mr-4',
```

src/app/_components/DesktopNavigation.tsx
```diff
@@ -70,13 +70,18 @@ function SubNavItem({
...
+}
+
```
mirhamasala, line 77: I think we should try to standardize these. Using `touchTarget` everywhere.

```
const touchTarget = {
  class: 'px-4',
  offsetClass: '-mr-4',
}
```

Happy to suffix with the element name, if you think it's clearer: `touchTargetMainNavItem`

CharlyMartin:
> The spacing after Get Involved, etc. - Per Figma design - Should be 16px - That's py-4 (or pb-4 - whichever you prefer) - Currently, it's only 2.5.

We can't use `py-4` on each element. Otherwise, we end up with a cumulated gap of 32px in between: 16px bottom padding on the upper element and 16px top padding on the lower element.

To simulate a 16px gap with touch target areas on links, we need to use `py-2` or `my-2`. Does that make sense? With this solution (https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/145/commits/1ba02f41c6b8a65b25e71067fe9dc8c53677f7fc), all links are 40px high, which might not be sufficient in terms of a11y.

src/app/_components/MobileNavigation.tsx
```diff
@@ -19,28 +19,40 @@ import { PATHS } from '@/constants/paths'
...
-      )}
-    >
+    <li className={clsx(nested && 'ml-6')}>
       <Link
         href={path}
-        className={clsx(linkBaseStyles, linkTouchTarget.class)}
+        className={clsx(
+          linkBaseStyles,
+          "relative before:absolute before:inset-0 before:-m-2 before:content-[''] after:absolute after:inset-0 after:-m-2 after:content-['']",
+        )}
```
mirhamasala, line 55: @CharlyMartin ~ What do you think about this solution for touch targets?

mirhamasala (approved):
@CharlyMartin ~ 

- I did a refactor pulling out the `NestedMenu` for `MobileNavigation`.
- I also moved most of the layout + display styles to the most outer parent component.
- Most significantly, I found a [solution](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/145#discussion_r1616098789) for increasing the touch target of text links. Let me know what you think. If you like it, we can create a ticket to refactor the rest of the touch targets with this solution. We can even add it to the config.

src/app/_components/MobileNavigation.tsx
CharlyMartin: @mirhamasala I like it very much 👏 Let's find a way to add it to our config!

I increased the negative margins of the before element to be as close to 48px as possible. I also removed the after pseudo-class (only one should be necessary)

PR #146, opened 2024-05-21
Rework Card Layout

CharlyMartin:
This PR updates `CardLayout` by moving away from page-specific layout names such as `home` and using utility names such as `lgTwo`.

`lgTwo` means the grid starts with one column until the screen reaches the `lg` breakpoint, at which point it becomes two columns.

The `cols` prop now sets the grid behaviour, instead of `layout` before. I couldn't find a default layout among all the ones we are using, so the `cols` prop is now mandatory.

**A note on the events page:** The current event cards on the website are different from the ones on Figma. Hence, the layout is also different. When we update these cards, we can easily update the layout, too.

### DIFF
--- a/src/app/_components/CardGrid.tsx
+++ b/src/app/_components/CardGrid.tsx
@@ -9,7 +9,7 @@ type GridColumnConfig =
   | 'mdThree'
   | 'lgThree'
 
-type CardLayoutProps = {
+type CardGridProps = {
   cols: GridColumnConfig
   as?: React.ElementType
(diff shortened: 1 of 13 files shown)


mirhamasala (approved):
Awesome, @CharlyMartin ~ I pushed a couple of PRs with naming nits and changed the component name so you don't have to go back in. This rework was very much needed. Thank you. 🤩

CharlyMartin:
Great, thank you!

PR #148, opened 2024-05-22
Heading component: remove and refactor className

barbaraperic:
## Changes 
- removed `className` prop from `Heading` component


src/app/_components/PageHeader.tsx
```diff
@@ -40,7 +40,7 @@ export function PageHeader({
       {isFeatured && <SectionDivider title="Featured" />}
       <div className="flex flex-col gap-6 md:flex-row">
         <div className="flex flex-col gap-4 md:w-1/2">
-          <Heading tag="h1" variant="4xl" className="text-balance">
+          <Heading tag="h1" variant="4xl">
```
mirhamasala, line 43: Have you checked the difference between `text-balance` and `text-pretty`?

src/app/_components/Card.tsx
```diff
@@ -90,7 +90,7 @@ export function Card({
         )}
 
         {title && typeof title === 'string' ? (
-          <Heading tag="h3" variant="lg" className="line-clamp-2 text-ellipsis">
+          <Heading tag="h3" variant="lg">
```
mirhamasala, line 93: Should we add a prop `isClamped`? Since we've now effectively stopped clamping headings for blog post, I believe.

src/app/_components/PageHeader.tsx
barbaraperic: We should use `text-balance` on headings and `text-pretty` on paragraphs 👍. [Source](https://blog.stephaniestimac.com/posts/2023/10/css-text-wrap/#:~:text=Use%20text%2Dwrap%3A%20balance%3B,good%20candidate%20for%20progressive%20enhancement.).

mirhamasala: Interesting! Thank you for sharing.

mirhamasala (approved):
LGTM ✨🙌🏼✨

PR #149, opened 2024-05-22
FIX: add missing links

barbaraperic:
## Note
- still missing links for:

1.  Privacy Policy
2. Terms
3. Governance Page - "Learn More" in the PageHeader
4. Governane Page - link to slack #fil-gov in 
5. Governance Page - docs

[Link to google doc here](https://docs.google.com/document/d/1A7ffTMAv6Y-R9JdGGdQVgFUPglUA06FUdao6FHTvwP4/edit#heading=h.n9fh3fk7lcmy)


mirhamasala (changes requested):
Can you confirm if you also checked for:

```
href: '#',
```

It would be good to check for as many variations as you can think of. Thank you.

barbaraperic:
> Can we confirm if we also checked for:
> 
> ```
> href: '#',
> ```
> 
> It would be good to check for as many variations as we can think of. 🙏🏼

Yes double checked!

mirhamasala (approved):
@barbaraperic ~ There were still a couple of missing links. Added those. And some refactor nits. All good. LGTM, too! 🤩🙏🏼

PR #152, opened 2024-05-22
Add Graphics

CharlyMartin:
This PR implements the graphics & headshot pictures provided by the design team.

**Graphics** can currently be found in 3 places:
1. The header section of all pages except for the blog, ecosystem, and events where we have featured entries. I reworked the `PageHeader` component and simplified its prop API to make it work.
2. The `Focus Areas` section of the about page. I created the `FocusAreaCard` component for that purpose. It will be refactored when we rework `Card`.
3. The `Get Involved` section on this events page.

**Headshots** can be found in the `Who We Are` and `Advisors` sections of the about page. I created the `KeyMemberCard` component for that purpose.

### Static vs Dynamic Images
Next's `Image` component works with both local (static) and remote (dynamic) images, but their API differs slightly. With [static images](https://nextjs.org/docs/app/building-your-application/optimizing/images#local-images), the `src` prop is an object with data about the image that Next determines for us. With [remote images](https://nextjs.org/docs/app/building-your-application/optimizing/images#remote-images), the `src` prop is a string, and we, developers, need to populate the props manually.

So, I've created a thin layer on top of `Image` called `NextStaticImage` specifically to harness these nice optimizations for static images. It has a slightly different prop API to mark this difference.

`Image` -> remote image
`NextStaticImage` -> local imported images

### Open questions

1. `advisorsData` and `boardMembersData` are JS files. Any reason why? I created equivalent TS files to add types.
2. The order of the advisor cards on Figma is different from the website. Should it match Figma 1 to 1?
3. Should the focus area cards be clickable? (link to a page on the website)

### Screenshots

CleanShot 2024-05-24 at 17 15 45@2x

CleanShot 2024-05-24 at 17 16 52@2x

CleanShot 2024-05-27 at 15 45 17@2x

CleanShot 2024-05-27 at 15 40 14@2x

~ Related to UXIT-817


mirhamasala (changes requested):
1. Have these images been thrown into an image optimizer like [ImageOptim](https://imageoptim.com/mac) or https://squoosh.app/? (Happy for you to do some further research into what would be the best tool (Eventually, this is something that might become part of our CI pipeline.)
2. Please have a look at resizing some of the pages. Odd things happen with the layout.
3. I have dropped a few questions to Heath and Filipa regarding image ratios in Slack 
4. Something odd also happens with the Focus Area Card images - It straightens out when the page is refreshed - But since it doesn't happen anywhere else - I can't help to wonder what's going on.
5. Not on purpose. 
    > advisorsData and boardMembersData are JS files. Any reason why?
6. I don't think so.
   > Should the focus area cards be clickable? (link to a page on the website)
7. I'm not sure it matters. Probably best to check with Jordan. At some point, I imagine, we could actually create a CMS collection so that the content team is able to change this list when needed.
   > The order of the advisor cards on Figma is different from the website. Should it match Figma 1 to 1?

---

Screenshot 2024-05-30 at 16 29 05
Screenshot 2024-05-30 at 16 36 11
Screenshot 2024-05-30 at 16 36 28
Screenshot 2024-05-30 at 16 51 29

src/app/_components/PageHeader.tsx
```diff
@@ -41,59 +100,47 @@ export function PageHeader({
...
   description,
   cta,
-  secondaryCta,
   image,
   metaData,
   isFeatured = false,
-  containImageSize = false,
 }: PageHeaderProps) {
+  const mainCTA = Array.isArray(cta) ? cta[0] : cta
+  const secondaryCTA = Array.isArray(cta) ? cta[1] : undefined
```
mirhamasala, line 108: What is this for?

src/app/_data/graphicsData.ts
```diff
@@ -0,0 +1,70 @@
+import { StaticImageProps } from '@/types/sharedProps/imageType'
```
mirhamasala, line 1: I like this file. ❤️

src/app/_types/sharedProps/imageType.ts
```diff
@@ -1,4 +1,9 @@
+import { StaticImageData, ImageProps as NextImageProps } from 'next/image'
+
 export type ImageProps = {
   url: string
-  alt: string
-}
+} & Omit<NextImageProps, 'src' | 'objectFit'>
```
mirhamasala, line 5: Are we omitting this because we have `url` instead of `src`?

src/app/_components/KeyMemberCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+      />
+
+      <div className="relative m-3 grow">
+        <Heading isClamped tag="h3" variant="lg">
+          {name}
+        </Heading>
+
+        <p className="mt-1 text-brand-300">{title}</p>
+
+        <TextLink
```
mirhamasala, line 35: At the moment, the entire card except the image in clickable. Is this intentional?

src/app/_components/KeyMemberCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+    <li className="flex rounded-lg border border-brand-500 p-1">
+      <Image
+        src={image}
+        alt={`Photo of ${name}`}
+        sizes="10rem"
+        className="aspect-[3/4] w-32 rounded object-cover"
+      />
+
+      <div className="relative m-3 grow">
+        <Heading isClamped tag="h3" variant="lg">
```
mirhamasala, line 29: We shouldn't clamp the heading here - Since there's no link to, let's say, another page where the user can see it in full - Like for blog posts or ecosystem projects or events that have their own individual pages.

src/app/_components/FocusAreaCard.tsx
```diff
@@ -0,0 +1,34 @@
+import { Heading } from '@/components/Heading'
```
mirhamasala, line 1: I'm tempted to say that we could use `Card` but I'm not sure without trying?

src/app/_components/KeyMemberCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+        </Heading>
+
+        <p className="mt-1 text-brand-300">{title}</p>
+
+        <TextLink
+          href={linkedin}
+          aria-label={`Visit ${name}'s LinkedIn profile.`}
+          className="absolute inset-0 rounded-lg focus:outline focus:outline-2 focus:outline-brand-100"
+        >
+          <span className="absolute bottom-0 left-0 flex items-center gap-2 text-brand-300">
```
mirhamasala, line 40: Perhaps we can align this implementation more closely to the way `GovernanceCalendarCards` is one with `Card.Link`?

CharlyMartin:
Thanks @mirhamasala 

> Have these images been thrown into an image optimizer

Next.js already optimises statically imported images: https://youtu.be/IU_qq_c_lKA?si=r3V3Xm4wl4pprVZB&t=112 But, we can push it further by pre-optimising these ourselves too.

On the same note, we should optimise dynamic images for blog entries, projects, etc. These ones don't get optimised by Next.js.

src/app/_components/PageHeader.tsx
```diff
@@ -8,25 +8,84 @@ import {
...
+function DynamicImage({
+  url,
+  alt,
+  objectFit = 'cover',
+  fallback,
+  ...rest
+}: DynamicImageProps) {
+  if (!url) {
+    return <StaticImage {...fallback} />
+  }
```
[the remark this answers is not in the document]
CharlyMartin: From Next.js perspective, both static and dynamic images rely on the native `Image` component, but how to use this component differs depending on the kind of image.

With [static images](https://nextjs.org/docs/app/building-your-application/optimizing/images#local-images), the `src` prop accepts a config object containing data about the image's source, width, height, placeholder, etc.

With [dynamic images](https://nextjs.org/docs/app/building-your-application/optimizing/images#remote-images), the `src` prop accepts a URL string, while width, height, placeholder, etc need to be configured manually.

With 2 components, it's easier to ensure that each prop API fits the use case. For instance, `NextStaticImage` does not accept a `width` API because it's computed by Next for us. But `Image` does because it's controlled by us.

Does it make sense?

src/app/_types/sharedProps/imageType.ts
CharlyMartin: Exactly!

`url` is for `ImageProps` and `data` is for `StaticImageProps`

src/app/_components/KeyMemberCard.tsx
CharlyMartin: Ah no, good catch!

src/app/_components/FocusAreaCard.tsx
CharlyMartin: The layout changes from vertical to horizontal and back to vertical, depending on the breakpoint. I don't think it's possible to do with the current implementation of `Card`?

There will be a proper refactor of the card component eventually to break it down into smaller pieces and make these layouts possible :)

CleanShot 2024-05-31 at 14 28 57

src/app/_components/KeyMemberCard.tsx
CharlyMartin: ✅

src/app/_components/KeyMemberCard.tsx
CharlyMartin: I like it! I wasn't aware of `Card.Link`.

I ended up moving the link up to make the whole card clickable.

src/app/_components/KeyMemberCard.tsx
CharlyMartin: ✅

mirhamasala:
> Next.js already optimises statically imported images: https://youtu.be/IU_qq_c_lKA?si=r3V3Xm4wl4pprVZB&t=112.

@CharlyMartin - Thank you for the resource. Can you please quote or point me to whatever fragment in a video you'd like me too watch in the future? Also, Next.js documentation might be better. 🙏🏼

That said, I'm aware that Next.js optimises images. However:

1. With pre-optimised images, Next.js may still perform some optimizations based on device and parameters, but the processing is minimal compared to processing large images.
2. Every commit you make is stored permanently in the repository's history. Removing files in a later commit does not remove them from the history. When someone clones your repository, they get the entire history, including all the large files you checked in and later optimized or deleted. Large files, especially binaries like images, can quickly increase the size of your repository, making it slower to clone, fetch, and interact with. One of the reasons why [Git Large File Storage (LFS)](https://git-lfs.com/) exists.


Though, as it is, this commit has already been made, and we do have a ticket open to optimize images across the project, we can pick it up in a subsequent ticket ~ However, I also want us to be aware, that optimising these assets should've been part of the scope of this ticket - and to underline, that optimization, especially, when images are added by the engineering team should be optimized before being pushed (@barbaraperic - FYI). Thank you.

mirhamasala:
> Thanks! I created a [ticket](https://www.notion.so/filecoin/Resize-dynamic-images-50100bdc00474556b572c5993203c086?pvs=4) for it. Let's not explode the scope of this PR, it's already big.

@CharlyMartin - If you check the current ticket, making sure image ratios look good are part of it. Happy to have a separate ticket - but let's make it a priority to pick up since having the images across the site look decent is the last important bit before the launch. 🙏🏼

src/app/_types/sharedProps/imageType.ts
mirhamasala: Let's create a ticket to change `url` to `src` and remove this omission. 🙏🏼

src/app/_components/PageHeader.tsx
```diff
@@ -8,25 +8,84 @@ import {
...
 
-type PageHeaderProps = {
+type SharedProps = {
   title: TitleProps['children']
   description?: DescriptionTextType
-  cta?: CTAProps
-  secondaryCta?: CTAProps
-  image?: ImageProps
-  metaData?: Array<string | null | undefined>
+  cta: CTAProps | [CTAProps, CTAProps?]
```
mirhamasala, line 28: In the case of the array - How come the second `CTAProps` is optional?

mirhamasala: Also why do we need a separate `SharedProps` type? Instead of just a `PageHeaderProps`?

```
type PageHeaderProps = {
  title: TitleProps['children']
  cta: CTAProps | [CTAProps, CTAProps]
  description?: DescriptionTextType
  metaData?: MetaDataType
  isFeatured?: boolean
  image:
    | ({ type: 'static' } & StaticImageProps)
    | ({ type: 'dynamic' } & DynamicImageProps)
}
```

src/app/_components/PageHeader.tsx
```diff
@@ -8,25 +8,84 @@ import {
...
+  metaData?: MetaDataType
   isFeatured?: boolean
-  containImageSize?: boolean
+}
+
+type DynamicImageProps = Partial<ImageProps> & {
+  fallback: StaticImageProps
+  objectFit?: 'contain' | 'cover'
+}
```
mirhamasala, line 36: Should this live under `imageTypes`?

src/app/_components/PageHeader.tsx
```diff
@@ -8,25 +8,84 @@ import {
...
-import { Meta } from '@/components/Meta'
+import { Meta, type MetaDataType } from '@/components/Meta'
+import { NextStaticImage } from '@/components/NextStaticImage'
 import { SectionDivider } from '@/components/SectionDivider'
 
 import { type CTAProps } from '@/types/sharedProps/ctaType'
-import { type ImageProps } from '@/types/sharedProps/imageType'
+import {
+  type StaticImageProps,
+  type ImageProps,
```
mirhamasala, line 18: `ImageProps`, `StaticImageProps`, `DynamicImageProps`, `NextImageProps` 😅 ~ I'm getting slightly confused. Let's talk about this later!

src/app/events/page.tsx
```diff
@@ -240,15 +246,31 @@ export default function Events({ searchParams }: Props) {
...
         title="Get in Touch With the Events Team"
       >
-        <CardGrid cols="mdThree">
-          {getInvolvedData.map(({ title, description, cta }) => (
-            <Card
-              key={title}
-              title={title}
-              description={description}
-              cta={cta}
+        <CardGrid cols="mdTwo">
```
mirhamasala, line 249: Would you say this is a list with list items?

mirhamasala (changes requested):
Let's pair to get his PR sorted and merged. 🙏🏼

src/app/_components/PageHeader.tsx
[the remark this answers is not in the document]
mirhamasala: Very cool! ❤️

src/app/_components/PageHeader.tsx
CharlyMartin: Yes good catch, the second `CTAProps` in the array doesn't have to be optional.

CharlyMartin: And also yes, we can group all the types in `PageHeaderProps`

src/app/_components/KeyMemberCard.tsx
mirhamasala: Maybe check ~

```
          <li
            key={id}
            className="flex flex-col rounded-lg border border-blue-500 p-1 sm:flex-row"
          >
            <Calendar startDate={start.dateTime} />
            <div className="relative">
              <div className="relative flex flex-1 flex-col gap-3 p-4">
                <div className="flex gap-2">
                  <Badge
                    borderColor="brand-100"
                    icon={Clock}
                  >{`UTC ${startTime} - ${endTime}`}</Badge>
                  <Badge>Zoom</Badge>
                </div>
                <Heading tag="h3" variant="lg">
                  {summary}
                </Heading>
              </div>
              <div className="mb-10 mt-3">
                <Card.Link
                  href={htmlLink}
                  icon={CalendarPlus}
                  text="Add to Google Calendar"
                />
              </div>
            </div>
          </li>
```

src/app/_components/PageHeader.tsx
CharlyMartin: ✅

src/app/_components/PageHeader.tsx
```diff
@@ -8,25 +8,84 @@ import {
...
+    return <StaticImage {...fallback} />
+  }
+
+  return (
+    <div className={clsx('relative', imageContainerStyle)}>
+      <Image
+        fill
+        priority
+        src={url}
+        alt={alt || 'Missing alt description'}
```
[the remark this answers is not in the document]
CharlyMartin: ✅

src/app/events/page.tsx
CharlyMartin: ✅

src/app/_components/PageHeader.tsx
CharlyMartin: ✅

src/app/_components/PageHeader.tsx
CharlyMartin: `DynamicImageProps` now lives alongside `DynamicImage`, same for `StaticImageProps`.

I kept `ImageProps`, and reverted it to its original version, because it's still used by entry types (`BlogPostData` and the like), and all content entries use `url` for images. 

```ts
export type ImageProps = {
  url: string
  alt: string
}
```

src/app/_components/KeyMemberCard.tsx
CharlyMartin: I fixed it with this commit https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/152/commits/80e7b9ed1e015036518ec74d866e6f1018a8b44f

I didn't use `Card.Link` because it needs a refactor to work for this use case. We can do it later; it's out of scope for this PR.

As mentioned earlier today, `GovernanceCalendarCards` is not fully clickable. Once `Card.Link` is refactored, we can implement it properly there, too.

https://www.notion.so/filecoin/FF-V4-Final-checks-and-cleanups-e09f9b0526c646b7ba533215964ed198?pvs=4#2efb468ceef14f8dab23b84b2007cf65

src/app/_utils/buildImageSizeProp.ts
```diff
@@ -0,0 +1,37 @@
...
+  const widthAsString = screens[screenSize].replace('px', '')
+  const widthAsNumber = Number(widthAsString)
+
+  return widthAsNumber
+}
+
+export function buildImageSizeProp(args: Args) {
+  const { then, ...pairs } = args
+
+  const entries = Object.entries(pairs) as Array<[Breakpoint, Width]>
```
mirhamasala, line 20: Something like

breakpointWidthPairsObject = pairs
breakpointWidthPairsArray = entries

src/app/events/page.tsx
```diff
@@ -240,15 +248,31 @@ export default function Events({ searchParams }: Props) {
...
-              key={title}
-              title={title}
-              description={description}
-              cta={cta}
+        <CardGrid cols="mdTwo" as="div">
+          <div className="row-span-2 h-96 md:h-auto">
+            <StaticImage
+              {...graphicsData.events2}
+              className="h-full rounded-lg object-cover"
+              sizes={buildImageSizeProp({ md: '50vw', then: '100vw' })}
```
mirhamasala, line 256:
```suggestion
              sizes={buildImageSizeProp({ md: '100vw', then: '50vw' })}
```

src/app/events/page.tsx
```diff
@@ -240,15 +248,31 @@ export default function Events({ searchParams }: Props) {
...
+                cta={cta}
+                as="div"
+              />
+            </div>
           ))}
+          <div className="h-48 md:h-56">
+            <StaticImage
+              {...graphicsData.events3}
+              className="h-full rounded-lg object-cover"
+              sizes={buildImageSizeProp({ md: '50vw', then: '100vw' })}
```
mirhamasala, line 273:
```suggestion
              sizes={buildImageSizeProp({ md: '100vw', then: '50vw' })}
```

src/app/_components/DynamicImage.tsx
```diff
@@ -0,0 +1,37 @@
+import Image, { type ImageProps } from 'next/image'
+
+import clsx from 'clsx'
+
+import { StaticImage, type StaticImageProps } from '@/components/StaticImage'
+
+export type DynamicImageProps = {
+  src: string
+  fallback: StaticImageProps
+  objectFit?: 'contain' | 'cover'
```
mirhamasala, line 10: Move `objectFit?: 'contain' | 'cover'` to `imageType`

mirhamasala (approved):
Only the few comments left ..

PR #164, opened 2024-05-28
Update blog post data with cateogries and seo

mirhamasala:


mirhamasala:
@barbaraperic @CharlyMartin Tagging for visibility. Definitely no need to review. 🙃

PR #169, opened 2024-05-29
Remove Validation Checks for Populated Fields

mirhamasala:
These fields were previously excluded from validation checks because we still hadn't populated all blog posts, events, and ecosystem projects with these required fields. Now that we have, we can remove them and the the validation function do its job. The only one left to remove will be `location`.


mirhamasala:
@barbaraperic @CharlyMartin Tagging for visibility. No need to review.

PR #174, opened 2024-06-03
SEO metadata tests - add terms and privacy policy page to e2e testing

barbaraperic:
## Changes
- amended `testPageMetadata` so it skips testing for `header` since terms and privacy policy page don't have one.
- also included a fix for [single_blog_post_page_spec.cy.ts](https://github.com/FilecoinFoundationWeb/filecoin-foundation/compare/bp/test-seo-metadata?expand=1#diff-c6bca5ace1c0d43e706b0739133dc79943a54246eafbb93ad41d640d2d9825f8)


cypress/e2e/single_blog_post_page_spec.cy.ts
```diff
@@ -12,7 +12,9 @@ describe('Single Blog Post Page', function () {
         if (typeof href === 'string') {
           cy.visit(href)
 
-          const pagePath = href.replace(BASE_URL, '')
+          const pagePath = href.startsWith(BASE_URL)
+            ? href.replace(BASE_URL, '')
+            : href
```
mirhamasala, line 17: What does this fix?

cypress/support/test-utils.ts
```diff
@@ -95,3 +102,21 @@ function verifyHeaderContent(content: {
...
+  if (content.description) {
+    if (Array.isArray(content.description)) {
+      content.description.forEach((text: string) => {
+        cy.get('p').should('contain.text', text)
+      })
+    } else {
+      cy.get('p').should('contain.text', content.description)
+    }
+  }
+}
```
mirhamasala, line 122: Why does we have `description` here? If this is meant to check pages without a PageHeader - like Privacy and Terms - then, we only check for the title, right? Or did I miss something?

PR #175, opened 2024-06-04
Card images optimisation

CharlyMartin:
This PR optimises images in `Card`  and `Card`-related components:

1. `Card` components now use `DynamicImage` or `StaticImage`
2. `Card` components with dynamic images have a `sizes` prop
3. `buildImageSizeProp` now uses `min-width` to match Tailwind media queries and provide a similar developer experience when using breakpoints
4. All fallback images are now the same: We don't use "Logo Coming Soon" anymore, only the one with the Filecoin logo
5. `Card` entries with logo now have some padding, thanks to the `padding` prop
6. Images in `PageHeader` and `Card` are reworked to avoid passing custom non-valid props, such as `padding` or `type`, to the final `Image` (which created warnings in the console)

[Notion](https://www.notion.so/filecoin/FF-V4-IMAGES-Handle-images-in-Card-component-Other-Cards-3a2993d3b62c4955acfd570eee5117f1?pvs=4)

### DIFF
--- a/src/app/_components/BlogPostHeader.tsx
+++ b/src/app/_components/BlogPostHeader.tsx
@@ -44,7 +44,10 @@ export function BlogPostHeader({
             src={image.url}
             alt={image.alt}
             className="rounded-lg object-cover"
-            sizes={buildImageSizeProp({ md: '100vw', fallbackSize: '700px' })}
+            sizes={buildImageSizeProp({
+              startSize: '100vw',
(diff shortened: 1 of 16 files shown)


mirhamasala (approved):
@CharlyMartin 🙌🏼🫣🔥 Totally nailed it! 🙏🏼 LGTM so our site is finally looking polished! Yay!

PR #178, opened 2024-06-05
Fixup SEO Titles

mirhamasala:
TL;DR - I found a solution; It's to use `title.absolute` [>>](https://nextjs.org/docs/app/api-reference/functions/generate-metadata#absolute) "which can be used to provide a title that ignores `title.template` set in parent segments."

---

Following David's advice on how we should construct our SEO titles, we find ourselve in a jam not being able to use a parent `template` because of the different use cases.

1. For all pages where we have an SEO title available through the content file - for example, `/about`

```
---
header:
  title: "About Us"
  description: "Filecoin is the world’s largest decentralized storage network, designed to store humanity’s most important information. Filecoin Foundation facilitates governance of the Filecoin network, funds research and development projects for decentralized web technologies, and supports the growth of the Filecoin community and ecosystem."
seo:
  title: "About Filecoin Foundation – Mission & Values"
  description: "The Foundation facilitates the governance of the Filecoin network, funds research and development projects for decentralized web technologies, and supports the growth of the Filecoin community and ecosystem."
---
```

We need to use the `seo.title` as is. However, since most already contain a mention to `Filecoin Foundation`, we don't want to suffix the title with `| Filecoin Foundation`.

2. We do want to suffix "simple" page titles like `Privacy Policy` and `Terms of Use` - However we can't use `template` because if we did, the suffix would get added everywhere. So now we need to explicitly add the suffix for these pages in the content file - Which is a pity - Since that's kinda what `template` is for.

3. We do also want to suffix dynamic paths like blog posts, events entries, and ecosystem projects. Again, however, we can't use the `template` for the same reason as for simple pages. Now, however, we also don't want to be adding the suffix to all blog posts, etc. hence we're now adding it in `createMetadata`.

Personally, I'm not sure if the SEO advice outweighs these complications. For now, we can move ahead. But we should discuss this more after the launch - and perhaps revert to using `template` if we think it's better.


mirhamasala:
Blocked by https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/174 - @barbaraperic

mirhamasala:
@CharlyMartin - Please go ahead and review - Ignoring the Cypress Tests which are currently blocked by https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/174 - Thank you.

mirhamasala:
This PR was live reviewed with @CharlyMartin - Thank you! 🙏🏼

PR #187, opened 2024-06-05
Breadcrumbs: change icon size

barbaraperic:
## Visuals
Screenshot 2024-06-05 at 20 39 38
Screenshot 2024-06-05 at 20 39 24


src/app/_components/BreadCrumbs.tsx
```diff
@@ -51,7 +51,7 @@ export function BreadCrumbs() {
                 <Icon
                   component={CaretRight}
                   color="brand-400"
-                  size={20}
+                  size={24}
```
mirhamasala, line 54: @barbaraperic 
```suggestion
```

mirhamasala (approved):
If `24` is what we need, we can remove it! It's the default.

```
export function Icon({
  component,
  color = 'inherit',
  size = 24,
  weight = 'regular',
}: IconProps) {
  const Component = component

  return (
    <span aria-hidden="true" className={colorStyles[color]}>
      <Component size={size} weight={weight} />
    </span>
  )
}
```

🚀 How many people are working behind your screen. This is like a marathon. ❤️ LGTM!

PR #189, opened 2024-06-06
Fix and clean up CMS Config

mirhamasala:
- Fix featured Ecosystem Projects collection name
- Fix CMS Config for featured Grant Graduates
- Include Image alt text hint
- Review required fields
- Hide old tags options from CMS - Which we do want to keep in the data for now
- Hide recommended post from CMS - We might use this in the future
- Clean up


barbaraperic:
What is the best way to test this?

mirhamasala:
> What is the best way to test this?

@barbaraperic You can log into the Admin CMS and check it all. Considering launch - Skimming the Git diff is enough. I checked the CMS twice.

PR #217, opened 2024-06-07
Remove Legacy Images

CharlyMartin:
This PR cleans up and harmonises the use of images, moving away from using standalone `Image` in favour of our image wrappers: `StaticImage` for imported images and `DynamicImage` for CMS content.

To do so, this PR:
1. Updates `DynamicImage` so that the configuration passed the dynamic image (`sizes`, `fill`, `priority`, `quality`) also applies to its fallback,
2. Adds `sizes` prop to all images for loading optimisation,
3. Removes `CustomImage` from `MarkdownContent` in favour of  `DynamicImage`. This `DynamicImage` instance uses absolute width and height values instead of `fill` with a relatively positioned parent. This is because the second option creates a `p` > `div` > `img` HTML tree, which throws a hydration error as a `div` cannot be the direct descendant of a paragraph.
4. Adds a forgotten graphic in the `Understanding Governance` page section

The linked [Notion](https://www.notion.so/filecoin/FF-V4-IMAGES-Remove-Legacy-Image-Props-601c17b3e1a344c89cbb412a8b92aa82?pvs=4) [Tickets](https://www.notion.so/filecoin/FF-V4-IMAGES-Check-CustomImage-used-in-MarkdownContent-b424bc6b8ca94f6a873540722200f743?pvs=4)


barbaraperic:
@CharlyMartin there is both `url` and `src` attribute in the `img` tag in `DynamicImage` component

Screenshot 2024-06-12 at 10 48 34

src/app/_components/Card.tsx
```diff
@@ -55,16 +55,11 @@ function CardImage({ image }: Pick<CardProps, 'image'>) {
   const isStaticImage = 'data' in image
```
barbaraperic, line 55: is this needed? do we use static images in Card component?

CharlyMartin:
> @CharlyMartin there is both `url` and `src` attribute in the `img` tag in `DynamicImage` component

Good catch, @barbaraperic; I didn't notice. It's obviously not ideal to have both `url` and `src` on the HTML image tag, but it's not an issue as per ChatGPT:
> the browser will ignore the `url` attribute and display the image specified by the `src` attribute. The `url` attribute will not cause any errors, but it won't serve any purpose either

It happens because Markdown content entries still use `url` for images, and it gets passed down to the final component when we spread the image object.

It will be fixed once we clean all the entries in the coming days with this [ticket](https://www.notion.so/filecoin/FF-V4-IMAGES-Change-Image-URL-Fields-to-SRC-across-project-04bdb46262d64c07b574cb05c3b792e5?pvs=4).

PS: I implemented a CR, but it doesn't fix the problem, it's a refactor so the `image` prop on `PageHeader`  is consistent with `Card`.

src/app/_components/Card.tsx
CharlyMartin: We do in the `Reports` on the about page

src/app/_components/MarkdownContent.tsx
```diff
@@ -14,7 +18,14 @@ export function MarkdownContent({ children }: MarkdownContentProps) {
       className="prose"
       components={{
         img: ({ src, alt }) => (
-          <CustomImage src={src as string} alt={alt as string} />
+          <DynamicImage
+            src={src || ''}
+            alt={alt || ''}
+            width={800}
+            height={450}
```
barbaraperic, line 25: These don't affect the image size right? Are we passing the values just because its required?

barbaraperic:
Sorry don't quite understand this "_Updates DynamicImage so that the configuration passed the dynamic image (sizes, fill, priority, quality) also applies to its fallback_" - could you point me to the code 🙏

src/app/_components/PageHeader.tsx
```diff
@@ -39,43 +34,40 @@ function Title({ children }: TitleProps) {
...
+const sharedImageStyle = 'h-full w-full rounded-lg border border-brand-100'
+const imageSizes = buildImageSizeProp({ startSize: '100vw', md: '660px' })
 
-  if (type === 'static') {
-    const staticImage = rest as StaticImageProps
+function PageHeaderImage({ image }: Pick<PageHeaderProps, 'image'>) {
+  const isDynamicImage = 'src' in image
+  const isStaticImage = 'data' in image
 
+  if (isStaticImage) {
```
barbaraperic, line 45: It's a bit confusing that we have a condition here and also in the `DynamicImage` Component that checks `src` in the `image` object, but couldn't figure out why it wouldn't work if we just put return `DynamicImage`... but I guess something for future refactor

src/app/_components/PageHeader.tsx
```diff
@@ -39,43 +34,40 @@ function Title({ children }: TitleProps) {
...
           priority
           quality={100}
           className={sharedImageStyle}
-          sizes={buildImageSizeProp({ startSize: '100vw', lg: '480px' })}
-          fallback={{
-            ...dynamicImage.fallback,
-            className: sharedImageStyle,
-            fill: true,
-          }}
+          sizes={imageSizes}
```
barbaraperic, line 69: I like this approach with `sizes`, wondering if we could do that we others so instead of` buildImageSizeProp({ startSize: '100vw', md: '480px' })`, we have `imageMediumSizes` or something like that, and we have functions centralised in case we need to change.

barbaraperic:
We have an issue with `StaticImages` in ecosystem-explorer/[slug] - see `ecosystem-explorer/cooperdb` example

Screenshot 2024-06-13 at 12 23 31

barbaraperic:
> We have an issue with `StaticImages` in ecosystem-explorer/[slug] - see `ecosystem-explorer/cooperdb` example
> 
> Screenshot 2024-06-13 at 12 23 31

Also on Safari, but think this should be fixed with [this PR ](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/235)
Screenshot 2024-06-13 at 12 30 23

src/app/_components/MarkdownContent.tsx
CharlyMartin: Exactly! I explained why in the PR description above:

This `DynamicImage` instance uses absolute `width` and `height` values instead of `fill` with a relatively positioned parent. This is because the second option creates a `p` > `div` > `img` HTML tree, which throws a hydration error as a `div` cannot be the direct descendant of a paragraph.

We pass absolute values because we have to, but the image keeps its original aspect ratio.

src/app/_components/PageHeader.tsx
CharlyMartin: If it's confusing then yes we might need a refactor.

PR #235, opened 2024-06-12
Fix card image on Safari

CharlyMartin:
This PR fixes card sizing on mobile safari (screenshots below).

To do so, it removes the `flex-col` property on `Card` as it didn't seem to add anything.

I checked that it didn't affect other `Card` components, but it's best to double-check.

---

_Before_
image

_After_
image


src/app/_components/Card.tsx
```diff
@@ -44,7 +44,9 @@ function Link({ href, ariaLabel, icon: Icon, text }: CTAProps) {
 }
 
 const containerStyle = 'relative aspect-video'
-const imageStyle = 'rounded-lg px-1 pt-1'
+
+const imageBaseStyle = 'rounded-lg px-1 pt-1'
+const imagePaddingStyle = 'mx-4 my-2'
```
barbaraperic, line 49: why are we calling it padding style? Should it be margin? Also wondering if it would be easier if we add padding-x to the parent element (and remove margin-x from the div wrapping the img and padding-x in the div wrapping the content below). So we have it only defined once. But otherwise looks good 
IMG_3322

CharlyMartin: > why are we calling it padding style? Should it be margin?

I used padding as a general term for "space," but I agree it's confusing. I reorganised the classes, hopefully it makes more sense now.

About moving the padding to the parent `Card` container: I like this idea, and it would make sense if all images in `Card` had the same padding as the rest of the content, but for blog or event cards, it's not the case. Images there take up all the space.

barbaraperic (approved):

PR #237, opened 2024-06-12
Event Page: add location to event cards

barbaraperic:
## Note
- when the text wraps in a new line, the spacing between the `date` and `location` is `8px` (see left card below). It's because we use `flex` with `gap:2`. However, the design in Figma indicates `4px` gap between date and location (when wrapped). Didn't find an elegant solution - only hacky ones with negative margins, but honestly don't think it looks bad visually, I will double check with Filipa if it's ok with her if we leave `8px` gap everywhere. -> [Slack convo here](https://filecoinfoundation.slack.com/archives/C04HCHS4RRS/p1718267296396529)

## Visuals
Screenshot 2024-06-12 at 15 53 22

### DIFF
--- a/src/app/_components/Meta.tsx
+++ b/src/app/_components/Meta.tsx
@@ -14,7 +14,7 @@ export function Meta({ metaData }: MetaProps) {
   }
 
   return (
-    <div className="flex gap-3 text-brand-300">
+    <div className="flex flex-wrap gap-2 text-brand-300">
       {metaData.filter(Boolean).map((data, index) => {
         const isNotLastItem = index < metaData.length - 1
         const isNotFirstItem = index > 0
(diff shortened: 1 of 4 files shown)


src/app/_components/Meta.tsx
```diff
@@ -14,7 +14,7 @@ export function Meta({ metaData }: MetaProps) {
   }
 
   return (
-    <div className="flex gap-3 text-brand-300">
+    <div className="flex flex-wrap gap-2 text-brand-300">
```
CharlyMartin, line 17:
```suggestion
    <div className="flex flex-wrap gap-x-2 gap-y-1 text-brand-300">
```

CharlyMartin (approved):
Looking good :)

src/app/_components/Meta.tsx
CharlyMartin: Would that solve the problem you mentioned @barbaraperic?

barbaraperic: Nice 😍

PR #239, opened 2024-06-13
Sentry - fix tracesSampleRate

barbaraperic:
## Changes
 - fix to adjust the % of the transactions are captured in Sentry


CharlyMartin (approved):
Only one - non-blocking - comment:

if `tracesSampleRate` has the same value in 3 configuration files, why not turn it into an `ENV` variable? Something like `SENTRY_TRACES_SAMPLE_RATE=0.05`?

I'm not sure if these values need to remain the same across configuration files or if they can differ over time.

barbaraperic:
> tracesSampleRate

It might change based on the traffic, so I'll leave it like this for now.

PR #240, opened 2024-06-13
Set up Percy

barbaraperic:
## Note
- you need to add `PERCY_TOKEN` in your `.env`, you can find it in 1password file
- waiting to add Git integration, but I don't think it will affect the code -> [Slack convo](https://filecoinfoundation.slack.com/archives/C0275N63GS1/p1718285260464369)
- once approved, I'll add `PERCY_TOKEN` to Netlify 
- I added a `test` script, run `npm run test` and see if you can see screenshots in Percy dashboard (if you don't have access, ping Baldy 😊 )


CharlyMartin (approved):
I don't have access to the Percy dashboard yet so I can't access the test results (https://percy.io/3e0eb7b2/filecoin-foundation-site/builds/34854358). I pinged Baldy.

But I tested locally and it's running fine :)

package.json
```diff
@@ -9,7 +9,8 @@
     "build": "next build",
     "start": "next start",
     "lint": "next lint --fix",
-    "validate-md": "node scripts/validateMarkdown.js"
+    "validate-md": "node scripts/validateMarkdown.js",
+    "test": "npx percy exec -- cypress run"
```
CharlyMartin, line 13: Just one question/comment: Do we need to use `npx percy` since we installed `@percy/cli` below?

barbaraperic: Fixed to `percy exec -- cypress run` 👍

CharlyMartin: 👌

PR #241, opened 2024-06-13
FIX: duplicate id

barbaraperic:
## Changes

- We had an error for duplicate id on the `SearchInput` form - [Ticket](https://www.notion.so/filecoin/FF-V4-Fix-Console-Errors-ffa24e8ecada4a258a8caa5dd77e1861?pvs=4)
- included `id` prop in `SearchInput` to differentiate between mobile and web

## How to test
- test the search bar and check if you get an error (screenshot attached in the  [Ticket](https://www.notion.so/filecoin/FF-V4-Fix-Console-Errors-ffa24e8ecada4a258a8caa5dd77e1861?pvs=4))


src/app/_components/SearchInput.tsx
```diff
@@ -4,23 +4,27 @@ import { MagnifyingGlass } from '@phosphor-icons/react/dist/ssr'
 
 import { Icon } from '@/components/Icon'
 
+import { SearchProps } from './Search'
+
 type SearchInputProps = {
+  id: SearchProps['id']
   searchQuery: string
   onSearchChange: (event: ChangeEvent<HTMLInputElement>) => void
 }
```
CharlyMartin, line 13: How about doing it this way?

```suggestion
type SearchInputProps = {
  searchQuery: string
  onSearchChange: (event: ChangeEvent<HTMLInputElement>) => void
} & React.ComponentProps<'input'>
```


Since `id` is one of the native attributes of the `input` element, we can "import" it with a union type instead of typing it manually.

What do you think?

CharlyMartin (commented):
The web and mobile search ids seem inverted. The `web-search` input is visible on mobile and `mobile-search` on desktop.

Maybe this issue is in `FilterContainer`, I haven't looked into it.

As long as the ids are different, this PR's job is done. But, let's open a ticket if `FilterContainer` does not work as intended.

src/app/_components/SearchInput.tsx
barbaraperic: Oh that's cool, didn't know about that.

One problem might be that typescript then doesn't show errors if we don't include `id` as a prop in `SearchInput`.

### with React.ComponentProps<'input'>
Screenshot 2024-06-17 at 12 02 20


### with  id: SearchProps['id']

Screenshot 2024-06-17 at 12 00 05

barbaraperic:
> The web and mobile search ids seem inverted. The `web-search` input is visible on mobile and `mobile-search` on desktop.
> 
> Maybe this issue is in `FilterContainer`, I haven't looked into it.
> 
> As long as the ids are different, this PR's job is done. But, let's open a ticket if `FilterContainer` does not work as intended.

Good catch! Fixed ✅

src/app/_components/SearchInput.tsx
CharlyMartin: I see. How about something like this?

```ts
type SearchInputProps = {
  id: React.ComponentProps<'input'>['id']
  searchQuery: string
  onSearchChange: (event: ChangeEvent<HTMLInputElement>) => void
}
```

In the future, we might want to use `SearchInput` separately from `Search` so that the type declaration seems more accurate.

CharlyMartin (approved):

src/app/_components/SearchInput.tsx
barbaraperic: gotcha, that works :)

PR #249, opened 2024-06-19
Lock PageHeader Images in 16:9

CharlyMartin:
This PR updates `PageHeader` so that all images are in a `16:9` ratio instead of mirroring the height of the parent container.

_Now_
CleanShot 2024-06-20 at 17 46 07

_Before_
CleanShot 2024-06-20 at 17 46 21

---

[Notion Ticket](https://www.notion.so/filecoin/FF-V4-Adapt-PageHeader-images-a3c99aed7db44f06aab477f108e6ee61?pvs=4)

### DIFF
--- a/src/app/_components/PageHeader.tsx
+++ b/src/app/_components/PageHeader.tsx
@@ -1,3 +1,5 @@
+import clsx from 'clsx'
+
 import { Button } from '@/components/Button'
 import {
   DescriptionText,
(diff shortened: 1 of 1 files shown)


src/app/_components/PageHeader.tsx
```diff
@@ -34,38 +36,35 @@ function Title({ children }: TitleProps) {
   )
 }
 
-const sharedContainerStyle = 'relative aspect-video lg:aspect-auto lg:w-1/2'
-const sharedImageStyle = 'h-full w-full rounded-lg border border-brand-100'
-const imageSizes = buildImageSizeProp({ startSize: '100vw', md: '660px' })
+const aspectRatioStyle = 'aspect-video'
+const borderStyle = 'rounded-lg border border-brand-100'
+const imageSizes = buildImageSizeProp({ startSize: '100vw', lg: '490px' })
```
barbaraperic, line 41: Should this maybe be `lg: '480px'`?

Screenshot 2024-06-21 at 13 43 12

barbaraperic (approved):
👌

src/app/_components/PageHeader.tsx
CharlyMartin: I usually add a few extra pixels to the `sizes` prop so that the image chosen by the browser is slightly bigger than the size of the container to avoid pixelization / too low quality images.

PR #245, opened 2024-06-14
Grants Page: add links to cards

barbaraperic:
## Note
- [Slack convo here ](https://filecoinfoundation.slack.com/archives/C06MTEQ3SP6/p1718038788656819)
- [Ticket](https://www.notion.so/filecoin/FF-V4-Grants-Page-Link-Grants-and-Funding-Opportunities-Cards-00b37a6254154fa98cdcbae9e6f6e85b?pvs=4)

PR #250, opened 2024-06-20
Make Card.Link reusable

CharlyMartin:


barbaraperic:
While we're already refactoring `Link` component - do you think we could also include 'if it's an external link we place the arrow icon after the text'. It's an edge case we have.

Screenshot 2024-06-20 at 14 44 17

src/app/_components/Card.tsx
```diff
@@ -21,21 +22,42 @@ type CardProps = {
...
+
+type LeftProperty = `left-${SpacingValue}`
+type BottomProperty = `bottom-${SpacingValue}`
+
+type ResponsiveLeftProperty = `${BreakpointValue}:${LeftProperty}`
+type ResponsiveBottomProperty = `${BreakpointValue}:${BottomProperty}`
+
+type LinkProps = {
+  left?: LeftProperty | [LeftProperty, ResponsiveLeftProperty]
+  bottom?: BottomProperty | [BottomProperty, ResponsiveBottomProperty]
```
barbaraperic, line 36: maybe we could limit a number of options we can pass to the prop? For example, atm we only use 3 values for the left, and since (I assume) we won't have much variations we can limit to those we have until we need some other (to reduce the cognitive load of calculating).

Love the solution btw ❤️

CharlyMartin:
> While we're already refactoring `Link` component - do you think we could also include 'if it's an external link we place the arrow icon after the text'. It's an edge case we have.

Why not! I don't see where we use this pattern in the code, is it part of another ticket?

src/app/_components/Card.tsx
CharlyMartin: Do you mean replacing the current type declaration with hard-coded values (`left-36` and `left-4 sm:left-40`)? It's true that this config is a bit overkill given we only use it for a handful of values. So maybe your idea is simpler.

The `bottom` prop might be unnecessary for now since the CTA will always be 16px (4) from the bottom regardless of the layout. I added it for consistency, but maybe we remove it.

CharlyMartin:
Reference for conflict resolution:

- https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/245/files#diff-79c7fb873ede175db17e683cf9d5996b5fa63bebde1b6ea5a0773da758be57ceR21

CharlyMartin:
@barbaraperic 

> Yeah, maybe how we do it in, for example, PageHeader . So something along the lines:

I don't like names like `responsiveImageCard` because they don't tell what they do, as opposed to `left-36`, which is self-explanatory. So I stuck with my original solution and removed the `bottom` prop as we never use it. Is that ok with you?

---

> Here's an example, in the HomeExploreSectionCard component, the Learn More cta should have an ArrowUpRight icon on the right

I didn't notice/remember you already implemented a solution for this until I resolved conflicts in this branch.

I kept my implementation as we don't need an extra `isExternal` prop on `Card`. We can derive it by using the `isInternalLink` utility function.

Here's how `Card.Link` looks like now:

```tsx
// Helper function
function renderTextIcon({
  href,
  text,
  icon: Icon,
}: Pick<CTAProps, 'href' | 'text' | 'icon'>) {
  const isExternal = isExternalLink(href)
  const textElement = <span>{text}</span>

  if (Icon) {
    return [<Icon key="custom-icon" size={24} />, textElement]
  }

  if (isExternal) {
    return [textElement, <ArrowUpRight key="arrow-icon" size={24} />]
  }

  return textElement
}

// Card.Link
function Link({ href, ariaLabel, icon, text, left = 'left-4' }: LinkProps) {
  const textIcon = renderTextIcon({ href, text, icon })

  return (
    <CustomLink
      href={href}
      aria-label={ariaLabel}
      className="absolute inset-0 rounded-lg focus:brand-outline"
    >
      <span
        className={clsx(
          'absolute bottom-4 inline-flex items-center gap-2 text-brand-300',
          left,
        )}
      >
        {textIcon}
      </span>
    </CustomLink>
  )
}
```

Let me know what you think :)

src/app/_components/Card.tsx
```diff
@@ -21,38 +25,51 @@ type CardProps = {
...
+}: Pick<CTAProps, 'href' | 'text' | 'icon'>) {
+  const isExternal = isExternalLink(href)
+  const textElement = <span>{text}</span>
+
+  if (Icon) {
+    return [<Icon key="custom-icon" size={24} />, textElement]
+  }
+
+  if (isExternal) {
+    return [textElement, <ArrowUpRight key="arrow-icon" size={24} />]
```
barbaraperic, line 51: maybe we reverse the order on these 2 conditions? If, for example, we pass an book icon to external link, it will render the book icon and not the arrow icon. Not sure if we want the option to overwrite the arrow icon (for consistency sake)?

barbaraperic: Scratch that, we need to overwrite it. 😅 we have the google calendar case in governance

barbaraperic:
Looks good! prefer your solution over mine, much more elegant 💅

barbaraperic (approved):

PR #252, opened 2024-06-20
Include Past Events Category for Events

CharlyMartin:
This PR adds a `Past Events` category to the events page.

Unlike other categories, `Past Events` does not come from the CMS. It is programmatically calculated based on the end date of a given event or the start date If no end date is available.

Also, unlike other categories, this category is not exclusive to other categories, meaning an event can be both `Hosted` and `Past Events`, but it cannot be both `Hosted` and `Supported.`

To implement this mechanism, I created:
- `useEventsCategory`, built on top of `useCetagory`
- `getEventsCategorySettings`, built on top of `getCategorySettings`

I added these 2 functions to avoid "polluting" the original ones with code that is specific to the events page.

---

_No filters_
CleanShot 2024-06-21 at 14 44 25

_Past Events category filter_
CleanShot 2024-06-21 at 14 45 14

---

[Notion](https://www.notion.so/filecoin/FF-V4-Include-Past-Events-Category-for-Events-ed3745b367fe46faae1a5b1867dea68c?pvs=4)

### DIFF
--- a/src/app/_hooks/useEventsCategory.ts
+++ b/src/app/_hooks/useEventsCategory.ts
@@ -1,0 +1,63 @@
+import { useMemo } from 'react'
+
+import { UseCategoryProps, useCategory } from '@/hooks/useCategory'
+
+import type { EventData } from '@/types/eventTypes'
+
+import { isDateValid } from '@/utils/formatDate'
+
+import { pastEventsSetting } from '@/constants/categoryConstants'
+
+export function useEventsCategory(props: UseCategoryProps<EventData>) {
+  const { entries } = props
+
+  const results = useCategory(props)
+  const { categoryQuery, categoryCounts, categorizedResults } = results
+
+  const countsWithPastEvents = useMemo(() => {
+    const pastEventsCount = entries.filter(filterByPastEvents).length
+
+    return { ...categoryCounts, [pastEventsSetting.id]: pastEventsCount }
+  }, [categoryCounts, entries])
+
+  const resultsWithPastEvents = useMemo(() => {
+    if (categoryQuery === pastEventsSetting.id) {
+      return entries.filter(filterByPastEvents)
+    }
+
+    return categorizedResults
+  }, [categoryQuery, categorizedResults, entries])
+
+  return {
+    categoryQuery,
+    categorizedResults: resultsWithPastEvents,
+    categoryCounts: countsWithPastEvents,
+  }
+}
+
+const today = getTodayDateUTCZero()
+
+function filterByPastEvents(entry: EventData) {
+  const { startDate, endDate, slug } = entry
+
+  if (endDate) {
+    if (!isDateValid(endDate)) {
+      throw new Error(`Invalid date provided for event: ${slug}`)
+    }
+    return new Date(endDate) < today
+  }
+
+  if (!isDateValid(startDate)) {
+    throw new Error(`Invalid date provided for event: ${slug}`)
+  }
+  return new Date(startDate) < today
+}
+
+function getTodayDateUTCZero() {
+  const today = new Date()
+  const utcYear = today.getUTCFullYear()
+  const utcMonth = today.getUTCMonth()
+  const utcDate = today.getUTCDate()
+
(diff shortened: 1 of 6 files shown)


src/app/_hooks/useEventsCategory.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (!isDateValid(endDate)) {
+      throw new Error(`Invalid date provided for event: ${slug}`)
+    }
+    return new Date(endDate) < today
+  }
+
+  if (!isDateValid(startDate)) {
+    throw new Error(`Invalid date provided for event: ${slug}`)
+  }
+  return new Date(startDate) < today
```
barbaraperic, line 53: Wondering if `endData < today` would be a better indication of a past event? If `startData < today` it could still be an ongoing event

src/app/_hooks/useEventsCategory.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+const today = getTodayDateUTCZero()
+
+function filterByPastEvents(entry: EventData) {
+  const { startDate, endDate, slug } = entry
+
+  if (endDate) {
+    if (!isDateValid(endDate)) {
+      throw new Error(`Invalid date provided for event: ${slug}`)
```
barbaraperic, line 45: Invalid `start` / `end` date provided for event...?

barbaraperic:
We should include some kind of filtering for categories in for the past events hook as well? For example, I tested with `cohosted` category, and it shown in the Past Events 

Screenshot 2024-06-24 at 11 20 31

src/app/_hooks/useEventsCategory.ts
CharlyMartin: Totally, but `endDate` is optional on `EventData`, so not guaranteed to be present here.

For reference,  I wrote in the PR description that:

> It is programmatically calculated based on the end date [...] or the start date If no end date is available.

The function looks like this:

```ts
function filterByPastEvents(entry: EventData) {
  const { startDate, endDate, slug } = entry

  if (endDate) {
    if (!isDateValid(endDate)) {
      throw new Error(`Invalid date provided for event: ${slug}`)
    }
    return new Date(endDate) < today
  }

  if (!isDateValid(startDate)) {
    throw new Error(`Invalid date provided for event: ${slug}`)
  }
  return new Date(startDate) < today
}
```

src/app/_hooks/useEventsCategory.ts
CharlyMartin: Yes :)

CharlyMartin:
> We should include some kind of filtering for categories in for the past events hook as well? For example, I tested with `cohosted` category, and it shown in the Past Events

@barbaraperic I don't understand, let's huddle :)

CharlyMartin:
> We should include some kind of filtering for categories in for the past events hook as well? For example, I tested with `cohosted` category, and it is shown in the Past Events

@barbaraperic The `filterByPastEvents` filter doesn't check whether a given category is supported; it only checks whether an event is in the past. So it's doing its job correctly, but you're right, it's not ideal.

We should fix this [invalid category](https://www.notion.so/filecoin/FF-V4-Remove-invalid-event-category-cohosted-6b408c05c21042e59a9d1724b1dda5d7?pvs=4)  in a different PR. Then find a way to programmatically check category validity before merging PRs.

src/app/_hooks/useEventsCategory.ts
mirhamasala: @barbaraperic - I've changed it for better readability to. Hopefully that clears up the confusion.

```
function filterByPastEvents(entry: EventData): boolean {
  const today = getUTCMidnightToday()
  const { startDate, endDate, slug } = entry

  if (endDate && !isDateValid(endDate)) {
    throw new Error(`Invalid endDate provided for event: ${slug}`)
  }

  if (!isDateValid(startDate)) {
    throw new Error(`Invalid startDate provided for event: ${slug}`)
  }

  const eventDate = endDate ? new Date(endDate) : new Date(startDate)
  return eventDate < today
}
```

PR #258, opened 2024-06-26
[UXIT-1161] Goodbye jQuery [skip percy]

CharlyMartin:
This PR removes the minified `JQuery` file from the assets.

~But, even now, `JQuery` is still active on the page. It's part of the JS bundle sent to the browser. I need to investigate it further.~


mirhamasala:
Closing due to inactivity.

mirhamasala:
@CharlyMartin - Is the description of this PR correct? You need to investigate it further?

PR #264, opened 2024-06-28
Orbit Page - Global Orbit Events Section

CharlyMartin:
This PR introduces an Events section to the Orbit page.

Events are fetched server-side from the Airtable Orbit base using the Airtable.js client.

⚠️ Ensure `AIRTABLE_READ_ONLY_TOKEN` is added to `.env.local` (1Password) and run `npm i` to test locally.

### Event Filtering and Sorting

Events are filtered based on `Approved` status and start date so that only future events are included:
```ts
filterByFormula: `AND(
  {${AIRTABLE_ORBIT_BASE.EVENTS_TABLE.FIELDS.STATUS}} = '${AIRTABLE_APPROVED_STATUS}',
  IS_AFTER({${AIRTABLE_ORBIT_BASE.EVENTS_TABLE.FIELDS.START_DATE}}, '${new Date().toISOString()}')
)`,
```

Events are sorted by start date in ascending order, the soonest to the furthest:
```ts
sort: [
  {
    field: AIRTABLE_ORBIT_BASE.EVENTS_TABLE.FIELDS.START_DATE,
    direction: 'asc',
  },
],
```

### Data Format

I used field IDs instead of names - using `returnFieldsByFieldId: true` - to avoid breaking changes. If the Orbit team renames the `Event Title` column to `Event Name`, our integration would break if we relied on names. Ids are safer as they cannot be changed.

Hence, we receive the data in the following format and format it in the frontend.

```ts
{
  fldhD7ClKauhqeD79: 'FILECOIN ORBIT GHANA HACKACTHON WORKSHOP',
  fldQJZQAS4Jm0eiiL: 'Accra',
  flddTSgPuKgL6h1JG: '2024-07-19T18:00:00.000Z',
  fldnRqMzqmu7VLUlD: 'Approved',
  fldlcV4631rFgbdd6: 'https://dn4pckeyqvf.typeform.com/to/B9PHIoDB'
},
```

There's one potential Issue: The status value `Approved` is hard-coded.

If the Orbit team changes this value, to `Accepted` for instance, the `filterByFormula` will no longer work as expected. Although this scenario is unlikely, we might consider adding a check here. What do you think?

---

CleanShot 2024-07-16 at 09 52 50

[Notion](https://www.notion.so/filecoin/FF-Orbit-Page-Add-Events-section-5d9593a5d2f442efb53648c22231fd8b?pvs=4)


src/app/_components/OrbitEventsSection.tsx
```diff
@@ -0,0 +1,167 @@
...
+  searchParams,
+}: OrbitEventsSectionProps) {
+  try {
+    const events = await fetchAndParseAirtableEvents()
+
+    if (events.length === 0) {
+      return (
+        <div className="max-w-readable">
+          <p>There are currently no upcoming events.</p>
+        </div>
```
mirhamasala, line 153:
```suggestion
          <p>There are currently no upcoming events.</p>
```

src/app/_components/OrbitEventsSection.tsx
```diff
@@ -0,0 +1,167 @@
...
+    .select({
+      fields: Object.values(AIRTABLE_ORBIT_BASE.EVENTS_TABLE.FIELDS),
+      returnFieldsByFieldId: true,
+      sort: [
+        {
+          field: AIRTABLE_ORBIT_BASE.EVENTS_TABLE.FIELDS.START_DATE,
+          direction: 'asc',
+        },
+      ],
+      filterByFormula: `AND(
```
mirhamasala, line 64: Filter before sorting

src/app/_components/OrbitEventsSection.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  return validatedRecords.map((event) => ({
+    title: event[TITLE],
+    city: event[CITY],
+    startDate: event[START_DATE],
+    registrationLink: event[REGISTRATION_LINK],
+  }))
+}
+
+type OrbitEventsGridProps = {
```
mirhamasala, line 83:
```suggestion
type OrbitEventsProps = {
```

src/app/_components/OrbitEventsSection.tsx
```diff
@@ -0,0 +1,167 @@
...
+import { FilterContainer } from '@/components/FilterContainer'
+import { NoResultsMessage } from '@/components/NoResultsMessage'
+import { Pagination } from '@/components/Pagination'
+import { Search } from '@/components/Search'
+
+import { NextServerSearchParams } from '@/types/searchParams'
+
+import { formatDate } from '@/utils/formatDate'
+
+import { FILECOIN_FOUNDATION_URLS } from '@/constants/siteMetadata'
```
mirhamasala, line 19:
```suggestion
import { FILECOIN_FOUNDATION_URLS } from '@/constants/siteMetadata'

import { fetchAndParseAirtableEvents } from './_api/airtable'

```

CharlyMartin: I've added the airtable-related code into `orbit/services/airtable.ts`.

I didn't name it `api` because it used to be a reserved keyword, and it might imply that this file creates an API endpoint.

I hope it makes sense.

- https://nextjs.org/docs/app/building-your-application/routing/route-handlers
- https://nextjs.org/docs/pages/building-your-application/routing/api-routes

CharlyMartin:
@mirhamasala I've added your first round of feedback 🙂

mirhamasala (approved):
@CharlyMartin - There's one comment from Barbara that we need to look into. Other than that, this is wonderful and good to go! ☀️

src/app/_components/Search.tsx
```diff
@@ -16,7 +16,7 @@ import { PAGE_KEY, SEARCH_KEY } from '@/constants/searchParams'
 
 export type SearchProps = {
   query: ReturnType<typeof useSearch>['searchQuery']
-  id: 'mobile-search' | 'web-search'
+  id: string
```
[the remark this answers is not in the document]
CharlyMartin: I see. I reverted back to `string` because this `Search` component is used for both desktop and mobile, so neither of these options is accurate.

Do you see what I mean @barbaraperic? Do you have a better idea of how to make it work, maybe?

PR #268, opened 2024-07-02
New footer layout

barbaraperic:
## Visuals
Screenshot 2024-07-17 at 11 12 09
Screenshot 2024-07-17 at 11 12 02
Screenshot 2024-07-17 at 11 11 55


CharlyMartin:
Shouldn't the footer button read `Subscribe to The Upload`?

Instead of `Subscribe to Newsletter` currently.

---

CleanShot 2024-07-17 at 17 08 54

CharlyMartin:
I added a few CRs to (try to) simplify and harmonize the code, but overall it looks good :)

CharlyMartin (approved):
I added a few CRs to (try to) simplify and harmonize the code, but overall it looks good :)

mirhamasala:
>Shouldn't the footer button read Subscribe to The Upload?

@CharlyMartin Where are you seeing that? [>>](https://www.figma.com/design/0pBfQs5tKI6bx5ypPIfVRA/FIL.org-V4-Prototypes?node-id=168-6670&t=9m9pLfwHdkbjNwsT-4)

src/app/_components/MobileNavigation.tsx
```diff
@@ -73,23 +79,17 @@ type NestedMenuItemProps = {
 
 type NestedMenuProps = {
   title: string
-  items: NestedMenuItemProps[]
+  items: Array<NestedMenuItemProps>
```
mirhamasala, line 82: @CharlyMartin ~ We recently went in and changed all of this to `[]` - Barbara mentioned she prefers the simplicity of it - And I think I agree. Any reason for the `Array` use?

src/app/_components/MobileNavigation.tsx
```diff
@@ -73,23 +79,17 @@ type NestedMenuItemProps = {
...
-          <LinkItem
-            key={item.path}
-            nested
-            label={item.label}
-            path={item.path}
-            setOpen={setOpen}
-          />
+      <ul className="items-center space-y-6 border-l">
+        {items.map((item, i) => (
+          <LinkItem key={i} nested {...item} setOpen={setOpen} />
```
mirhamasala, line 92: @CharlyMartin - Another one - I think O prefer seeing what's passed down..

mirhamasala (approved):
@CharlyMartin - Thanks so much for the extra review - This was cool doing it all together. Let's get this merged @barbaraperic (We can go without the Career link for now)! ✨

CharlyMartin:
> @CharlyMartin Where are you seeing that? [>>](https://www.figma.com/design/0pBfQs5tKI6bx5ypPIfVRA/FIL.org-V4-Prototypes?node-id=168-6670&t=9m9pLfwHdkbjNwsT-4)

@mirhamasala The FIl+ mockup on Figma has a footer with `Subscribe to The Upload`, other pages don't. So it must be a mistake :)

src/app/_components/MobileNavigation.tsx
CharlyMartin: Ah my bad, I didn't know.

I prefer `Array` now for the readability aspect. I read [this article](https://tkdodo.eu/blog/array-types-in-type-script) a while ago about it.

> We usually read from left to right, so the things that are more important should come first. We say: "this is an array of strings"

Happy with `[]` too!

src/app/_components/MobileNavigation.tsx
CharlyMartin: Not sure I have a strong preference for either, it just made sense in the moment :)

src/app/_components/MobileNavigation.tsx
mirhamasala: Thank you for sharing the article - Like the say - Let's implement at least https://typescript-eslint.io/rules/array-type/

mirhamasala: And let's create a ticket to talk about the article - It makes sense to me to move on to your notation.

barbaraperic: Happy to go with `Array`, my comment before was about using one format. Thanks for sharing @CharlyMartin!

PR #271, opened 2024-07-03
Orbit Page - Orbit Program Goals Section

mirhamasala:
This PR styles the Orbit Program Goals Section. Using the `FocusAreaCard` for now - This will be refactored once we pick up the Card Refactoring ticket.

Note: We also adapted spacing for Cards that don't have a CTA in https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/271/commits/ff45f51fe2aa7f3b389e93b80bf61095d9cc00e7 to remove unnecessary spacing when not present.

---

[Notion](https://www.notion.so/filecoin/FF-Orbit-Page-Orbit-Program-Goals-Section-105898a5628b4e129d7fa393e0910d40?pvs=4) | [Figma](https://www.figma.com/design/0pBfQs5tKI6bx5ypPIfVRA/FIL.org-V4-Prototypes?node-id=2932-35120&t=kitr4iBq2hbfmZlI-4)

---

Screenshot 2024-07-03 at 08 13 36


src/app/_components/Card.tsx
```diff
@@ -165,7 +165,7 @@ export function Card({
           title
         )}
 
-        <div className="mb-10 mt-3">
+        <div className={clsx('mt-3', cta && 'mb-10')}>
```
barbaraperic, line 168: not sure if we need mb-10? but we can leave that for refactor

barbaraperic (approved):
Looks good 👌

src/app/_components/Card.tsx
mirhamasala: We need it for the Cards with CTA.

PR #273, opened 2024-07-03
Orbit Page - Style Orbit Program Features & Perks section

mirhamasala:
This PR styles the Orbit Program Features & Perks section.

Notable changes:

- `Badge` component (Renamed previous `Badge` component to `TagLabel`)
- Refactor cards with badges into `CardWithBadge`
- Refactor grids with card with badges into `BadgeCardGrid` (decided not to extend `CardGrid` since at this point there weren't enough commonalities)

[Notion](https://www.notion.so/filecoin/FF-Orbit-Page-Orbit-Program-Features-Perks-Section-10d55a44417842528ca75c0bab1d1266?pvs=4) | [Figma](https://www.figma.com/design/0pBfQs5tKI6bx5ypPIfVRA/FIL.org-V4-Prototypes?node-id=2932-35583&t=riAEnkc6VN2vn4hV-4)

---

Screenshot 2024-07-03 at 11 56 17


src/app/_components/NumberBadge.tsx
```diff
@@ -0,0 +1,13 @@
...
+  number: number
+}
+
+export function NumberBadge({ number }: NumberBadgeProps) {
+  return (
+    <BadgeWrapper>
+      <span className="text-2xl font-bold">{number}</span>
+    </BadgeWrapper>
+  )
+}
```
barbaraperic, line 13: `IconBadge` and `NumberBadge` are quite similar - would it make sense to have one component - `Badge`?

src/app/_components/BadgeCardGrid.tsx
```diff
@@ -0,0 +1,22 @@
...
+  smTwoLgFour: 'sm:grid-cols-2 lg:grid-cols-4',
+}
+
+export function BadgeCardGrid({ cols, children }: BadgeCardGridProps) {
+  return (
+    <ul className={clsx(baseGridStyles, extendedGridStyles[cols])}>
+      {children}
+    </ul>
+  )
+}
```
barbaraperic, line 22: Wondering if it would make sense to move the `BadgeCardGrid` component in the same file as `CardWithBadge`? Since it's specifically made for `CardWithBadge` layout, and I assume we're not gonna use it in other cases? Just an thought.

src/app/_components/NumberBadge.tsx
mirhamasala: I did think of that – But I'm looking at creating more components, having a more modularized codebase. Also, if we had `Badge` it would effectively have two props and both would be optional. Just something I wanted to give a go.

src/app/_components/BadgeCardGrid.tsx
mirhamasala: Same [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/273/files#r1664258302) 🙏🏼

barbaraperic (approved):

src/app/_components/NumberBadge.tsx
mirhamasala: Actually! I just had an idea. Give me a moment ;-)

src/app/_components/BadgeCardGrid.tsx
mirhamasala: Let's keep these concerns separate for now!

src/app/_components/NumberBadge.tsx
mirhamasala: Sometimes we do odd things. Thank you. Done!

PR #281, opened 2024-07-04
BUG FIX: PageHeader layout on sm screens

barbaraperic:
### Homepage
https://github.com/FilecoinFoundationWeb/filecoin-foundation/assets/50910606/9e3520c1-26f5-4b9a-a2c1-3298c3d1af83

### Orbit page

NOTE: There's still a slight glitch on the Orbit page. It's cause the text is too big on the second button. What we could do is rename it to `Orbit Ambassador's Portal` and double check with Filipa?  

https://github.com/FilecoinFoundationWeb/filecoin-foundation/assets/50910606/06e9043b-fdb9-423d-99d1-95f553c020d1


mirhamasala (commented):
I'm afraid that adapting text length isn't best as a longterm solution.

Though, I do think we need to talk to Filipa about adapting `padding` for different breakpoints - We do have a ticket open for this and I added a [comment in Figma](https://www.figma.com/design/0pBfQs5tKI6bx5ypPIfVRA?node-id=75-118#869138664).

I also asked Filipa to look into the spacings between buttons but it’s not a blocker [>>](https://www.figma.com/design/0pBfQs5tKI6bx5ypPIfVRA?node-id=52-17217#869157046)

I pushed a CR in which I removed the `whitespace` property. The rest of it is mostly a scouting refactor. Let me know what you think. 🌈

barbaraperic:
Feels a bit odd on smaller screens...?

Screenshot 2024-07-10 at 14 18 55

mirhamasala:
@barbaraperic - We’re currently designing and developing for six different breakpoints. The current solution looks good for all of them. It’s okay if some aspects aren’t ideal when resizing; the important thing is that all content remains visible and no elements break the layout. In the near future, we’ll collaborate with Filipa to redesign some components and layouts for different breakpoints, which will help improve the overall design. Thank you for reviewing –merging! 🚀

PR #282, opened 2024-07-08
Search input bugfix

barbaraperic:
## Note
- removed the native Safari styling of input element
- [Notion](https://www.notion.so/filecoin/FF-Bug-Search-bar-has-rounded-corners-on-mobile-safari-5ced4a4afa334df1af10fec4eaecdd95?pvs=4)

### Safari mobile 16.3.1

IMG_3609


mirhamasala (changes requested):
Though, this may fix the problem, it's not very scalable or extensible. I would look into using Tailwind CSS form plugin or extending the `components` field at `tailwind.config.ts`. Happy to pair on this so you can share your screen and we can have a look live.

barbaraperic:
> Though, this may fix the problem, it's not very scalable or extensible. I would look into using Tailwind CSS form plugin or extending the `components` field at `tailwind.config.ts`. Happy to pair on this so you can share your screen and we can have a look live.

@mirhamasala great suggestion! I added typescript forms plugin, adding `form-input `class to the input removes all native styles.

mirhamasala (approved):

PR #321, opened 2024-07-11
CMS Config 

mirhamasala:
This PR sets a max. character limit for Ecosystem Explorer project entry's `short description` field to 160.

It also refactors the `image_config` and disallows uploading multiple images. Finally, it includes better hints for the `SEO` and `Image URL` and `Alt Text` fields.

### DIFF
--- a/src/app/_data/cmsConfigSchema.json
+++ b/src/app/_data/cmsConfigSchema.json
@@ -24,6 +24,27 @@
       }
     ]
   },
+  "image_config": {
+    "name": "image",
+    "label": "Image",
+    "widget": "object",
+    "fields": [
+      {
+        "name": "url",
+        "label": "URL",
+        "widget": "image",
+        "allow_multiple": false,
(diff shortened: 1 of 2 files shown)


src/app/_data/cmsConfigSchema.json
```diff
@@ -24,6 +24,27 @@
...
+    "name": "image",
+    "label": "Image",
+    "widget": "object",
+    "fields": [
+      {
+        "name": "url",
+        "label": "URL",
+        "widget": "image",
+        "allow_multiple": false,
+        "hint": "Use https://squoosh.app/ to compress images before uploading. Choose the WebP format and resize the image to 2000px width."
```
barbaraperic, line 37: Should this be 200px width?

mirhamasala: No – The largest size is `1000px` - 2x for Retina to be on the safe side.

PR #323, opened 2024-07-16
Add Orbit and Security pages tests

mirhamasala:
This PR includes test for the Orbit and Security pages. It also add a property `hasSubpaths` to the `PathObjectParams` so that we can nest the Markdown content paths and match the folder structure of Next.js app router.


cypress/e2e/security_pages_spec.cy.ts
```diff
@@ -0,0 +1,12 @@
...
+import { testPageMetadata } from '../support/test-utils'
+
+describe('Security - Main Page', () => {
+  testPageMetadata({ path: PATHS.SECURITY, overrideDefaultTitle: true })
+})
+
+describe('Security - Coordinated Disclosure Policy Page', () => {
+  testPageMetadata({
+    path: PATHS.COORDINATED_DISCLOSURE_POLICY,
+  })
```
CharlyMartin, line 11: It looks like it's missing `overrideDefaultTitle: true` in the test?

It's used in the `createMetadata` for this page.

```suggestion
  testPageMetadata({
    path: PATHS.COORDINATED_DISCLOSURE_POLICY,
    overrideDefaultTitle: true,
  })
```

src/app/_constants/paths.ts
```diff
@@ -43,19 +44,23 @@ function createPathObject({
   includesEntries = false,
   customMainContentPath,
   customEntriesContentPath,
+  hasSubpaths = false,
 }: PathObjectParams): PathConfig {
-  const config: PathConfig = {
+  const mainContentPath = hasSubpaths
+    ? `${CONTENT_PAGES_ROOT}${path}${path}`
```
CharlyMartin, line 50: I get what you want to do, but it took me a while to understand the line below.

`${CONTENT_PAGES_ROOT}${path}${path}`

If we want to _match the folder structure of Next.js app router_, how about we do this:

`${CONTENT_PAGES_ROOT}${path}/index.md`, where `index.md` is like the `page.tsx`.

If it requires big changes for this PR, nevermind for now.

src/app/_constants/paths.ts
```diff
@@ -43,19 +44,23 @@ function createPathObject({
   includesEntries = false,
   customMainContentPath,
   customEntriesContentPath,
+  hasSubpaths = false,
 }: PathObjectParams): PathConfig {
```
CharlyMartin, line 48: The object parameter for this function is getting big and a bit confusing. 

The first two parameters are required and related to routing, the rest are optional and related to the CMS configuration. So perhaps we can rework the parameters as below?

```ts
function createPathObject(path, label, cmsConfig) {
  const {
    includesEntries = false,
    customMainContentPath,
    customEntriesContentPath,
    hasSubpaths = false,
  } = cmsConfig;

  // ...
}
```

I feel like `includesEntries` and `hasSubpaths` make more sense when grouped under `cmsConfig` (not attached to this name, just a suggestion)

CharlyMartin (commented):
It took me a while to review this PR as I'm not too familiar with this area of the codebase.

I hope my comments make sense :)

src/app/_constants/paths.ts
mirhamasala: Yes, it would require to big of a change. Perhaps it would be too ambitious to match the folder structure of Next.js. More of an organizational choice to help group all security pages in this case.

src/app/_constants/paths.ts
mirhamasala: Yes, you are right.

cypress/e2e/security_pages_spec.cy.ts
mirhamasala: Yes! That's the reason. Thank you.

PR #326, opened 2024-07-17
Remove broken events links

barbaraperic:
## Note
- this PR removes the broken external links on events pages
- should probably go thru another round of tests once we have them in place


mirhamasala (approved):
YAY for better UX and SEO scores! 🤗

PR #333, opened 2024-07-18
Generate map markdown to data files (WIP)

barbaraperic:
## Note
- command to generate 4 files based on a specific `config.yml` collection -` cmsConfigSchema.js`, `getData.js`, `mapMarkdownToData.js `and `types`.
- I've used this for Digest, still WIP (needs refining), but we could use for future collections


barbaraperic:
No need for this generator

PR #334, opened 2024-07-18
Fix cancel icon on Search input

CharlyMartin:
This PR fixes the `X` icon style on the Search input.

Before this PR, the Search showed the native X icon on Chrome/Safari, and nothing on Firefox. Now, the behavior is the same on all browsers.

To achieve this, I first hid with CSS native `X` icon using `[&::-webkit-search-cancel-button]:appearance-none`. It's a non-standard feature but is [supported](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-search-cancel-button#browser_compatibility) by Chrome-based browsers and Safari, where it's needed.

Second, I added a custom `X` icon with the same features:
1. Only appears when the search input is filled
2. Clears the search input when clicked

It's keyboard accessible and comes with an aria-label.

CleanShot 2024-07-18 at 15 31 35

[Notion](https://www.notion.so/filecoin/FF-V4-BUG-Fix-clear-icon-colour-in-Search-bar-567190e23ac24139853ecc6e6668ab64?pvs=4)

### DIFF
--- a/src/app/_components/Search.tsx
+++ b/src/app/_components/Search.tsx
@@ -40,8 +40,7 @@ export function Search({ query, id }: SearchProps) {
     DEBOUNCE_DELAY,
   )
 
-  function handleSearchChange(event: React.ChangeEvent<HTMLInputElement>) {
-    const newValue = event.target.value
+  function handleSearchChange(newValue: string) {
     setValue(newValue)
 
     const existingPageParam = params.get(PAGE_KEY)
--- a/src/app/_components/SearchInput.tsx
+++ b/src/app/_components/SearchInput.tsx
@@ -33,7 +33,7 @@ export function SearchInput({
         </div>
         {searchQuery && (
           <button
-            className="absolute inset-y-0 right-0 flex items-center px-3 text-brand-300 peer-hover:text-brand-400 peer-focus:text-brand-100"
+            className="absolute inset-y-0.5 right-0.5 flex items-center rounded-md px-3 text-brand-300 focus:brand-outline peer-hover:text-brand-400 peer-focus:text-brand-100"
             aria-label="Clear search input"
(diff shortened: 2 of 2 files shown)


src/app/_components/Search.tsx
```diff
@@ -40,8 +40,7 @@ export function Search({ query, id }: SearchProps) {
     DEBOUNCE_DELAY,
   )
 
-  function handleSearchChange(event: React.ChangeEvent<HTMLInputElement>) {
-    const newValue = event.target.value
+  function handleSearchChange(newValue: string) {
```
mirhamasala, line 43: Honest curiosity: Why this change?

src/app/_components/SearchInput.tsx
```diff
@@ -24,15 +22,24 @@ export function SearchInput({
...
+          <button
+            className="absolute inset-y-0.5 right-0.5 flex items-center rounded-md px-3 text-brand-300 focus:brand-outline peer-hover:text-brand-400 peer-focus:text-brand-100"
+            aria-label="Clear search input"
+            onClick={() => onSearchChange('')}
+          >
+            <Icon component={X} size={16} weight="bold" />
+          </button>
+        )}
       </div>
     </div>
```
mirhamasala, line 44: Would you mind walking @filipagr through the implementation in a screen share - and show her the different states? My only comment is that we might want the focus state to look a bit more square like the menu button. 🙏🏼 Pending Filipa's approval, good to go.

CharlyMartin: Makes sense! I'm gonna push a CR to make it square. Thank you!

CharlyMartin: @filipagr's quiet approval 

CleanShot 2024-07-19 at 15 33 37

PR #335, opened 2024-07-18
Fix Search with apostrophe

CharlyMartin:
This PR improves the search experience by stripping away special characters and most diacritical marks - a fancy word chatGPT taught me - which refers to any mark attached to a letter, such as:
- accents (_résumé_)
- tilde (_jalapeño_)
- diaeresis (_naïve_)
- ... you get the point

To do so, it "slugifies" both the source value and the query to ensure they are in the same format before looking for matches.

It solves the original problem: searching with an apostrophe variant in the query would return nothing. Now both `democracy's` and `democracy’s` return the same results.

CleanShot 2024-07-19 at 12 05 06

CleanShot 2024-07-19 at 12 10 34

[Notion](https://www.notion.so/filecoin/FF-BUG-Fix-search-with-apostrophe-65d912df28cb4310bd85e3b5f42e7464?pvs=4)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,6 +34,7 @@
         "sass": "^1.69.5",
         "schema-dts": "^1.1.2",
         "sharp": "^0.33.4",
+        "slugify": "^1.6.6",
         "swr": "^2.2.5",
         "tailwindcss": "^3.4.0",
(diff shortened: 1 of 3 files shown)


mirhamasala (approved):
So clean! 🧼

PR #363, opened 2024-07-19
Add hover state with semi-transparent background to cards.

filipagr:
- Added brand-400 with 15% opacity to the background on hover
- Applied transition effect for smooth hover interaction


mirhamasala:
Closing: Let's revisit once we get here.

PR #364, opened 2024-07-19
Fixes for the generate:page command

barbaraperic:
## Note
- this PR includes the following fixes
1. removes ";" from template files 
2. updates in the paths file as we've changed the `createPathObject`
- the command now also generates a new field on the `config.yml` file

## How to test
- run the command and check if the files are generated correctly


mirhamasala (approved):
Cool! ❤️

PR #370, opened 2024-07-19
Fix redirects with different casing

CharlyMartin:
This PR adds a redirect from `/blog/Filecoin-nv19-Lightning-Upgrade` to `/blog/filecoin-nv19-lightning-upgrade` and ensures that any blog URL with capital letters is redirected to its lowercase equivalent.

### Problem
The standard redirect solution - `{ source: 'About', destination: '/about' }` in `next.config.js` - did not work because redirects are not case-sensitive. It created an infinite loop of redirects and eventually crashed the website.

### Solution
To solve this problem, I implemented Next's [middleware](https://nextjs.org/docs/app/building-your-application/routing/middleware) function, which runs after redirects but before the filesystem routes.

The middleware function lives at the root of the project, `src/middleware.ts`, and runs by default on each request. To limit the middleware to only blog URLs containing capital letters, I added a [matcher](https://nextjs.org/docs/app/building-your-application/routing/middleware#matcher).

```ts
export const config = {
  matcher: '/blog/([A-Za-z0-9-]*[A-Z][A-Za-z0-9-]*)',
}
```

Now `/blog/Filecoin-nv19-Lightning-Upgrade` redirects to `blog/filecoin-nv19-lightning-upgrade`.

Additionally, I cleaned the redirects in `next.config.js` by grouping them with this [commit](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/370/commits/1e6abed91d9f02d749683396156a72e3e886ef9d). It's not essential to this PR and can be discarded if necessary.

---

[Notion](https://www.notion.so/filecoin/FF-BUG-Fix-redirects-for-same-URLs-with-different-casing-be287fccf1b04d69a0950a5a716cb315?pvs=4)

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -28,6 +28,7 @@ const nextConfig = {
   },
   async redirects() {
     return [
+      // BLOG POST REDIRECTS
       {
         source:
           '/blog/ecosystem-spotlight-ghostdrive%E2%80%99s-secure-decentralized-storage-now-on-mobile',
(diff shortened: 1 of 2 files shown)


mirhamasala (approved):
Squash them 🐛!

PR #397, opened 2024-07-22
Fix Ecosystem Project Video Iframe Bug - UXIT-1275

CharlyMartin:
This PR improves a few problems on the ecosystem project page:

- it fixes the weird iframe bug
- it improves the video performance by using [react-lite-youtube-embed](https://github.com/ibrahimcesar/react-lite-youtube-embed) 
- It improves the video design and error state design (to be reviewed by @filipagr)

There are now 3 possible states for the YouTube video, depending on the video link validity:
1. Invalid link format
2. Valid format but broken link
3. Valid working link

### The video link doesn't have a valid format
http://localhost:3000/ecosystem-explorer/a-bad-format

In this case, we know there's no chance the video will work because `isUrlFormatValid` is `false`. Hence, instead of showing an `iframe` with an invalid `src` (which caused the weird bug), we show a `div` placeholder.

CleanShot 2024-07-23 at 10 11 52

### The video link is valid but broken
http://localhost:3000/ecosystem-explorer/a-broken-url

If the link format is valid, then we try fetching the metadata for this video using `fetchYouTubeVideoMetadata`. This is a workaround to avoid authenticating with the YouTube API to check if a given id exists.

If we're able to fetch the video's metadata, then the video link is working. If not, the link is broken, and we dynamically replace the default video gray "screen of death" with the same placeholder as above.

### The video link is valid
http://localhost:3000/ecosystem-explorer/archly

When `fetchYouTubeVideoMetadata` confirms the validity of the link, we render `LiteYouTubeEmbed` and pass the `videoId`. Since we have the video's metadata, we update the `title` prop with the video's actual title instead of the generic `YouTube video player`

CleanShot 2024-07-23 at 10 17 40

---

Side-note: Once we host the ecosystem project submit form on our side, we may consider only storing the video IDs in the Markdown entries. This way, we can remove the _Invalid link format_ step and inject the video ID into the URL we control instead of manipulating a fully formed URL coming from "outside."

[Notion #1](https://www.notion.so/filecoin/FF-BUG-Fix-Ecosystem-Explorer-iFrame-bug-7b47f8e887c74bb48c1747de13e44201?pvs=4)
[Notion #2](https://www.notion.so/filecoin/FF-Optimize-YouTube-iFrames-72813df85a324d6c9fc79ca846179d6d?pvs=4)


mirhamasala:
Fixes UXIT-1275

src/app/ecosystem-explorer/[slug]/services/youtube.ts
```diff
@@ -0,0 +1,25 @@
+import z from 'zod'
+
+const videoMetedataSchema = z.object({
```
Comment (mirhamasala) [inline src/app/ecosystem-explorer/[slug]/services/youtube.ts:3]:
```suggestion
const videoMetadataSchema = z.object({
```

src/app/ecosystem-explorer/[slug]/services/youtube.ts
```diff
@@ -0,0 +1,25 @@
...
+  thumbnail_url: z.string(),
+  html: z.string(),
+})
+
+export async function fetchYouTubeVideoMetadata(videoId: string) {
+  return fetch(
+    `https://www.youtube.com/oembed?url=http://www.youtube.com/watch?v=${videoId}&format=json`,
+  )
+    .then((response) => response.json())
+    .then(videoMetedataSchema.parse)
```
Comment (mirhamasala) [inline src/app/ecosystem-explorer/[slug]/services/youtube.ts:24]:
```suggestion
    .then(videoMetadataSchema.parse)
```

src/app/ecosystem-explorer/[slug]/services/youtube.ts
Comment (mirhamasala) [inline src/app/ecosystem-explorer/[slug]/services/youtube.ts:1]: I think we shouldn't nest this deep - Having a `services` folder under `ecosystem-explorer` is nice.

src/app/ecosystem-explorer/[slug]/components/VideoSection.tsx
Comment (mirhamasala) [inline src/app/ecosystem-explorer/[slug]/components/VideoSection.tsx:1]: Same - Feel like `src/app/ecosystem-explorer/components/VideoSection` is alright!

src/app/ecosystem-explorer/[slug]/components/VideoSection.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+import { fetchYouTubeVideoMetadata } from '../services/youtube'
+
+const YOUTUBE_EMBED_BASE_URL = 'https://www.youtube.com/embed/'
+
+type YouTubeEmbedProps = {
+  videoUrl: string
+}
+
+const removeMoreVideosSuggestionsOnPause = 'rel=0'
```
Comment (mirhamasala) [inline src/app/ecosystem-explorer/[slug]/components/VideoSection.tsx:14]:
```suggestion

type YouTubeEmbedProps = {
  videoUrl: string
}

const YOUTUBE_EMBED_BASE_URL = 'https://www.youtube.com/embed/'
const removeMoreVideosSuggestionsOnPause = 'rel=0'
```

src/app/ecosystem-explorer/[slug]/components/VideoSection.tsx
```diff
@@ -0,0 +1,52 @@
...
+  if (!isUrlFormatValid || error) {
+    return (
+      <div
+        className="flex aspect-video items-center justify-center rounded-lg bg-brand-700"
+        role="status"
+      >
+        <p>Video unavailable</p>
+      </div>
+    )
+  }
```
Comment (mirhamasala) [inline src/app/ecosystem-explorer/[slug]/components/VideoSection.tsx:39]: Did we ask @filipagr to look into the design for this? 🙏🏼

src/app/ecosystem-explorer/[slug]/services/youtube.ts
Comment (mirhamasala) [inline src/app/ecosystem-explorer/[slug]/services/youtube.ts:3]: Why are we having to create this Schema ourselves?

src/content/ecosystem-explorer/projects/a-broken-url.md
mirhamasala, line 1: There's a flash of the "YT error frame" before it shows our `Video unavailable` element  - Try refreshing to reproduce.

src/app/ecosystem-explorer/[slug]/services/youtube.ts
Comment (CharlyMartin) [reply src/app/ecosystem-explorer/[slug]/services/youtube.ts:1]: Why not. We already have`src/app/ecosystem-explorer/[slug]/utils/generateStructuredData.ts` nested, so I kept the same logic.

I feel it makes sense to keep the code as close as possible to where it's used and move it up as needed.

src/app/ecosystem-explorer/[slug]/components/VideoSection.tsx
Comment (CharlyMartin) [reply src/app/ecosystem-explorer/[slug]/components/VideoSection.tsx:38]: Not yet, but I'm planning on it :)

src/content/ecosystem-explorer/projects/a-broken-url.md
CharlyMartin: Yes, I know.

The flash happens because we are optimistically rendering the video element first - hoping the link works - as opposed to showing nothing until we know it works.

Would you prefer to show nothing first? Or a loading state?

I don't find this implementation problematic for now. Most ecosystem projects don't have videos. And most, if not all, videos work. So this is a tiny corner case. But let me know :)

src/app/ecosystem-explorer/[slug]/components/VideoSection.tsx
Comment (CharlyMartin) [reply src/app/ecosystem-explorer/[slug]/components/VideoSection.tsx:38]: https://filecoinfoundation.slack.com/archives/C06MTEQ3SP6/p1721810299369289

src/content/ecosystem-explorer/projects/a-broken-url.md
mirhamasala: Yes to loading state!

src/app/ecosystem-explorer/[slug]/services/youtube.ts
Comment (mirhamasala) [reply src/app/ecosystem-explorer/[slug]/services/youtube.ts]: All good then!

src/app/ecosystem-explorer/[slug]/services/youtube.ts
Comment (mirhamasala) [reply src/app/ecosystem-explorer/[slug]/services/youtube.ts:1]: Ah, yes. You are right. It's different for the pages that have dynamic routes. All good then!

src/content/ecosystem-explorer/projects/a-broken-url.md
CharlyMartin: ✅

PR #423, opened 2024-07-24
[UXIT-1351] Encrypt Project Email Addresses

CharlyMartin:
This PR introduces encryption for email addresses and names saved in Markdown files.

Here's the breakdown:

### Next
1. Adds `encrypt` & `decrypt` utility functions using AES encryption with a secret key pulled from environment variables. Check 1Password to add it to your env
2. Creates a `POST /api/encryption/` endpoint to manage encryption and decryption requests

### CMS
1. Adds a name and email field in the config that uses a custom widget to decrypt values and display them in clear text
2. Introduces a `preSave` event listener to encrypt values before writing them to the Markdown files

### Notes
The API endpoint is necessary because the CMS admin file is in the `public` folder, outside of `src`. We can't import the `encrypt` & `decrypt` utility functions there. We might want to add a layer of security to it by whitelisting origins, for instance.

### How to test
1. Make sure you have `ENCRYPTION_SECRET_KEY` set in `.env` locally. It can be any string value, like `filecoin`.
2. Start the CMS locally and click on an ecosystem project entry
3. Enter an email value, like `charly@hey.com` and click Publish
4. Check the source Markdown file and make sure the `email` field is encrypted, it should look like this `encrypted::U2...`
5. Refresh the CMS form and make sure you still see the email in clear

### Todo before merging
- [x] ⚠️ Delete `1-encrypted-test.md`
- [x] Set `ENCRYPTION_SECRET_KEY` as a Netlify variable
- [x] Set `ENCRYPTION_SECRET_KEY` as a GitHub variable
- [x] Migrate all email addresses to Markdown files (can also be done after merging I suppose)
- [x] Test on the preview link [successful](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/428/files#diff-8e706dde32939f2e711ae64ba1f37c9e03d87cea0d7e873b86f3edea462419c6R3)


public/admin/config.yml
```diff
@@ -339,6 +339,14 @@ collections:
       - name: title
         label: Company Name
         widget: string
+      - name: email
+        label: Email
+        widget: decrypted_widget
+        required: false # TODO: Remove this once all emails have been migrated to Markdown entries
```
mirhamasala, line 345:
```suggestion
        required: false
```

public/admin/config.yml
```diff
@@ -339,6 +339,14 @@ collections:
...
+      - name: full-name
+        label: Full Name
```
mirhamasala, line 347:
```suggestion
        label: Submitter's Full Name
```

public/admin/config.yml
```diff
@@ -339,6 +339,14 @@ collections:
       - name: title
         label: Company Name
         widget: string
+      - name: email
+        label: Email
```
mirhamasala, line 343:
```suggestion
        label: Submitter's Email
```

src/app/_utils/encryption.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export function decrypt(encryptedText: string) {
+  if (!encryptedText.startsWith(PREFIX)) {
+    throw new Error('Invalid encrypted text')
+  }
+
+  const encryptedData = encryptedText.slice(PREFIX.length)
+  const decrypted = CryptoJS.AES.decrypt(encryptedData, String(SECRET_KEY))
+  return decrypted.toString(CryptoJS.enc.Utf8)
+}
```
mirhamasala, line 24: Think of throwing errors for these two functions.

CharlyMartin: For encrypt function, if `encrypted` is not a string, throw an error.
What if it's an empty string?

public/admin/cmsEncryptionWidget.js
```diff
@@ -0,0 +1,143 @@
+const EMAIL_CMS_FIELD = 'email'
+const FULL_NAME_CMS_FIELD = 'full-name'
```
mirhamasala, line 2: Use Route to check field names

CharlyMartin: Use the GET route to return the name of the CMS fields using CMS utils functions.

mirhamasala (approved):
Love it! I pushed a CR - mostly nits. I do quite like having the try-catch for error isolation and detailed error messages. If all is good with the CR, please let's get this merged! 🍤❤️

PR #425, opened 2024-07-25
Set up React Hook Forms - UXIT-687

barbaraperic:


src/app/_components/Input.tsx
```diff
@@ -0,0 +1,54 @@
...
+  type: 'email' | 'password' | 'text'
+  value?: string
+  placeholder?: string
+  errors?: string
+  disabled?: boolean
+  onChange?: (e: React.ChangeEvent<HTMLInputElement>) => void
+  onBlur?: () => void
+}
+
+const Input = React.forwardRef<HTMLInputElement, InputProps>(
```
CharlyMartin, line 21: Do we still need to use `React.forwardRef` now we have a `ControlledInput`?

src/app/_components/ControlledInput.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+
+import { useFormContext, Controller, RegisterOptions } from 'react-hook-form'
+
+import Input, { InputProps } from './Input'
```
CharlyMartin, line 5:
```suggestion
import Input, { type InputProps } from '@/components/Input'
```

src/app/_components/Form.tsx
```diff
@@ -0,0 +1,30 @@
...
+} & Omit<React.ComponentProps<'form'>, 'onSubmit'>
+
+export default function Form<Type extends FieldValues>(props: FormProps<Type>) {
+  const { methods, children, onSubmit, className, ...rest } = props
+
+  return (
+    <FormProvider {...methods}>
+      <form
+        autoComplete="on"
+        className={className}
```
CharlyMartin, line 22: Nit: `className` can be part of `...rest`

src/app/_components/NewsletterForm.tsx
```diff
@@ -0,0 +1,64 @@
...
+          <ControlledInput
+            key={newsletterField.id}
+            name={newsletterField.id}
+            {...newsletterField}
+            errors={getError(
+              methods.formState.errors,
+              newsletterField.id as keyof FormType,
+            )}
+          />
+        ))}
```
CharlyMartin, line 57: I understand what you are doing, but it feels like it's hindering readability for not much gain, since there's only one input. Unless I'm missing something?

src/app/_components/NewsletterForm.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const { isSubmitting } = methods.formState
+
+  const onSubmit: SubmitHandler<FormType> = async (values: FormType) => {
+    await new Promise((resolve) => setTimeout(resolve, 2000))
+    methods.resetField('email')
+  }
+
+  const getError = (errors: FieldErrors<FormType>, name: keyof FormType) => {
+    return errors[name]?.message
+  }
```
CharlyMartin, line 42: Nit: Do these 2 functions need to be declated with `const`? We generally use the `function` keyword throughout the codebase.

src/app/_components/NewsletterForm.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+import { Button } from './Button'
+import ControlledInput from './ControlledInput'
+import Form from './Form'
+import { InputProps } from './Input'
+
+export type FormType = {
+  email: string
+}
+
```
CharlyMartin, line 18: it looks like we can infer the FormType from the schema, so that we avoid duplicating information.

What do you think?

```suggestion
export type FormType = z.infer<typeof NewsletterSchema>
```

CharlyMartin (changes requested):

CharlyMartin:
@barbaraperic I added my feedback. Let me know if it makes sense, happy to call again if necessary :)

barbaraperic:
> @barbaraperic I added my feedback. Let me know if it makes sense, happy to call again if necessary 🙂

Awesome, thank you for the great feedback! ✨

src/app/_components/NewsletterForm.tsx
CharlyMartin: To make this work properly, we have to remove `ZodSchema` from `export const NewsletterSchema`.

Otherwise, TypeScript cannot infer the type and gives us `FormType = any`.

src/app/_components/NewsletterForm.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { zodResolver } from '@hookform/resolvers/zod'
+import { FieldErrors, useForm } from 'react-hook-form'
+import { z, ZodSchema } from 'zod'
+
+import { Button } from '@/components/Button'
+import ControlledInput from '@/components/ControlledInput'
+import Form from '@/components/Form'
+
+export const NewsletterSchema: ZodSchema = z.object({
```
CharlyMartin, line 13:
```suggestion
export const NewsletterSchema = z.object({
```

src/app/_components/NewsletterForm.tsx
```diff
@@ -0,0 +1,59 @@
...
+    await new Promise((resolve) => setTimeout(resolve, 2000))
+    methods.resetField('email')
+  }
+
+  function getError(
+    errors: FieldErrors<FormType>,
+    name: Extract<keyof FormType, string>,
+  ): string | undefined {
+    return errors[name]?.message as string | undefined
+  }
```
CharlyMartin, line 39:
```suggestion
  async function onSubmit(values: FormType) {
    await new Promise((resolve) => setTimeout(resolve, 2000))
    methods.resetField('email')
  }

  function getError(
    errors: FieldErrors<FormType>,
    name: Extract<keyof FormType, string>,
  ) {
    return errors[name]?.message
  }
```

As I shared with Mirha, I usually don't use return types for functions as TypeScript is able to infer them for us. Sometimes they are not inferrable and we need to add them manually, but not in this case.

Do you agree? Would you prefer to have them?

src/app/_components/Input.tsx
```diff
@@ -0,0 +1,38 @@
...
+  label: string
+  error?: string
+} & React.ComponentProps<'input'>
+
+export function Input({ id, label, error, value = '', ...rest }: InputProps) {
+  return (
+    <div className="relative space-y-2">
+      <label className="text-sm font-bold text-brand-300" htmlFor={id}>
+        {capitalize(label)}
+      </label>
```
CharlyMartin, line 19: The design for labels have changed a little bit since you started working on it:
https://www.figma.com/design/0pBfQs5tKI6bx5ypPIfVRA/FIL.org-V4-Prototypes?node-id=5285-76307&t=Kp5DZQTQ4NoPKmrn-4

Also it would be nice to encapsulate labels in their own `InputLabel` component, since we will use them with other inputs as well.

BUT, I realise that `NewsletterForm` does not use labels, so it can be done in a separate PR also :)

src/app/_components/Input.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <div className="relative">
+        <input
+          id={id}
+          value={value}
+          className={`peer form-input relative block w-full min-w-48 rounded-lg border border-brand-300 bg-brand-800 px-4 py-3 pr-9 focus:brand-outline placeholder:text-brand-300 hover:border-brand-400 placeholder:hover:text-brand-400 focus:text-brand-100 placeholder:focus:text-brand-100 [&::-webkit-search-cancel-button]:appearance-none ${error && 'border-error'}`}
+          {...rest}
+        />
```
CharlyMartin, line 26: Mirha upgraded Headless UI so we can use their Input component now :)

https://headlessui.com/react/input
https://headlessui.com/react/fieldset

They even export a `Label` component, so my comment above might not be relevant anymore :)

CharlyMartin (commented):
This is starting to look good @barbara :)

CharlyMartin (approved):

src/app/_components/NewsletterForm.tsx
```diff
@@ -0,0 +1,58 @@
...
+            hideLabel
+            label="Email"
+            name="email"
+            type="email"
+            placeholder="Enter your email"
+            autoComplete="email"
+            error={getError(methods.formState.errors, 'email')}
+          />
+        </div>
+        <Button className="min-w-44" type="submit" disabled={isSubmitting}>
```
mirhamasala, line 52: Let's see if we can remove the width rule from here and use either flex or gird or a div wrapper to specify layout styles.

PR #433, opened 2024-07-29
[UXIT-1358] Send form data to GitHub to open PR

CharlyMartin:
This PR integrates with GitHub API so that user-submitted information about ecosystem projects opens a PR in our repo, like https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/463 or https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/462

### Details
The feature primarily consists of backend services that handle GitHub interactions.

These services are called in order when `submitProjectToGithub` is called after a form submission:
1. `getLatestCommitOnMain` - Retrieves the latest commit on the main branch
2. `createBlob` - Creates blobs from content
3. `createTree` - Creates a tree from specified blobs
4. `createCommit` - Creates a  commit associated with a tree
5. `createPR` - Creates a branch and opens a pull request

To authenticate the requests made to GitHub, I created a personal access token that expires on Sun, Aug 25 2024.
This is for testing purposes, we will need to create another one with a longer validity.

### How to test
1. Create a [personal access token](https://github.com/settings/tokens) and add the value to your env under `GITHUB_AUTH_TOKEN`. **The token should have access to the repo scope.**
6. Go to http://localhost:3000/ecosystem-explorer/project-form and fill the form with a project name and an image.
7. Submit the form and wait a few seconds - You should be able to click on `View Submission` to check the PR you just opened.

### Todo before merging

- [x] Create a long-lasting GitHub access token
- [x] Add `GITHUB_AUTH_TOKEN` to Vercel
- [x] Delete `app/ecosystem-explorer/project-form/page.tsx`


barbaraperic:
Nice, works well 👏 haven't looked into it, but it would be nice to add a label and maybe a reviewer - someone from the marketing team?

src/app/ecosystem-explorer/project-form/services/github/api.ts
```diff
@@ -0,0 +1,121 @@
...
+type Tree = {
+  path: string
+  mode: '100644'
+  type: 'blob'
+  sha: SHA
+}
+
+const owner = 'FilecoinFoundationWeb'
+const repo = 'filecoin-foundation'
+const baseBranch = 'main'
```
barbaraperic, line 18: should we use capitalised variables?

barbaraperic:
Cypress test have failed tho, something to look into - https://github.com/FilecoinFoundationWeb/filecoin-foundation/actions/runs/10285161846

```
  reason: 'end of the stream or a document separator is expected',
  mark: r {
    name: null,
    buffer: '\n' +
      '    title: FF test\n' +
      '    image:\n' +
      '      src: /assets/images/ff-test.jpg\n' +
      '  ---\n' +
      '\x00',
    position: 71,
    line: 4,
    column: 2
  }
}
```

src/app/ecosystem-explorer/project-form/services/github/api.ts
mirhamasala: Yes - Also feel like we might want to add these to a constants file somewhere.

src/app/ecosystem-explorer/project-form/page.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { useState } from 'react'
+
+import { Button } from '@/components/Button'
+
+import { submitProjectToGithub } from './services/github'
+import {
+  ALLOWED_IMAGE_FORMATS,
+  getFileFormat,
+  convertToBase64,
+} from './services/github/utils'
```
mirhamasala, line 12: I think this is one of those cases - Do we use `@` or?

src/app/ecosystem-explorer/project-form/page.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export default function ProjectForm() {
+  const [files, setFiles] = useState<FileList | null>()
+  const [projectName, setProjectName] = useState<string | undefined>('')
+
+  const [link, setLink] = useState<string | undefined>('')
```
mirhamasala, line 18: Are these the right types for these?

mirhamasala (approved):
So awesome! Really excited about this. 🙌🏼

CR is mostly nits. `catch-try` is there and I also moved all the types up in files - I find it quite confusing to have them throughout the code - something to discuss with Barbara as well next week so we can align on these practices.

There are a few comments but nothing serious.

I'll also open a PR to also add/fix some import orders.

mirhamasala:
@CharlyMartin I pushed a couple of CRs - mostly trying to better organize all the utils. Let me know what you think.

src/app/ecosystem-explorer/project-form/page.tsx
CharlyMartin: This was just a dummy page to test the integration. I will remove it when the PR is ready to merge.

src/app/ecosystem-explorer/project-form/page.tsx
CharlyMartin: I don't understand. Do you want to use `@` here?

CharlyMartin:
@mirhamasala Thanks for the review!

The `date-fns` is a nice addition, it makes everything more consistent 👌 

I don't mind having all the API calls and their types in separate files.

However, I'm torn about the `try-catch` statements everywhere, wrapping every function call in a `try-catch` clutters the code and reduces readability. We can have just one top-level `try-catch` for the same result (besides the custom error messages)

About some of the changes in `services/github/index.ts`, I feel like these new thin abstractions add unnecessary complexity as well. The `createNewTree` function takes 7 arguments, which tells me there's something off.

I don't want to undo what you've already changed so we can discuss it live :)

src/app/ecosystem-explorer/project-form/services/github/api.ts
CharlyMartin: I would usually agree with you, but because the GitHub client expects lowercase variable names, it's easier this way.

src/app/ecosystem-explorer/project-form/services/github/config/repoConfig.ts
```diff
@@ -0,0 +1,5 @@
+export const repoConfig = {
+  owner: 'FilecoinFoundationWeb',
+  repo: 'filecoin-foundation',
+  baseBranch: 'main',
+} as const
```
CharlyMartin, line 5: Does this mean we include `baseBranch` everywhere we spread this object?

src/app/ecosystem-explorer/project-form/services/github/api.ts
mirhamasala: I already changed it in a CR 😊

src/app/ecosystem-explorer/project-form/page.tsx
mirhamasala: I don't know - I thought that's what you said on a call last time?

src/app/page.tsx
```diff
@@ -24,6 +24,7 @@ import { PATHS } from '@/constants/paths'
 import { FILECOIN_URLS } from '@/constants/siteMetadata'
 import { BASE_ORGANIZATION_SCHEMA } from '@/constants/structuredDataConstants'
 
+console.log(PATHS.ECOSYSTEM_EXPLORER.entriesContentPath)
```
mirhamasala, line 27:
```suggestion
```

src/app/ecosystem-explorer/project-form/services/github/config/repoConfig.ts
mirhamasala: Yes - If that's a problem, we could create two objects ..?

mirhamasala:
>However, I'm torn about the try-catch statements everywhere, wrapping every function call in a try-catch clutters the code and reduces readability. We can have just one top-level try-catch for the same result (besides the custom error messages)

I'm not sure I agree with this - Multiple try-catch blocks allow for more granular error handling, providing context-specific error messages and recovery actions. It also improves debugging by logging errors closer to their source, making it easier to trace issues.

I also think that handling errors within the same function where they occur keeps related logic together, following the principle of separation of concerns. I think it can make the code easier to understand because all relevant operations, including error handling, are in one place.

mirhamasala:
>The createNewTree function takes 7 arguments, which tells me there's something off.

I agree - One potential solution is to group related parameters

```
type TreeParams = {
  latestCommitOnMain: any
  paths: {
    mediaFolder: string
    ecosystemFolder: string
    slug: string
  }
  content: {
    logo: { format: AllowedImageFormats }
    markdownBlob: any
    imageBlob: any
  }
}
```

Which would give us `createNewTree({ latestCommitOnMain, paths, content }: TreeParams)`

CharlyMartin:
> I'm not sure I agree with this - Multiple try-catch blocks allow for more granular error handling, providing context-specific error messages and recovery actions. 

@mirhamasala - As an experiment, I removed all the intermediary `try-catch` statements and only left a top-level one around `submitProjectToGithub`.

Here's what we see when I introduce an error in the `createCommit` function:

CleanShot 2024-08-13 at 12 48 46@2x

It seems clear enough to me where the error happened and potentially why it happened. The error bubbles up the call stack and is caught where we need it :)

My issue with the current setup is that it gives us a false sense of security, a false sense that the error is handled, while it's not. `handleError` does not recover from the error. It logs it and throws a new one, which doesn't add a lot of value to me.

Now, if you feel strongly about having this setup, let's keep it in place, but we need to keep in mind that we still need to recover from the error somewhere.

src/app/ecosystem-explorer/project-form/services/github/config/repoConfig.ts
CharlyMartin: It still works with the current setup, so it's not a problem it seems :)

mirhamasala:
>The error bubbles up the call stack and is caught where we need it.

Perfect @CharlyMartin - In that case, please go ahead with your implementation!

PR #458, opened 2024-08-04
[UXIT-1357] Newsletter form

barbaraperic:
## Note
- this PR includes new `Dialog` component which we use for notification status
- `env` variables are in 1password. 
-`NEXT_PUBLIC_NEWSLETTER_SUBSCRIPTION_API_URL=https://stoplight.io/mocks/beehiiv/v2/104190750/` is a mock server url, I'll replace with the real one once we're done testing

## How to test
- subscribing with a regular email should show a success message
- subscribing when there's a `throw Error` in the `postSubscription` should show a warning message
- test responsiveness

### success
https://github.com/user-attachments/assets/e1835805-d5fb-45db-9d0c-7aeb4033f777

### warning
https://github.com/user-attachments/assets/82358257-b569-47e9-bc2c-1214fdb4d5ad


src/app/_components/Footer.tsx
```diff
@@ -1,16 +1,15 @@
...
 
 import { footerNavigationItems } from '@/constants/navigationItems'
-import { FILECOIN_FOUNDATION_URLS } from '@/constants/siteMetadata'
 
 export function Footer() {
   return (
     <footer className="mt-16 flex flex-col gap-6">
       <hr />
-      <div className="grid grid-cols-1 gap-y-6 sm:grid-cols-5">
+      <div className="my-2 grid grid-cols-1 gap-y-6 sm:grid-cols-5">
```
mirhamasala, line 12: I think we might need to increase this so there's also some space when there's an error message.

src/app/_components/Icon.tsx
```diff
@@ -5,25 +5,31 @@ export type IconProps = {
...
 export function Icon({
   component,
   color = 'inherit',
   size = 24,
   weight = 'regular',
+  inheritHoverStyle,
 }: IconProps) {
   const Component = component
 
+  const className = `${colorStyles[color]} ${inheritHoverStyle ? ` hover:text-inherit` : ''}`
```
mirhamasala, line 29:
```suggestion
const className = clsx(colorStyles[color], {
  'hover:text-inherit': inheritHoverStyle,
});
```

src/app/_components/Icon.tsx
```diff
@@ -5,25 +5,31 @@ export type IconProps = {
   color?: keyof typeof colorStyles
   size?: number
   weight?: 'light' | 'regular' | 'bold'
+  inheritHoverStyle?: boolean
```
mirhamasala, line 8: Are there instances where we don't want this to happen?

mirhamasala (changes requested):
There's a weird line

mirhamasala:
@filipagr - Can you check the preview [>>](https://filecoin-foundation-si-git-4f42dc-filecoin-foundations-projects.vercel.app/?vercelToolbarCode=ysaxdsD6Zc7zUFm)
Screenshot 2024-08-12 at 17 07 43

Shall we wait with breaking into two columns until we reach `md` so that the input with subscribe button can fit well?

src/app/_components/NewsletterForm.tsx
```diff
@@ -49,12 +76,19 @@ export function NewsletterForm() {
             error={getError(methods.formState.errors, 'email')}
           />
         </div>
-        <div className="flex min-w-44 [&>*:first-child]:flex-1">
+        <div className="min-w-44 flex-shrink-0 flex-grow">
```
mirhamasala, line 79: @barbaraperic - I changed `[&>*:first-child]:flex-1` because this, to me, is still "reaching into the component."

src/app/_components/NewsletterForm.tsx
```diff
@@ -19,16 +21,41 @@ export const NewsletterSchema = z.object({
...
+function postSubscription(email: string): Promise<number> {
+  return fetch(NEWSLETTER_URL, {
+    method: 'POST',
+    headers: {
+      Authorization: AUTHORIZATION_HEADER,
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({ email }),
+  }).then((response) => response.status)
+}
```
mirhamasala, line 36: I extracted `postSubscription` to help separate concerns and keep the component logic clean.

src/app/_components/Dialog.tsx
mirhamasala, line 1: Should this be called something else than `Dialog`?

src/app/_components/Dialog.tsx
```diff
@@ -0,0 +1,54 @@
...
+    title: 'Successfully subscribed!',
+    icon: <Icon component={CheckCircle} color="success" />,
+  },
+  warning: {
+    title: 'An error has occurred. Please try again.',
+    icon: <Icon component={XCircle} color="warning" />,
+  },
+}
+
+export default function DialogComponent({
```
mirhamasala, line 26: Let's not use `default` exports..

mirhamasala:
@CharlyMartin - I did a first pass. Feel free to take over! 🚀

src/app/_components/Dialog.tsx
barbaraperic: I like `NotificationDialog` 👌

src/app/_components/Icon.tsx
barbaraperic: This would have impact on some other parts of the UI, thought it was safer if we explicitly pass a prop

https://github.com/user-attachments/assets/f335f77c-72cf-4a0f-b114-1953deb3b0c7

src/app/_components/NewsletterForm.tsx
mirhamasala: Oh 😞 I though this worked. Sorry. I can have a second look..

CharlyMartin:
@barbaraperic My initial feedback after trying the feature for the first time (I haven't looked at the code yet)
1. It works, nice 😄
2. When using a valid email, there's first a 400 response then a valid one. I'm not sure if it's related to the dummy endpoint we're using
3. There's a noticeable page jump when the toast toast appears
4. The toast message does not have a timer. These things usually disappear after some time, right?
5. Nit: There's no transition out when closing the toast message, even though there's one when it comes in.

Screenshare: https://share.cleanshot.com/Z99528vw

I hope it makes sense :)

src/app/_components/NotificationDialog.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+type NotificationDialogProps = {
+  isOpen: boolean
+  setIsOpen: (arg: boolean) => void
+  status?: StatusType
+}
+
+const dialogStatus = {
+  success: {
+    title: 'Successfully subscribed!',
```
CharlyMartin, line 17: I feel like `title` - the message of the toast - should be passed in as a prop intead of being hardcoded.

In theory, a success or error toast could be used in other contexts. The ecosystem project form might use it for instance :)

src/app/_components/NewsletterForm.tsx
```diff
@@ -49,12 +79,19 @@ export function NewsletterForm() {
...
           />
         </div>
-        <div className="flex min-w-44 [&>*:first-child]:flex-1">
+        <div className="min-w-44 flex-shrink-0 flex-grow">
           <Button type="submit" disabled={isSubmitting}>
             {isSubmitting ? 'Subscribing' : 'Subscribe'}
           </Button>
         </div>
       </div>
+      {isOpen && (
```
CharlyMartin, line 88: This line is probably why the toast doesn't transition out of the screen. When `false`, it gets unmounted instead of being animated out.

src/app/_components/NewsletterForm.tsx
```diff
@@ -19,16 +24,41 @@ export const NewsletterSchema = z.object({
...
+}
+
 export function NewsletterForm() {
   const methods = useForm<FormType>({
     resolver: zodResolver(NewsletterSchema),
   })
 
   const { isSubmitting } = methods.formState
+  const [isOpen, setIsOpen] = useState(false)
+  const [notificationStatus, setNotificationStatus] = useState<StatusType>()
```
CharlyMartin, line 48: These two states are related and should live together, meaning be merged into one state, don't you think?

src/app/_components/NewsletterForm.tsx
```diff
@@ -49,12 +79,19 @@ export function NewsletterForm() {
...
         </div>
       </div>
+      {isOpen && (
+        <NotificationDialog
+          status={notificationStatus}
+          isOpen={isOpen}
+          setIsOpen={setIsOpen}
+        />
+      )}
     </Form>
```
CharlyMartin, line 95: This might be out of scope / too much rework. So, let me know what you think 🙂

Wouldn't it be nice NOT to have to include the toast component here, and only call a function to control it?

Something like this:

```ts
try {
    await postSubscription(values.email)
    toast("Successfully subscribed!", { type: "success" })
} catch (error) {
    toast("Something went wrong!", { type: "error" })
    Sentry.captureException(error)
}
```

Or alternatively:

```ts
try {
    await postSubscription(values.email)
    toast.success("Successfully subscribed!")
} catch (error) {
    toast.error("Something went wrong!")
    Sentry.captureException(error)
}
```

tailwind.config.js
```diff
@@ -81,6 +81,21 @@ module.exports = {
       maxWidth: {
         readable: '60ch',
       },
+      animation: {
+        'slide-in-from-top':
+          'slide-in-from-top 0.3s cubic-bezier(.41,.73,.51,1.02)',
+        leave: 'leave 150ms ease-in forwards',
```
CharlyMartin, line 87: `leave` is very generic, we should probably rename it but I can't think of a better name right now.

GPT suggests `fadeOutShrink`, but I'm not a fan.

CharlyMartin (dismissed):

src/app/_components/NotificationDialog.tsx
barbaraperic: I see what you mean, but I feel like it's easier to have an overview of statuses in the component in the case we want to change the copy or structure, we have it in one place. We can be more specific and have success-subscription, success-registration if there's ever a need. @mirhamasala any thoughts?

barbaraperic:
> @barbaraperic My initial feedback after trying the feature for the first time (I haven't looked at the code yet)
> 
> 1. It works, nice 😄
> 2. When using a valid email, there's first a 400 response then a valid one. I'm not sure if it's related to the dummy endpoint we're using
> 3. There's a noticeable page jump when the toast toast appears
> 4. The toast message does not have a timer. These things usually disappear after some time, right?
> 5. Nit: There's no transition out when closing the toast message, even though there's one when it comes in.
> 
> Screenshare: https://share.cleanshot.com/Z99528vw
> 
> I hope it makes sense :)

Thanks @CharlyMartin!
1. 😊
2. I couldn't recreate, it might be the dummy endpoint. We'll keep an eye when we replace the endpoint.
3. I couldn't recreate - which browser are you using? 
4. It wasn't in the specs, but I'll add a 5s timer
5. Working on it!

tailwind.config.js
CharlyMartin: Also, maybe it's not a big deal, but the `leave` animation isn't working. The notification disappears all at once.

CharlyMartin:
> I couldn't recreate - which browser are you using?

I'm using Firefox, but it looks like it's fixed now :)

mirhamasala (changes requested):
- Include focus style close button

src/app/_components/NotificationDialog.tsx
mirhamasala: Agree with @CharlyMartin on this one.

barbaraperic:
**Include focus style close button** ✅

**Touch taget should be `48px x 48px`** - I've set it `24px x 24px`, like in Figma? 48px seemed way too big 

**Icon seems smaller on Figma** - ✅ Set 16px 

**Test Headless UI's close button** - ✅ think this worked for the animations, good catch

**Use prop to pass message** - ✅

**Can we look into `EXIT_ANIMATION_DURATION_MS` - The example on Headless doesn't seem to need it?** - ✅ Removed it, we don't need it. Think the CloseButton worked for exit animation, tho it only worked with their example `data-[closed]:transform-[scale(95%)]`, not with `animate:leave`. I left the class in tailwind in case you want to try it?

mirhamasala (approved):

CharlyMartin (approved):
Good to go :)

PR #474, opened 2024-08-07
[UXIT-1319] CMS - Fix Featured Entries

mirhamasala:
This PR addresses the handling of featured entries by ensuring that we correctly extract the slug from the provided file path. The featured entries field now uses the full file path, allowing us to dynamically parse and obtain the slug for further operations, ensuring compatibility with file-based CMS structures.

Screenshot 2024-08-07 at 10 48 27

Screenshot 2024-08-07 at 10 48 52

Screenshot 2024-08-07 at 10 48 11

Screenshot 2024-08-07 at 10 49 15

### DIFF
--- a/public/admin/config.yml
+++ b/public/admin/config.yml
@@ -369,17 +369,9 @@ collections:
       - name: title
         label: Company Name
         widget: string
-      - name: created-on
-        label: Created On
-        widget: datetime
-      - name: updated-on
-        label: Updated On
-        widget: datetime
-        required: false
-      - name: published-on
-        label: Published On
-        widget: datetime
-        required: false
+      - *created_on
+      - *updated_on
+      - *published_on
       - *image_config
       - name: category
         label: Category
--- a/public/admin/config.yml
+++ b/public/admin/config.yml
@@ -140,8 +140,8 @@ collections:
           - name: "featured_entry"
             label: "Featured Post"
             widget: "relation"
-            collection: "blog"
-            value_field: "slug"
+            collection: "blog_posts"
+            value_field: "src/content/blog/{{slug}}.md"
             search_fields:
               - "slug"
               - "title"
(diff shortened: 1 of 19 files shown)


public/admin/config.yml
```diff
@@ -140,8 +140,8 @@ collections:
           - name: "featured_entry"
             label: "Featured Post"
             widget: "relation"
-            collection: "blog"
-            value_field: "slug"
+            collection: "blog_posts"
```
CharlyMartin, line 143: We use both `_` and `-` for spaces. I think we should harmonize that.

src/app/_types/cmsConfig.ts
```diff
@@ -1,4 +1,4 @@
-export type CMSCollectionName = 'blog' | 'ecosystem' | 'events'
+export type CMSCollectionName = 'blog_posts' | 'ecosystem' | 'events'
```
CharlyMartin: Shouldn't `events` be `event-entries` now if we follow the same logic?

Also, I think `events` is a self-explanatory name.

mirhamasala: Yes, it should be `event_entries`. The problem with just using `events` is that we also have the Events page, which is also called `events` in the CMS config.

public/admin/config.yml
mirhamasala: I was trying to use `_` in the CMS config for fields related to names - This way, it's very easy to spot in our codebase that this is CMS data related. We can definitely look into harmonizing.

src/app/_types/cmsConfig.ts
mirhamasala: Actually, on my side it's already `event_entries` - Probably a CR that didn't get there in time.

CharlyMartin:
@mirhamasala I don't fully understand why these changes are needed but it works on my side!

mirhamasala:
> @mirhamasala I don't fully understand why these changes are needed but it works on my side!

At the moment, if you try to change the featured entries for the homepage, Blog page, Events page, or Grants page, in the CMS - It won't work.

CharlyMartin (approved):

PR #479, opened 2024-08-07
[UXIT-1411] - Form Components

CharlyMartin:
## 📝 Description

This PR adds all the form components necessary for the ecosystem project form [[Figma](https://www.figma.com/design/0pBfQs5tKI6bx5ypPIfVRA/FIL.org-V4-Prototypes?node-id=5196-68950&t=TUzl5LhG3xTIwhGU-4)].

The ecosystem project form will be built in a subsequent PR.

Intro video: https://share.cleanshot.com/kH9cs2l6

## 🛠️ Key Changes

1. Updates the text input & form label components
2. Adds a text area component and its controlled version - [Docs](https://headlessui.com/react/textarea)
3. Adds a checkbox component and its controlled version - [Docs](https://headlessui.com/react/checkbox)
4. Adds a select component (listbox) and its controlled version - [Docs](https://headlessui.com/react/listbox) (This also partially tackles this [ticket](https://www.notion.so/filecoin/FF-Update-HeadlessUI-components-to-new-props-API-c101e79bf68e42f9863f3b6e928d16b4?pvs=4))
5. Adds a file picker component and its controlled version - [Docs](https://headlessui.com/react/input)
6. Adds a radio input component - [Docs](https://headlessui.com/react/radio-group)
7. Adds a combobox component - [Docs](https://headlessui.com/react/combobox)
8. Adds a form error component

## 📌 To-Do Before Merging

- [x] Delete `src/app/storybook/page.tsx`
- [x] Harmonise error positioning. The `NewsletterForm` relies on `-mb-8` to offset the height of the error container.

## 🧪 How to Test

Visit `/storybook` and make sure all the components look and behave as specified on Figma

## 📸 Screenshots

CleanShot 2024-08-22 at 14 56 45

## 🔖 Resources

- [Headless UI](https://headlessui.com/)
- [React Hook form](https://react-hook-form.com/)
- [Zod Resolver](https://github.com/react-hook-form/resolvers?tab=readme-ov-file#zod)
- [Zod](https://zod.dev/)


barbaraperic:
You're setting some high standards on PR descriptions! 😀 thanks for the video, super helpful.

Some initial visual feedback, haven't looked at the code yet.

- FormRadioInput component checked state should be `brand-400`
- FormCheckbox checked state should be `brand-400`

- Would be good that FormCombobox is scrollable? You can't see all the options
 Screenshot 2024-08-23 at 09 19 44

- FormInput and FormTextArea have padding left and right 14px in Figma?
Screenshot 2024-08-23 at 09 30 51

- FormListbox seems broken on smaller screens
Screenshot 2024-08-23 at 09 34 20

- FormFileInput text should be 12px and there seems to be a bigger padding on smaller screens according to Figma
Screenshot 2024-08-23 at 09 34 20

src/app/_components/Form/ControlledFormCheckbox.tsx
```diff
@@ -0,0 +1,51 @@
...
+} from '@/components/Form/FormCheckbox'
+
+type ControlledProps =
+  | 'checked'
+  | 'defaultChecked'
+  | 'value'
+  | 'defaultValue'
+  | 'onChange'
+  | 'onBlur'
+  | 'required'
```
barbaraperic, line 20: Should we exclude 'required'? What if we have a case something like, "you need to agree with the Privacy Policy"

src/app/_components/Form/FormFileInput.tsx
```diff
@@ -0,0 +1,179 @@
...
+  function resetFiles() {
+    onChange(null)
+  }
+
+  function loadFiles(e: React.ChangeEvent<HTMLInputElement>) {
+    const files = e.target.files
+
+    if (files) {
+      onChange(files)
+    }
```
barbaraperic, line 52: should we handle if the user uploads > MAX_SIZE?

```suggestion
 if (files && files.length > 0) {
    if (files[0].size > maxSize) {
      // Handle file size error
      console.error('File size exceeds maximum allowed size');
      return;
    }
    onChange(files)
  }
```

src/app/_components/Form/FormFileInput.tsx
```diff
@@ -0,0 +1,179 @@
...
+  files: FileList | null
+  accept: Array<`.${string}`>
+  maxSize: number
+  description?: string | React.ReactNode
+  onChange: (files: FileList | null) => void
+} & Omit<HeadlessInputProps, HardCodedProps> &
+  FormLabelProps &
+  FormErrorProps
+
+export function FormFileInput({
```
barbaraperic, line 28: Maybe we should include keyboard accessibility? Currently you can only add a file by clicking

src/app/_components/Form/FormFileInput.tsx
```diff
@@ -0,0 +1,179 @@
...
+) {
+  const formats = formatsWithLeadingDot.map((format) => format.replace('.', ''))
+
+  if (formats.length === 0) {
+    return 'any file type'
+  }
+
+  if (formats.length === 1) {
+    return `${formats[0]} files`
+  }
```
barbaraperic, line 163: singular if it's 1 file? "upload .png file"

src/app/_components/Form/FormFileInput.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+function SelectedFile({ file, onReset }: SelectedFileProps) {
+  return (
+    <div className="h-full w-full">
+      <NextImage
+        fill
+        src={URL.createObjectURL(file)}
+        alt={file.name}
+        className="rounded-lg object-cover"
+      />
```
barbaraperic, line 109: If it's a wide photo it might `object-cover` might cut it... `object-contain`? We use that in ecosystem projects

src/app/_components/Form/ControlledFormInput.tsx
```diff
@@ -3,39 +3,49 @@ import React from 'react'
...
-  name: keyof FormType
-  rules?: RegisterOptions
-} & FormInputProps
+type ControlledProps =
+  | 'value'
+  | 'defaultValue'
+  | 'error'
+  | 'onChange'
+  | 'onBlur'
+  | 'required'
```
barbaraperic, line 18: Same as for the checkbox - do we want to omit `required`? Would we nice to pass it as a prop to ControlledFormInput if needed?

src/app/_components/Form/ControlledFormListbox.tsx
```diff
@@ -0,0 +1,52 @@
...
+        const value = field.value
+        const error = formState.errors[name]
+
+        return (
+          <FormListbox
+            {...rest}
+            value={value}
+            error={error && String(error.message)}
+            onChange={field.onChange}
+          />
```
barbaraperic, line 47: should we add `onBlur`?

src/app/_components/Form/ControlledFormTextarea.tsx
```diff
@@ -0,0 +1,54 @@
...
+      control={control}
+      render={({ field }) => {
+        const value = field.value ?? ''
+        const error = formState.errors[name]
+
+        return (
+          <FormTextarea
+            {...rest}
+            value={value}
+            error={error && String(error.message)}
```
barbaraperic, line 45: if `error` is undefined would this throw an error? 🤔

src/app/_components/Form/FormCombobox.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+export function FormCombobox<Value extends OptionType = OptionType>({
+  label,
+  hideLabel,
+  options,
+  ...props
+}: FormComboboxProps<Value>) {
+  const [query, setQuery] = useState<string>('')
+
+  const filteredOptions = filerOptions(options, query)
```
barbaraperic, line 41: typo? `filterOptions`?

barbaraperic:
For the `FormCombobox` - **not sure if this is an issue**, but I have a value and try to leave it empty, it doesn't allow it. 

https://github.com/user-attachments/assets/8b80e9fa-15fb-40d5-b3a7-de689dba1792

src/app/_components/Icon.tsx
```diff
@@ -1,10 +1,10 @@
-import { type Icon } from '@phosphor-icons/react'
+import type { Icon, IconWeight } from '@phosphor-icons/react'
 
 export type IconProps = {
   component: Icon
   color?: keyof typeof colorStyles
   size?: number
-  weight?: 'light' | 'regular' | 'bold'
+  weight?: IconWeight
```
barbaraperic, line 7: nice!

src/app/_components/Form/FormTextarea.tsx
```diff
@@ -0,0 +1,65 @@
...
+        {...rest}
+        invalid={Boolean(error)}
+        className={clsx(
+          'peer form-input block min-h-32 w-full rounded-lg border border-brand-300 bg-brand-800 px-4 py-3 focus:brand-outline placeholder:text-brand-300 hover:border-brand-400 placeholder:hover:text-brand-400 focus:text-brand-100 placeholder:focus:text-brand-100 data-[disabled]:cursor-not-allowed [&::-webkit-search-cancel-button]:appearance-none',
+          error && 'border-red-400',
+        )}
+      />
+      <div className="mt-2 flex h-6 items-center justify-between">
+        <FormError.Message error={error} />
+        <CharacterCounter count={characterCount} limit={charcterLimit} />
```
barbaraperic, line 43:
```suggestion
          <CharacterCounter count={characterCount} limit={characterLimit} />
```

src/app/_components/Form/FormTextarea.tsx
```diff
@@ -0,0 +1,65 @@
...
+  characterCount: number
+} & Omit<HeadlessTextareaProps, HardCodedProps> &
+  FormLabelProps
+
+export function FormTextarea({
+  error,
+  label,
+  hideLabel,
+  characterCount,
+  charcterLimit,
```
barbaraperic, line 27:
```suggestion
  characterLimit
```

src/app/_components/Form/FormTextarea.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { clsx } from 'clsx'
+
+import { FormError, type FormErrorProps } from '@/components/Form/FormError'
+import { FormLabel, type FormLabelProps } from '@/components/Form/FormLabel'
+
+type HardCodedProps = 'invalid' | 'className'
+
+export type FormTextareaProps = {
+  error?: FormErrorProps['error']
+  charcterLimit: number
```
barbaraperic, line 17:
```suggestion
  characterLimit: number
```

src/app/_components/Form/FormTextarea.tsx
```diff
@@ -0,0 +1,65 @@
...
+        <FormError.Message error={error} />
+        <CharacterCounter count={characterCount} limit={charcterLimit} />
+      </div>
+    </Field>
+  )
+}
+
+type CharacterCounterProps = {
+  count: FormTextareaProps['characterCount']
+  limit: FormTextareaProps['charcterLimit']
```
barbaraperic, line 51:
```suggestion
  limit: FormTextareaProps['characterLimit']
```

src/app/_components/Form/FormRadioInput.tsx
```diff
@@ -0,0 +1,50 @@
...
+}: FormRadioInputProps<Value>) {
+  return (
+    <RadioGroup<'div', Value>
+      {...rest}
+      as="div"
+      className="flex flex-wrap items-center gap-x-10 gap-y-6 data-[disabled]:cursor-not-allowed"
+    >
+      {options.map((option) => (
+        <Field
+          key={option.name}
```
barbaraperic, line 33: maybe `option.id` for a key?

barbaraperic:
@mirhamasala did the first go on the PR, over to you 🚀

src/app/_components/Form/ControlledFormCheckbox.tsx
CharlyMartin: Here's how it's currently set up:

For controlled components, the Zod schema (and the `zodResolver` integration) determines whether a field is required. For instance:

```ts 
const TestSchema = z.object({
  isCool: z.boolean({ message: 'This field is required' }),
```

For a privacy policy check, we can use the uncontrolled version of the component, in which case we can pass `required` as a prop.

Does that make sense?

src/app/_components/Form/FormCheckbox.tsx
```diff
@@ -0,0 +1,29 @@
...
+        {...rest}
+        className="group size-5 rounded bg-brand-100 p-0.5 text-brand-100 focus:brand-outline data-[disabled]:cursor-not-allowed ui-checked:bg-brand-500"
+      >
+        <span className="hidden group-data-[checked]:block">
+          <Icon component={Check} size={16} weight="bold" />
+        </span>
+      </Checkbox>
+      <Label className={clsx('text-brand-100', hideLabel && 'sr-only')}>
+        {label}
+      </Label>
```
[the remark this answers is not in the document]
CharlyMartin: I added it to make it consistent with the other components, but it's true. I'll remove it.

src/app/_components/Form/ControlledFormInput.tsx
CharlyMartin: Same comment as above

src/app/_components/Form/FormFileInput.tsx
CharlyMartin: Good idea, done!

src/app/_components/Form/FormFileInput.tsx
CharlyMartin: As discussed with Filipa, we need `object-cover` so that the image uses the full width and the close button is always above it.

src/app/_components/Form/ControlledFormTextarea.tsx
CharlyMartin: If `error` is `undefined`, then the `&&` operator doesn't evaluate the second part of the operation and returns `undefined`

CharlyMartin:
@barbaraperic thank you so much for your review 🙏

I've added most of your feedback and left some comments.

Some additional comments:
> Would be good that FormCombobox is scrollable? You can't see all the options

Yes, I agree. When I made it scrollable, the large number of options would make the navigation complicated, so I removed it.

> For the FormCombobox - not sure if this is an issue, but I have a value and try to leave it empty, it doesn't allow it.

Good catch! I played with it a little bit, and for some reason, it emptied the value on the second try. In our case, it should not be a problem as users will select their project and then move on to the form; I don't see why they would need to empty the field for now. But we should fix it at some point, nonetheless! Maybe in the next PR.

> FormFileInput text should be 12px

Indeed, but text `12px` or smaller is not really accessible. I made it 14px.

mirhamasala:
@CharlyMartin (CC: @barbaraperic) - Keeping track of some notes here besides the live pairing.

- I noticed the use of `<br />` for layout purposes, which is generally not considered best practice. It’s better to use CSS for controlling layout and spacing to maintain clean, semantic HTML and improve accessibility.
- Similarly, I noticed the use of `<strong>` for styling text to be bold. Since `<strong>` is meant to convey semantic importance rather than just visual emphasis, it’s preferable to use a `<span>` with appropriate CSS classes for styling when there’s no intended emphasis on the content. This approach keeps the HTML semantically correct and accessible.

CharlyMartin:
Thanks for the comments and live review @mirhamasala 🙏 

I've pushed an update with what we discussed this morning:
1. Extracting companion components in separate files (`FormFileInput`, `FormListbox` and `FormTextarea`)
2. Creating a `FormField` component handling the spacing for labels and errors. I couldn't use it everywhere, for instance in `FormTextarea`, but I made sure to export and reuse styles instead of duplicating.

mirhamasala (approved):
Whoa @CharlyMartin! 🤩 It's looking so good! Congrats! LGTM. ❤️

PR #483, opened 2024-08-09
UXIT-1363 - Filter Forms

barbaraperic:
## Description
- This PR includes 2 key things:
1. new sort functionality - aside from sorting by date, we can now also sort alphabetically
2. new `All` category which serves as a "count and reset category" - shows the total count and on click resets the filters

## How to test

1. Blog and Events page should have `All` category and no `Results` and `Reset Filter` buttons
2. Ecosystem Explorer page should have Alphabetical sort, and `Results` and `Reset Filter` buttons included

| Page              | All Category | `Results` and `Reset Filter` buttons | Sort type                        |
|-------------------|--------------|--------------------------------------|----------------------------------|
| Blog              | ✅            | ❌                                    | chronological (newest-oldest)    |
| Events            | ✅            | ❌                                    | chronological (newest-oldest)    |
| Ecosystem Explorer| ❌            | ✅                                    | alphabetical (az-za)             |


src/app/_components/FilterContainer.tsx
```diff
@@ -62,6 +62,18 @@ function ResultsAndCategory({ results, category }: ResultsAndCategoryProps) {
...
+  category,
+}: ResultsAndCategoryProps) {
+  return (
+    <div className="hidden flex-col gap-8 lg:flex">
+      {results}
+      {category}
+    </div>
+  )
+}
+
```
barbaraperic, line 76: Created a new component that has a smaller gap than ResultsAndCategory. See below how it looks with gap-10. Alternatively we could pass a prop, but thought this looks cleaner in the code base.

Screenshot 2024-08-13 at 17 34 34

CharlyMartin:
## Visual tests
**The active background is full-width on Figma Vs min-content on the website.**

Figma
CleanShot 2024-08-14 at 11 49 10@2x

Website
CleanShot 2024-08-14 at 11 50 32@2x

src/app/_components/CategorySelect.tsx
```diff
@@ -35,36 +37,65 @@ export function CategorySelect({
...
-    </ul>
+    </CategoryWrapper>
+  )
+}
+
+export function CategoryListItem({
+  name,
+  count,
+  isSelected,
+  handleClick,
```
CharlyMartin, line 78: FYI, I get this error when I hover over `handleClick`. I'm not sure if it's relevant to us, just sharing.

_Props must be serializable for components in the "use client" entry file, "handleClick" is invalid.ts(71007)_

src/app/events/page.tsx
```diff
@@ -123,8 +129,13 @@ export default function Events({ searchParams }: Props) {
 
       <PageSection kicker="Events" title="Network Events">
         <FilterContainer>
-          <FilterContainer.ResultsAndCategory
-            results={<ResultsAndReset results={categorizedResults.length} />}
+          <FilterContainer.ResultsAndCategoryCompact
+            results={
+              <CountAndReset
+                counts={filteredCategoryCounts}
+                isSelected={isResetCategoryApplied(searchParams)}
```
CharlyMartin, line 136: The name of the function `isResetCategoryApplied` is a bit misleading. What about `areFiltersReset`?

src/app/_utils/getCountForResetCategory.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export function getCountForResetCategory(counts: CategoryCounts) {
+  const totalSum = Object.values(counts).reduce((sum, value) => sum + value, 0)
+
+  if (totalSum !== 0) {
+    return {
+      [DEFAULT_CATEGORY]: totalSum,
+    }
+  }
+}
```
CharlyMartin, line 13: I don't understand this function. What is its purpose?

src/app/_constants/sortConstants.ts
```diff
@@ -1,10 +1,46 @@
...
+  },
+  alphabeticalSort: {
+    default: 'a-z' as const,
+    values: [
+      { id: 'a-z', name: 'Alphabetical (A-Z)' },
+      { id: 'z-a', name: 'Alphabetical (Z-A)' },
+    ] as SortSetting[],
+  },
+}
+
```
CharlyMartin, line 19: I feel like this could be more strictly typed, the `as SortSetting[]` cast is a workaround.

Maybe using a global `as const` would be more accurate? I haven't tested it though, it's quite hard at the review stage :)

```ts
const sortOptions = {
  cronologicalSort: {
    default: 'newest',
    values: [
      { id: 'newest', name: 'Newest' },
      { id: 'oldest', name: 'Oldest' },
    ],
  },
  alphabeticalSort: {
    default: 'a-z',
    values: [
      { id: 'a-z', name: 'Alphabetical (A-Z)' },
      { id: 'z-a', name: 'Alphabetical (Z-A)' },
    ],
  },
} as const
```

src/app/_components/CategoryListbox.tsx
```diff
@@ -27,6 +30,8 @@ export function CategoryListbox({
   categoryCounts,
   onCategoryOptionChange,
 }: CategoryListboxProps) {
+  const totalCount = getCountForResetCategory(categoryCounts || {})
```
CharlyMartin, line 33: Why not make the type optional instead of passing an empty object?

CharlyMartin (changes requested):

src/app/_utils/getCountForResetCategory.ts
barbaraperic: The purpose is to get a count for `DEFAULT_CATEGORY`, which in our case is "All", so we pass `categoryCounts` as a prop , for example `{ "use_case": 12, "interview": 33 }` and as a result we get an object, for example `{"All": 79 }`

barbaraperic:
> Hi @barbaraperic!
> 
> I've had an initial look at the PR and made a few code comments. I think we can improve the types so that they work better for us, but it's quite hard to do at the review stage. I'm happy to have a second look or pair program to improve it.
> 
> I've also added some feedback about the visual bugs I've found.
> 
> Other than that, it works as expected :)
> 
> ## Visual Tests
> 1. The active background is full-width on Figma Vs min-content on the website.
> 
> Figma	Website
> CleanShot 2024-08-14 at 11 54 29@2x	CleanShot 2024-08-14 at 11 50 32@2x
> 2. I know it works as expected, but clicking `All` to reset filters is counterintuitive, I find.
> 3. The images of the first 2 events have different padding than the others. I'm not sure if it's a coincidence or a bug.
> 
> CleanShot 2024-08-14 at 11 59 45@2x
> 
> 4. The ecosystem project cards have a weird height
> 
> CleanShot 2024-08-14 at 12 04 00@2x

Thanks @CharlyMartin!

1. [Discussed with Filipa](https://filecoinfoundation.slack.com/archives/C06MTEQ3SP6/p1724066746802839), we can leave the hover/focus state content-width
2. Any suggestions on alternatives?
3. and 4. are current bugs we have - @mirhamasala something to include in the upcoming sprint?

src/app/_utils/geCountForResetCategory.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export function getCountForResetCategory(counts: CategoryCounts) {
+  const totalSum = Object.values(counts).reduce((sum, value) => sum + value, 0)
+
+  if (totalSum !== 0) {
+    return {
+      [DEFAULT_CATEGORY]: totalSum,
+    }
+  }
+}
```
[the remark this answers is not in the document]
barbaraperic: Yes, good catch!

src/app/_components/CategorySelect.tsx
barbaraperic: Hm I couldn't recreate, what were the exact steps? Or could you record a video?

src/app/_constants/sortConstants.ts
barbaraperic: Ah yes, we can make it immutable, good one. 👍

barbaraperic: Is there a benefit to make it more strict? We use the values to get SortSettings, the `id` is types with type `SortOption`, not sure if the `name` needs to be strictly typed?

src/app/_components/CategorySelect.tsx
barbaraperic: Hmm not sure, I don't see it. You probably have some vs code extension I don't? `handleClick` is passed from a client component so not sure what could be an issue 🤔

src/app/_components/CategorySelect.tsx
```diff
@@ -35,36 +37,65 @@ export function CategorySelect({
...
+            handleClick={() => onCategoryOptionChange(option.id)}
+          />
         )
       })}
-    </ul>
+    </CategoryWrapper>
+  )
+}
+
+export function CategoryListItem({
```
CharlyMartin, line 74: If we have 3 components in the same file, they should be namespaced like we do in `Card` or `FilterContainer`, for readability :)

CharlyMartin (changes requested):

src/app/_components/FilterContainer.tsx
CharlyMartin: Fine by me :)

mirhamasala (changes requested):
Please hold - Still reviewing!


- Rename `getCountForResetForResetCategory` to `getTotalCategoryCount` - Should this live in `categoryUtils`?
- Rename file `countAndReset` to `CategoryResetButton`
- By making the `countAndReset` use `CategorySelect` components - we don't get the right HTML - we now have double lists 
- I have a refactor ready for `sortEntriesByDate` but won't be able to implement it until we get Orbit events functional again


@barbaraperic 

> and 4. are current bugs we have - @mirhamasala something to include in the upcoming sprint?

Yes - Please add tickets. 🙏🏼

src/app/_components/CategorySelect.tsx
mirhamasala: Done.

src/app/_utils/sortEntriesAlphabetically.ts
```diff
@@ -1,19 +1,24 @@
...
+  entries,
+  sortId,
+}: SortEntriesParams<Entry>): Entry[] {
+  const sortedEntries = [...entries].sort((a, b) => {
+    const nameA = (a.title || '').toLowerCase()
+    const nameB = (b.title || '').toLowerCase()
+
+    return sortId === 'z-a'
+      ? nameB.localeCompare(nameA)
+      : nameA.localeCompare(nameB)
```
mirhamasala, line 20: Should this be `title` instead of `name`?

CharlyMartin (approved):
@barbaraperic I renamed and moved a few things around for readability. Most of it was unrelated to your code changes, I just found it misleading.

1. I renamed `option` to `id`, and `setting` to `option`, to be more consistent with the rest of the code and the HTML specs - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option
2. I renamed `CategorySelect` to `CategorySidebar`. A select element implies a dropdown list, but in our case, it's a vertical list of buttons - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select
3. I removed the redundant prop names, `selectedCategory` to `selected`, `categoryOptions` to `options`, etc - https://github.com/ryanmcdermott/clean-code-javascript?tab=readme-ov-file#dont-add-unneeded-context
4. I moved `DEFAULT_CATEGORY` to the constants file


@mirhamasala About this:
> Rename getCountForResetForResetCategory to getTotalCategoryCount - Should this live in categoryUtils?

I did and it was causing a [build error](https://vercel.com/filecoin-foundations-projects/filecoin-foundation-site/Dk6mSc7fRrd7kdLhF39ohpQUTgDH?filter=errors), so I reverted it

CleanShot 2024-08-23 at 10 23 40

mirhamasala:
@barbaraperic @CharlyMartin - That was a big one! It's looking awesome. Thanks, both. ❤️

PR #513, opened 2024-08-26
UXIT-1323 - Custom Eslint rule to enforce generic Array syntax 

barbaraperic:
## 📝 Description

Added `"@typescript-eslint/array-type": ["error", { "default": "generic" }],` to enforce the preference for using Array<T> over T[].

- **Type:** Refactor

## 🛠️ Key Changes

- refactors in the code where `T[]` is replaced with `Array<T>`

## 🧪 How to Test

- **Steps to Test:**
  1. try adding a variable with a type `const example: string[] = []`

- **Expected Results:** You should get a eslint error that says `Use Array<string> instead of string[]`
- **Additional Notes:** You might need to restart the code editor, because rules are sometimes cached.

## 📸 Screenshots

https://github.com/user-attachments/assets/bc61e612-4c94-4a60-af95-bbf4e248d596


package.json
```diff
@@ -63,6 +63,8 @@
     "@types/node": "^22",
     "@types/react": "^18",
     "@types/react-dom": "^18",
+    "@typescript-eslint/eslint-plugin": "^7.2.0",
+    "@typescript-eslint/parser": "^7.2.0",
```
mirhamasala, line 67: Do we need these?

.eslintrc.json
```diff
@@ -4,6 +4,8 @@
     "plugin:import/errors",
     "plugin:import/warnings"
   ],
+  "plugins": ["@typescript-eslint", "custom-rules"],
```
mirhamasala, line 7: Do we need `"@typescript-eslint"`?

mirhamasala (commented):
@barbaraperic - This is very cool. Do you know why the default rule isn't working? Did you check GitHub issues or something like that?

barbaraperic:
> @barbaraperic - This is very cool. Do you know why the default rule isn't working? Did you check GitHub issues or something like that?

I tried again, and it's working! think it was the cache... I saw resources on custom rule for that specific case so gave it a go, but looks like typescript handles it after all. It was fun to build tho lol.

package.json
mirhamasala: @barbaraperic ^^

mirhamasala:
@barbaraperic >It was fun to build tho lol.

Yeah, super cool.

Just wanted to confirm if we really need this? It's not part of Next.js? [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/513#discussion_r1730973419)

mirhamasala (approved):
@barbaraperic One step closer to automating all the things! ❤️

PR #543, opened 2024-08-28
[UXIT-1480] Bug Bounty Page Leaderboard

CharlyMartin:
## 📝 Description

This PR adds a leaderboard section to the bug bounty page, available at `/security/bug-bounty`

## 🛠️ Key Changes

- Adds a `Leaderboard` component
- Installs [Tanstack Table](https://tanstack.com/table/latest), previously React Table, to manage the table state.
- Adds a new declaration file `react-table.d.ts`

## 🧪 How to Test

- Visit `/security/bug-bounty` and make sure the table looks right [[Figma](https://www.figma.com/design/0pBfQs5tKI6bx5ypPIfVRA/FIL.org-V4-Prototypes?node-id=5773-88176&t=iqcBMnjIBjJkSyaB-4)]
- On xs screens, the table should be scrollable horizontally

## 📸 Screenshots

CleanShot 2024-08-30 at 10 05 04

## 🔖 Resources

- https://tanstack.com/table/latest/docs/framework/react/examples/basic
- https://tanstack.com/table/v8/docs/api/core/column-def#meta

## Note on Tanstack Table

I understand that this feature could have been done without Tanstack Table, as this table is fully static.

I usually Implement Tanstack Table from the start, as, in my experience, I reach for it as soon as I need to add interactivity to a table. It also forces us to separate the data layer from the UI, which is always a good idea.

That being said, it can be removed if you think it's unnecessary here.


src/app/_types/react-table.d.ts
```diff
@@ -0,0 +1,8 @@
+import '@tanstack/react-table' //or vue, svelte, solid, qwik, etc.
```
mirhamasala, line 1:
```suggestion
import '@tanstack/react-table'
```

mirhamasala (approved):
Absolutely amazing, @CharlyMartin! Super clean! And it looks awesome. ✨

PR #598, opened 2024-09-02
UXIT-1329 - Avatar component

barbaraperic:
## 📝 Description

This PR includes a new `Avatar` component.

- **Type:** New feature

## 🛠️ Key Changes

- `Avatar` component
- style changes to `digestArticleHeader`
- modification to `DigestArticleFrontMatterSchema` - removed `alt` from image as we don't necessarily need it.

## 🧪 How to Test

- **Steps to Test:**
  1. check UI on `digest` and `digest/[slug]`
  2. test with longer names
  3. test with adding an image

## 📸 Screenshots
Screenshot 2024-09-02 at 10 49 54
Screenshot 2024-09-02 at 10 49 30

### DIFF
--- a/src/app/_components/Avatar.tsx
+++ b/src/app/_components/Avatar.tsx
@@ -1,0 +1,65 @@
+import React from 'react'
+
+import Image from 'next/image'
+
+import type { ImageProps } from '@/types/sharedProps/imageType'
+
+type ImagePropsWithoutAlt = Omit<ImageProps, 'alt'>
+
+export type Author = {
+  firstName: string
+  lastName: string
+  image?: ImagePropsWithoutAlt
+}
+
+export function Avatar({ authors }: { authors: Array<Author> }) {
+  const imageSize = 32
+
+  if (!authors || authors.length === 0) {
+    throw new Error('Avatar component requires at least one author')
+  }
+
+  const authorsNames = authors
+    .map((author) => `${author.firstName} ${author.lastName}`)
+    .join(' & ')
+
+  const renderContent = ({ firstName, lastName, image }: Author) => {
+    if (image?.src) {
+      return (
+        <Image
+          className="inline-block rounded-full object-cover ring-1 ring-brand-100"
+          src={image.src}
+          alt={`Photo of ${firstName} ${lastName}`}
+          width={imageSize}
+          height={imageSize}
+        />
+      )
+    } else {
+      const initials = (firstName[0] + lastName[0]).toUpperCase()
+      return (
+        <div
+          style={{ width: imageSize, height: imageSize }}
+          className="flex items-center justify-center rounded-full bg-brand-700 ring-1 ring-brand-100"
+          aria-label={`Initials of ${firstName} ${lastName}`}
+        >
+          <span className="text-sm font-medium leading-none text-brand-300">
+            {initials}
+          </span>
+        </div>
+      )
+    }
+  }
+
+  return (
+    <div className="flex items-center space-x-2 text-brand-100">
+      <div className="flex -space-x-1">
+        {authors.map((author) => (
+          <React.Fragment key={`${author.firstName}-${author.lastName}`}>
+            {renderContent(author)}
+          </React.Fragment>
+        ))}
+      </div>
+      <span className="text-sm">{authorsNames}</span>
+    </div>
+  )
+}
--- a/src/app/_components/Card.tsx
+++ b/src/app/_components/Card.tsx
@@ -2,6 +2,7 @@ import { ArrowUpRight } from '@phosphor-icons/react/dist/ssr'
 import { clsx } from 'clsx'
 import theme from 'tailwindcss/defaultTheme'
 
+import { type Author, Avatar } from '@/components/Avatar'
 import { CustomLink } from '@/components/CustomLink'
 import { DynamicImage, type DynamicImageProps } from '@/components/DynamicImage'
 import { Heading } from '@/components/Heading'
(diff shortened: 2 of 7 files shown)


src/app/_components/Avatar.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  if (!authors || authors.length === 0) {
+    throw new Error('Avatar component requires at least one author')
+  }
+
+  const authorsNames = authors
+    .map((author) => `${author.firstName} ${author.lastName}`)
+    .join(' & ')
+
+  const renderContent = ({ firstName, lastName, image }: Author) => {
```
mirhamasala, line 26: I'll let @CharlyMartin review this properly first but I wanted to drop in an early review. Let's refactor this using Clean Code JS principles.

Also - Can we start making the `Card` component more modular? We'll be doing a refactor soon - however, we can already have something like `Card.Avatar`.

src/app/_components/Avatar.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+import Image from 'next/image'
+
+import { type DigestArticleData } from '@/types/digestTypes'
+import type { ImageProps } from '@/types/sharedProps/imageType'
+
+type ImagePropsWithoutAlt = Omit<ImageProps, 'alt'>
+
+export function Avatar({ authors }: { authors: DigestArticleData['authors'] }) {
+  const imageSize = 32
```
CharlyMartin, line 11:
```suggestion
type AvatarProps = {
  authors: DigestArticleData['authors']
}

export function Avatar({ authors }: AvatarProps) {
```

src/app/_components/Avatar.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  if (!authors || authors.length === 0) {
```
CharlyMartin, line 13: Nit: `authors` is mandatory so checking for its existence is unnecessary?

```suggestion
  if (authors.length === 0) {
```

src/app/_components/Avatar.tsx
```diff
@@ -0,0 +1,67 @@
...
+    throw new Error('Avatar component requires at least one author')
```
CharlyMartin, line 14: Why are we're throwing an error here? Would rendering nothing be a bad thing?

Just curious, sometimes throwing an error is the right thing to do :)

src/app/_components/Avatar.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  function renderAuthorContent({
+    firstName,
+    lastName,
+    image,
+  }: DigestArticleData['authors'][number]) {
+    return image?.src
+      ? renderImage(firstName, lastName, image)
+      : renderInitials(firstName, lastName)
+  }
```
CharlyMartin, line 56: Just curious as to why these are functions and not components?

src/app/_components/AvatarGroup.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import { formatAuthors } from '@/utils/formatAuthors'
+
+import { Avatar } from '@/components/Avatar'
+
+export function AvatarGroup({
+  authors,
+}: {
+  authors: DigestArticleData['authors']
+}) {
```
CharlyMartin, line 11: type AvatarGroupProps = {
  authors: DigestArticleData['authors']
}

export function AvatarGroup({ authors }: AvatarGroupProps) {

src/app/_components/Card.tsx
```diff
@@ -76,6 +79,18 @@ function Link({ href, ariaLabel, icon, text, left = 'left-4' }: LinkProps) {
   )
 }
 
+function CardAvatarGroup({
+  authors,
+}: {
+  authors: DigestArticleData['authors']
+}) {
```
CharlyMartin, line 86: Same comment, define type props outside the component :)

src/app/_schemas/digestArticleFrontMatterSchema.ts
```diff
@@ -3,6 +3,8 @@ import { z } from 'zod'
 import { DynamicBaseDataSchema } from '@/schemas/dynamicDataBaseSchema'
 import { ImagePropsSchema } from '@/schemas/imagePropsSchema'
 
+const ImagePropsWithoutAltSchema = ImagePropsSchema.omit({ alt: true })
```
CharlyMartin, line 6: I didn't know about `omit()`, cool stuff :)

CharlyMartin (changes requested):

CharlyMartin:
@barbaraperic - one more general comment about the avatar components.

I feel like `Avatar` is doing too much and `AvatarGroup` not enough, meaning:
- `Avatar` should take in one object, instead of an array, and only render one element, either an image or initials.
- `AvatarGroup` should be responsible for rendering the list of avatars, along with their full names.

It would make more semantic sense this way, and probably simplify the `Avatar` component by getting rid of some of these helper functions.

What do you think? Is it even possible?

Let me know, thanks 🙂

src/app/_components/Avatar.tsx
barbaraperic: I find it that they're specific to the Avatar logic and aren't meant to be reused elsewhere.

barbaraperic:
> I feel like `Avatar` is doing too much and `AvatarGroup` not enough, meaning:
> 
> * `Avatar` should take in one object, instead of an array, and only render one element, either an image or initials.
> * `AvatarGroup` should be responsible for rendering the list of avatars, along with their full names.
> 
> It would make more semantic sense this way, and probably simplify the `Avatar` component by getting rid of some of these helper functions.

I agree! I refactored, let me know what you think.

CharlyMartin (approved):

mirhamasala:
@barbaraperic @CharlyMartin - This is glorious! 👑 I pushed a CR to start modularizing the `Card` component [4703fc8](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/598/commits/4703fc8ccaaef7ff727845ef2e64ba6a8d637dfe) - If nothing is alarming there, we can get this merged. Thank you! ❤️

mirhamasala (approved):
❤️

PR #599, opened 2024-09-02
[UXIT-1471] Optimise Percy snapshots [skip percy]

barbaraperic:
## 📝 Description

This PR includes a few options to optimise executing Percy Snapshots - see Key Changes. ❗️See part with including [husky](https://github.com/typicode/husky#readme).

- **Type:** Refactor

## 🛠️ Key Changes

- updates on github action - don't run tests if PR is in draft
- updates on github action - if you include "`[skip percy]`" in your PR title, only Cypress tests will run (without Percy). Not an ideal solution, adds an extra load to think about. I was looking into [Inquirer](https://github.com/SBoudrias/Inquirer.js/blob/main/packages/core/README.md) - we could use that library to implement a terminal prompt that could ask "Was there a visual change" when you run "git commit ...", and if the answer is positive that would add a condition to github actions to include Percy (otherwise it wouldn't include it in Cypress tests). I was playing around tho, but couldn't make it work. Could give it another go if you all think it would be a better solution.
- I included [husky](https://github.com/typicode/husky#readme) -  it's a tool that basically runs commands pre-commit. I thought it would be useful to run `npm run lint` before you push to the branch. Let me know if you want to keep it, or I can remove it. 
- only runs `percy` on deskop viewport - that reduces snapshots from 24 per commit to 12 per commit.

## 📸 Screenshots

Screenshot 2024-09-02 at 15 53 04

## 🔖 Resources

- [Be more strategic with our Percy snapshot quota](https://github.com/carbon-design-system/carbon/issues/14779)
- [StackOverflow Github Actions](https://stackoverflow.com/questions/60916931/github-actions-does-the-if-have-an-else)


CharlyMartin (commented):
>  don't run tests if PR is in draft
That's great!! It will for sure reduce our usage.

> if you include "[skip percy]" in your PR title, only Cypress tests will run (without Percy)
Not an ideal solution as you pointed out, I'm sure I will forget it 😅 Inquirer could be an improvement I think, but I don't know whether it's complicated to set.

The optimal solution would be that Percy only runs on pages that changed, but I don't know how to set that up. Maybe it will be possible when we setup NX or Turborepo.

> I included husky
I did a test commit locally and I don't think the linter ran. Did I miss something?


> only runs percy on deskop viewport
I feel like we might miss on important information by removing small screens from Percy?

Let me know what you think :)

mirhamasala:
@barbaraperic - Dropping in to say I agree with @CharlyMartin on:

> I feel like we might miss on important information by removing small screens from Percy?

and yes, let's wait until we setup our monorepo structure

>The optimal solution would be to only run Percy on pages that changed, but I don't know how do that. Maybe it will be possible when we setup NX or Turborepo.

🙏🏼

barbaraperic:
> > only runs percy on deskop viewport
> 
> I feel like we might miss on important information by removing small screens from Percy?

What if we keep just desktop screenshots for all browsers and keep both desktop and mobile screenshot for only 1 (most used) browser? Total screenshots per build would drop from 24 to 18 (saved 6). **Which is 25%.** 

Doesn't seem a lot, but, for example, if you look at the dashboard, in one day we used 384 screenshots, with 25% saving it would have been 288 (96 saved).

Thoughts? 

Screenshot 2024-09-04 at 13 17 18

barbaraperic:
>
> > I included husky
> 
> I did a test commit locally, and I don't think the linter ran. Did I miss something?
> 
> CleanShot 2024-09-04 at 10 35 25@2x

Hm not sure what could be the issue. You ran `npm install`?

.github/workflows/cypress-percy.yml
```diff
@@ -36,5 +37,25 @@ jobs:
...
+            COMMIT_MESSAGE="${{ github.event.pull_request.title }}"
+          else
+            COMMIT_MESSAGE="${{ github.event.head_commit.message }}"
+          fi
+
+          if [[ "$COMMIT_MESSAGE" == *"[skip percy]"* ]]; then
+            echo "skip_percy=true" >> $GITHUB_OUTPUT
+          else
+            echo "skip_percy=false" >> $GITHUB_OUTPUT
+          fi
```
mirhamasala, line 53:
```suggestion
      - name: Determine Test Type
        id: test-type
        run: |
          COMMIT_MESSAGE=""

          if [[ "${{ github.event_name }}" == "pull_request" ]]; then
            COMMIT_MESSAGE="${{ github.event.pull_request.title }}"
          else
            COMMIT_MESSAGE="${{ github.event.head_commit.message }}"
          fi

          SKIP_PERCY="false"
          if [[ "$COMMIT_MESSAGE" == *"[skip percy]"* ]]; then
            SKIP_PERCY="true"
          fi

          echo "skip_percy=$SKIP_PERCY" >> $GITHUB_OUTPUT_PATH
```

CharlyMartin (approved):
Looks good to me 👏

FYI, I did a test commit and the linter ran with this message:

CleanShot 2024-09-06 at 11 39 53@2x

mirhamasala (approved):
Yay! LGTM! ✨

barbaraperic:
> Looks good to me 👏
> 
> FYI, I did a test commit and the linter ran with this message:
> 
> CleanShot 2024-09-06 at 11 39 53@2x

Thanks Charly! I'll open a ticket to look into it 👍

PR #601, opened 2024-09-02
[UXIT-1487] Security Maturity Model Table of Contents

CharlyMartin:
## 📝 Description

This PR adds a table of contents on the `/security/maturity-model` page.

The `TableOfContents` behaves as follows:

1. It sticks to the top of the page on large screens.
2. It smoothly scrolls to sections when titles are clicked.
4. It automatically updates the URL hash based on the scroll position.

## 🛠️ Key Changes

- Adds a `TableOfContents` component
- Adds a prop to `PageSection` to allow full-width layout
- Adds a `useSectionHashObserver` hook to keep track of the current section hash in the URL
- Adds a `Article` component where the `useIntersectionObserver` is implemented

## 🧪 How to Test

Visit `/security/maturity-model` and make sure the table of contents behaves as described above.

## 📸 Screenshots

Video: https://share.cleanshot.com/YhH40Y4d

CleanShot 2024-09-04 at 15 39 21

## 🔖 Resources

- https://usehooks-ts.com/react-hook/use-intersection-observer


src/app/_components/PageSection.tsx
```diff
@@ -33,16 +34,17 @@ export function PageSection({
   image,
   children,
   cta,
+  fullWidth,
```
mirhamasala, line 37: I would find it more intuitive if using full width is the default - And we have a prop, like, `shouldConstrainTextWidth` set to default `true`.

In that case, we wouldn't need the double negation either.

```
      <div
        className={clsx('mb-6 mt-4', {
          'grid grid-cols-1 gap-6 lg:grid-cols-2': image,
          'max-w-readable': shouldConstrainTextWidth,
        })}
      >
```

src/app/security/maturity-model/components/Article.tsx
```diff
@@ -0,0 +1,50 @@
...
+      >
+        <h3 id={slug} className="flex items-center gap-2">
+          {title}
+          <span className="invisible group-hover:visible">
+            <Icon component={LinkSimple} size={18} />
+          </span>
+        </h3>
+      </Link>
+      {children}
+    </article>
```
mirhamasala, line 48: We should avoid wrapping heading elements in links, as this compromises their semantic purpose. Headings are meant to structure content and aid navigation for both screen readers and search engines.

n addition, wrapping the entire heading in a link can lead to keyboard accessibility issues, where these links might be skipped over or incorrectly focused – as is currently the case.

What about something like this?

```suggestion
    <article ref={ref}>
      <h3 id={slug}>
        <Link
          href={sectionHash as Route}
          className="group inline-flex items-center gap-2 text-brand-100 hover:no-underline"
        >
          {title}
          <span className="invisible group-hover:visible">
            <Icon component={LinkSimple} size={18} />
          </span>
        </Link>
      </h3>
      {children}
    </article>
```

src/app/security/maturity-model/data/scrollToSection.ts
mirhamasala, line 1: Should this be in `utils` instead of `data`?

src/app/security/maturity-model/page.tsx
```diff
@@ -18,6 +20,14 @@ import { CoreFunctions } from './components/CoreFunctions'
 import { applicationAndUseData } from './data/applicationAndUseData'
 import { generateStructuredData } from './utils/generateStructuredData'
 
+const DynamicTableOfContent = dynamic(
```
mirhamasala, line 23:
```suggestion
const DynamicTableOfContents = dynamic(
```

src/app/security/maturity-model/page.tsx
```diff
@@ -63,10 +73,18 @@ export default function MaturityModel() {
...
         kicker="Building Blocks of Security"
         title="Explore the Core Functions"
       >
-        <CoreFunctions />
+        <div className="flex flex-col gap-8 lg:relative lg:flex-row lg:items-start lg:gap-12">
+          <div className="grow">
+            <CoreFunctions />
+          </div>
+          <div className="order-first lg:sticky lg:top-12 lg:order-last lg:w-72">
+            <DynamicTableOfContent />
```
mirhamasala, line 85:
```suggestion
            <DynamicTableOfContents />
```

src/app/security/maturity-model/data/coreFunctionsData.ts
```diff
@@ -37,13 +37,7 @@ import {
...
 } from '../content/software-security.md'
 
-type CoreFunction = {
-  order: SecurityMaturityModelAttributes['order']
-  title: SecurityMaturityModelAttributes['title']
-  Content: React.ComponentType
-}
-
-export const coreFunctionsData: Array<CoreFunction> = [
+const coreFunctionsBaseData = [
```
mirhamasala, line 40: I know this is working but why - How is TS inferring this?

src/app/security/maturity-model/utils/useSectionHashObserver.ts
```diff
@@ -0,0 +1,84 @@
...
+    const newHash = getHashFromEvent(event)
+    setHash(newHash)
+  }
+
+  useEffect(() => {
+    if (windowIsDefined()) {
+      window.addEventListener('hashchange', onHashChange)
+      return () => window.removeEventListener('hashchange', onHashChange)
+    }
+  }, [pathname])
```
mirhamasala, line 51: Is adding the listener only once not enough?

```suggestion
  useEffect(() => {
    if (windowIsDefined()) {
      window.addEventListener('hashchange', onHashChange)
      return () => window.removeEventListener('hashchange', onHashChange)
    }
  }, [])
```

src/app/security/maturity-model/components/TableOfContents.tsx
```diff
@@ -0,0 +1,55 @@
...
+        Table of Contents
+      </p>
+
+      <ul>
+        {coreFunctionsData.map(({ slug, title }, index) => {
+          const sectionHash: SectionHash = `${HASH_SIGN}${slug}`
+          const isCurrentSection = currentHash === sectionHash
+
+          return (
+            <li key={index}>
```
mirhamasala, line 37: Same [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/601/files#r1746064222)

src/app/security/maturity-model/components/TableOfContents.tsx
```diff
@@ -0,0 +1,55 @@
...
+  })
+
+  return (
+    <nav ref={ref}>
+      <p className="mb-4 ml-6 text-sm font-semibold uppercase text-brand-200 lg:mb-6">
+        Table of Contents
+      </p>
+
+      <ul>
+        {coreFunctionsData.map(({ slug, title }, index) => {
```
barbaraperic, line 32: We should maybe reconsider markup https://css-tricks.com/a-perfect-table-of-contents-with-html-css/

barbaraperic:
When you click on the section, scroll down and click on the section again, the "active" styles don't get applied

[CleanShot 2024-09-04 at 20.24.03.mp4.zip](https://github.com/user-attachments/files/16905258/CleanShot.2024-09-04.at.20.24.03.mp4.zip)

mirhamasala:
Some libraries for inspiration:

- https://streamich.github.io/react-use/?path=/story/sensors-usehash--docs
- https://react-intersection-observer.vercel.app/?path=/docs/intro--docs
- https://github.com/fisshy/react-scroll/blob/master/README.md

src/app/_components/PageSection.tsx
CharlyMartin: After a second look, we don't need any any of that!

`max-w-readable` only applies to the `Heading` and `DescriptionText, not `children`, which is outside of this container.

So, it's all good. Thanks for flagging :)

src/app/security/maturity-model/components/Article.tsx
CharlyMartin: I wasn't aware of that, thanks for sharing :)

src/app/security/maturity-model/components/TableOfContents.tsx
mirhamasala: Good idea! @CharlyMartin

CharlyMartin:
Thank you, @barbaraperic and @mirhamasala, for your review; very helpful!

About the libraries. I tried using `useHash` from `react-use` but because it calls `window.location.hash` internally, the navigation jumps instead of scrolling. But good call nonetheless, it could have been useful.

https://github.com/streamich/react-use/blob/master/src/useHash.ts#L27

src/app/_components/PageSection.tsx
```diff
@@ -39,10 +39,10 @@ export function PageSection({
...
 
       <div
-        className={clsx(
-          'mb-6 mt-4',
-          image ? 'grid grid-cols-1 gap-6 lg:grid-cols-2' : 'max-w-readable',
-        )}
+        className={clsx('mb-6 mt-4', {
+          'grid grid-cols-1 gap-6 lg:grid-cols-2': image,
+          'max-w-readable': !image,
+        })}
```
barbaraperic, line 45:
```suggestion
        className={clsx('mt-4', {
          'grid grid-cols-1 gap-6 lg:grid-cols-2': image,
          'mb-6 max-w-readable': !image,
        })}
```

I think we need to include `mb-6 `only when there's not an image. Otherwise we include this extra margin with the 64px gap between sections.
Example on the `security` page:
Screenshot 2024-09-12 at 09 38 04

src/app/security/maturity-model/components/Article.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import { LinkSimple } from '@phosphor-icons/react/dist/ssr'
+import type { Route } from 'next'
+import { useIntersectionObserver } from 'usehooks-ts'
+
+import { Icon } from '@/components/Icon'
+
+import { useUrlHash, HASH_SIGN } from '../utils/useUrlHash'
+
+type LinkTitleProps = {
```
barbaraperic, line 13: Should we name this `ArticleProps`?

barbaraperic:
Wondering if we should handle a case where we have a really long title? 🧐

Screenshot 2024-09-12 at 10 07 20

src/app/security/maturity-model/components/TableOfContents.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { useUrlHash, HASH_SIGN, type SectionHash } from '../utils/useUrlHash'
+
+export function TableOfContents() {
+  const { currentHash } = useUrlHash()
+
+  return (
+    <table role="navigation" aria-label="Table of Contents">
+      <p className="mb-4 ml-6 text-sm font-semibold uppercase text-brand-200 lg:mb-6">
+        Table of Contents
+      </p>
```
barbaraperic, line 16: Screenshot 2024-09-12 at 10 13 02

src/app/security/maturity-model/utils/useUrlHash.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export function useUrlHash() {
+  const pathname = usePathname()
+  const router = useRouter()
+
+  const [hash, setHash] = useState(getHashFromWindow)
+
+  function updateHash(sectionId: string) {
+    const sectionHash: SectionHash = `${HASH_SIGN}${sectionId}`
+    const pathnameWithNewSectionHash = pathname + sectionHash
```
barbaraperic, line 22: Maybe we can use template literals here so it's consistent with line 21?
```suggestion
    const pathnameWithNewSectionHash = `${pathname}${sectionHash}`
```
We can also move that after the `sectionHash === hash` condition as we don't need it before? A minor optimisation, but also works with having consts in one place.

barbaraperic:
Should we use scroll to view when the user clicks on the section? For example, [MDN has it implemented](https://developer.mozilla.org/en-US/docs/Web/API/Window/caches#examples)
 
https://github.com/user-attachments/assets/3c9c85ad-ceef-4883-a139-3f5f8f1ff274

src/app/security/maturity-model/components/TableOfContents.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+      <ol>
+        {coreFunctionsData.map(({ slug, title }) => {
+          const sectionHash: SectionHash = `${HASH_SIGN}${slug}`
+          const isCurrentSection = currentHash === sectionHash
+
+          return (
+            <li key={slug}>
```
barbaraperic, line 24: Think the hover effect is missing https://www.figma.com/design/0pBfQs5tKI6bx5ypPIfVRA/FIL.org-V4-Prototypes?node-id=5961-74272&t=4CbuL2Y0K1wJdP3C-4

src/app/security/maturity-model/components/Article.tsx
CharlyMartin: Yes!

src/app/security/maturity-model/components/TableOfContents.tsx
CharlyMartin: Good find! I was trying to use more explicit markup, as per your article recommendation, but it didn't work. I'll stick to `nav` for now.

src/app/security/maturity-model/utils/useUrlHash.ts
CharlyMartin: Love it!

src/app/security/maturity-model/components/TableOfContents.tsx
CharlyMartin: Nice thanks!

PR #648, opened 2024-09-11
UXIT-1415 - Rework Image Logic

barbaraperic:
## 📝 Description

This PR refactors the code to use the native Next image directly, replacing the `StaticImage` and `DynamicImage` components.

- **Type:** Refactor

## 🛠️ Key Changes

- Deleted `StaticImage` and `DynamicImage` component and replaced with Next's `Image`

## 📌 To-Do Before Merging

- [x] Handle fallback cases

## 🧪 How to Test

- **Setup:** [Setup details]
- **Steps to Test:**
  1. Check images on each page
  2. Test fallback cases


src/app/_components/FeaturedBlogPosts.tsx
```diff
@@ -36,6 +36,8 @@ export function FeaturedBlogPosts() {
             category,
           })
```
CharlyMartin, line 38: The import of `graphicsData` can be deleted at the top of the file.

src/app/_components/FeaturedEcosystemProjects.tsx
```diff
@@ -24,35 +24,35 @@ export function FeaturedEcosystemProjects({
...
-              startSize: '100vw',
-              sm: '320px',
-              md: '440px',
-              lg: '280px',
-            }),
-          }}
-        />
-      ))}
+      {ecosystemProjects.map(({ slug, title, description, image }) => {
+        const { src } = image || {}
```
CharlyMartin, line 28: The import of graphicsData can be deleted at the top of the file.

CharlyMartin:
First off, this implementation is a readability improvement over the initial one, so good one! 🙂

I’ve left some comments, but my main question is about `ImageWithFallback`: **Should this component handle static imports?**

```ts
const isStaticImport = typeof src === 'object'
```

From what I understand, local images “never” fail to load because they’re imported at build time. If there’s an issue with the import, the build itself would fail.

Because of this, I’m tempted to suggest that `ImageWithFallback` should only handle cases where an image might fail to load - when it comes from a remote URL. This would let us rely on the native `Image` from `next/image` wherever we import local images and `ImageWithFallback` with remote ones. And it would simplify the implementation of `ImageWithFallback`.

This split also aligns with the (subtle) differences in handling local vs. remote images in Next.js. Typically, local images don’t need a wrapper container since they come with predefined `width` and `height`, computed by Next. Remote images, on the other hand, often need a wrapper as they are usually set to fill the parent.

I understand that this approach means they are two image components, which might not be what we want. I don't see it as a problem, but maybe it will be confusing in the long run, Not sure, let me know what you think :)

src/app/_components/ArticleHeader.tsx
```diff
@@ -20,17 +21,15 @@ export function ArticleHeader({ image, children }: ArticleHeaderProps) {
   return (
     <header className="space-y-6">
       <div className="space-y-6">{children}</div>
-
       <div className="relative aspect-video">
-        <DynamicImage
+        <ImageWithFallback
```
mirhamasala, line 25: Why do we need to name it explicitly `WithFallback`? I kinda expect a fallback out of the box?

src/app/_components/ArticleHeader.tsx
```diff
@@ -1,14 +1,15 @@
...
 
 import { buildImageSizeProp } from '@/utils/buildImageSizeProp'
 
-import { DynamicImage } from '@/components/DynamicImage'
 import { type HeadingProps, Heading } from '@/components/Heading'
+import { ImageWithFallback } from '@/components/ImageWithFallback'
 
 type ArticleHeaderProps = {
-  image?: ImageProps
+  image?: RemoteImageData | LocalImageData
```
mirhamasala, line 12: What's the reason that this is both types?

src/app/_components/ImageWithFallback.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { graphicsData } from '@/data/graphicsData'
+
+type ImageWithFallbackProps = ImageProps & {
+  fallbackStyle?: CSSProperties
+}
+
+const { src: fallbackSrc, alt: fallbackAlt } = graphicsData.imageFallback
+
+async function getImageSrc(src: string | StaticImport) {
+  const isStaticImport = typeof src === 'object'
```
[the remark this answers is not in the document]
mirhamasala: I would probably rename everything to use `static` or `dynamic` since those terms come straight from Next.js docs. Is there an example where this wouldn't work?

src/app/_components/FeaturedBlogPosts.tsx
```diff
@@ -50,10 +52,8 @@ export function FeaturedBlogPosts() {
               }}
               image={{
                 alt: '',
-                ...(image || {
-                  ...graphicsData.imageFallback,
-                }),
-                fallback: graphicsData.imageFallback,
+                src: src || '',
```
mirhamasala, line 55: I was hoping we wouldn't have to use the fallbacks to `''` here anymore..

src/app/_components/FeaturedEcosystemProjects.tsx
mirhamasala, line 28: Same comment[ >>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/648/files#r1782129015)

src/app/_components/FeaturedGrantGraduates.tsx
```diff
@@ -24,33 +22,33 @@ export function FeaturedGrantsGraduates({
...
-            sizes: buildImageSizeProp({
-              startSize: '100vw',
-              sm: '175px',
-              md: '250px',
-            }),
-          }}
-        />
-      ))}
+      {grantGraduates.map(({ title, description, slug, image }) => {
+        const { src } = image || {}
```
mirhamasala, line 26: Same comment [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/648/files#r1782129015)

mirhamasala:
@barbaraperic @CharlyMartin I’m still feeling a bit unclear about this implementation. It’s hard to distinguish how images are being used, specifically what parts are static and what parts are dynamic. My hope was that we could first deconstruct everything, use Next.js’ Image component consistently throughout, and then, if necessary, start refactoring from there ([as outlined in the spec](https://www.notion.so/filecoin/FF-Rework-Image-Logic-6956f5e23ee244b59cacbb6144f786a4?pvs=4)).

mirhamasala:
Closing in favor of #697 and #700.

PR #674, opened 2024-09-25
[UXIT-1541] Event Page - Include calendar option

mirhamasala:
## 📝 Description

- **Type:** New feature

## 🛠️ Key Changes

- ...

## 📸 Screenshots

### DIFF
--- a/public/admin/config.yml
+++ b/public/admin/config.yml
@@ -668,5 +668,10 @@ collections:
         widget: "string"
         required: false
         hint: "External link to the event page"
+      - name: "luma-calendar-link"
+        label: "Luma Calendar URL"
+        widget: "string"
+        required: false
+        hint: "Luma Calendar URL for the event. Include relevant tags when appropriate, e.g., https://lu.ma/filecoin?tag=fil%20bangkok"
       - *image_config
       - *meta_config
--- a/public/admin/config.yml
+++ b/public/admin/config.yml
@@ -672,6 +672,6 @@ collections:
         label: "Luma Calendar URL"
         widget: "string"
         required: false
-        hint: "Luma Calendar URL for the event. Include relevant tags when appropriate, e.g., https://lu.ma/filecoin?tag=fil%20bangkok"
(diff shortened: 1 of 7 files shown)


src/app/events/[slug]/page.tsx
```diff
@@ -59,3 +65,20 @@ export default function EventEntry({ params }: EventProps) {
     </>
   )
 }
+
+function buildCtaArray(
+  externalLink: Event['externalLink'],
```
Comment (barbaraperic) [inline src/app/events/[slug]/page.tsx:70]: `externalLink` should be also optional right?

Maybe we could extract the type?

```
type EventCtaProps = {
  externalLink?: Event['externalLink']
  lumaCalendarLink?: Event['lumaCalendarLink']
}
```

src/app/events/[slug]/page.tsx
```diff
@@ -59,3 +65,20 @@ export default function EventEntry({ params }: EventProps) {
     </>
   )
 }
+
+function buildCtaArray(
```
Comment (barbaraperic) [inline src/app/events/[slug]/page.tsx:69]: Should we move the function in a util file? In case we want to reuse it?

barbaraperic (approved):
Looks good, works well 👍  left a few tiny comments

src/app/events/[slug]/page.tsx
Comment (mirhamasala) [reply src/app/events/[slug]/page.tsx:69]: In a local `utils` folder?

Comment (mirhamasala) [reply src/app/events/[slug]/page.tsx:69]: Yes! Good idea. Thank you.

src/app/events/[slug]/page.tsx
Comment (mirhamasala) [reply src/app/events/[slug]/page.tsx:70]: Yes!

public/admin/config.yml
```diff
@@ -668,5 +668,10 @@ collections:
         widget: "string"
         required: false
         hint: "External link to the event page"
+      - name: "luma-calendar-link"
+        label: "Luma Calendar URL"
+        widget: "string"
+        required: false
+        hint: "Luma Calendar URL for the event. Include relevant tags when appropriate, e.g., https://lu.ma/filecoin?tag=fil%20bangkok"
```
barbaraperic, line 675: Just curious - why adding the tag? We want to have the title of the event as a tag? Do they edit the url manually?

PR #689, opened 2024-10-02
UXIT-1555 - Run linters on PRs

barbaraperic:
## 📝 Description

Modify `lint.yml` to run on **every** open pull request.

- **Type:** Refactor

## 🧪 How to Test

Can't test locally, so once it's merged, we can test with a PR coming from the CMS.

## 🔖 Resources

- [Github Actions](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: - '**'
+    branches: ["*"]
 
 jobs:
   lint:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
+    branches: - '**'
 
 jobs:
   lint:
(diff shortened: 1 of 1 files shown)


mirhamasala (changes requested):
This is a very broad approach, as it includes all branches, even those that may not require linting or testing, such as temporary, experimental, or draft branches. While it ensures that every pull request gets coverage, it might result in unnecessary workflow runs, potentially wasting resources and increasing build times for branches we may not care about.

What if we did something more targeted, like the example below? This would limit the workflow to `main` and any branches under `cms/*` or `ecosystem-submission/*`, ensuring we focus on the branches that matter most right now:

```
on:
  push:
    branches:
      - main
      - cms/*
      - ecosystem-submission/*
  pull_request:
    branches:
      - main
      - cms/*
      - ecosystem-submission/*
```
      
This way, we only trigger the workflow for relevant branches, avoiding excess runs while still ensuring coverage for critical parts of the project.

mirhamasala (approved):
❤️🙌🏼 Thank you!

PR #691, opened 2024-10-02
[UXIT-1509] Enable strict mode for ecosystem entries [skip percy]

CharlyMartin:
## 📝 Description

This PR updates how we handle Zod validations, starting with ecosystem projects. I prioritized these updates over digest articles because the comms team is now merging PR on their own, so this will be a necessary safety net.

I had to rework our Zod Validations in order to enable strict mode, which throws an error when unrecognized keys are found in the object.

## 🛠️ Key Changes

- Updated the `EcosystemProjectFrontMatter` schema - and the schemas it extended - to use `kebab-case` for property keys, aligning with the data format from the CMS.
- Since the rest of the codebase uses camelCase, I installed the `camelcase-keys` package ([npm](https://www.npmjs.com/package/camelcase-keys)) to convert the object keys to standard JavaScript format.

Some code duplication was required to maintain compatibility with the other Zod Schemas:
- `BaseDataSchemaKebabCase` is a temporary duplicate of `DynamicBaseDataSchema`
- `getMarkdownData` is a temporary duplicate of `getData`

## 🧪 How to Test

Add an extra field to any ecosystem entry, build the project, and ensure it fails as expected.


src/app/_schemas/ecosystemProject/FrontMatterSchema.ts
```diff
@@ -1,22 +1,25 @@
 import { z } from 'zod'
 
-import { DynamicBaseDataSchema } from '@/schemas/dynamicDataBaseSchema'
+import { BaseDataSchemaKebabCase } from '@/schemas/dynamicDataBaseSchema'
 
 import { CategorySchema, SubcategorySchema } from './CategorySchemas'
 
-export const EcosystemProjectFrontMatter = DynamicBaseDataSchema.extend({
+// email and full-name are required in the CMS but optional here because some entries still don't have them
```
mirhamasala, line 7: Let's change the data - If I remember correctly, Matt was putting in his own email address for the ones that don't have it.

src/app/_utils/getMarkdownData.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    throw error
+  }
+}
+
+type GetAllData<T> = {
+  directoryPath: string
+  zodParser: ZodType<T>['parse']
+}
```
mirhamasala, line 57: Let's keep moving type definitions to the top. 🙏🏼

package.json
```diff
@@ -30,6 +30,7 @@
     "@vercel/speed-insights": "^1.0.12",
     "airtable": "^0.12.2",
     "autoprefixer": "^10.4.20",
+    "camelcase-keys": "^9.1.3",
```
mirhamasala, line 33: So nice! 👌🏽

mirhamasala (commented):
@CharlyMartin - Super happy about this! Looking good! I would prefer if we make the `email` and `full-name` required and [address the data inconsistencies in this PR](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/691/files#r1789994613). Thank you.

src/app/_schemas/ecosystemProject/FrontMatterSchema.ts
CharlyMartin: Will do!

CharlyMartin:
@mirhamasala I added `name` and `full-name` to all Markdown entries where they were missing. It was the case for most of them as you can see in the file diff. I used Matt's full name and email as the default.

mirhamasala (commented):
@CharlyMartin - This is good to go!

PR #701, opened 2024-10-09
UXIT-1428 - Headless UI v2 component refactor 

barbaraperic:
## 📝 Description

The V2 Headless UI components no longer use the <Parent.Child> component pattern, and rely on a different props API.

- **Type:** Refactor

## 🧪 How to Test

Check the following and see if everything works as intended (there should be no changes):

- Project Form 
- Navigation (desktop and mobile)

## 🔖 Resources

- [Popover](https://headlessui.com/react/popover)
- [Dialog](https://headlessui.com/react/dialog)
- [ListBox](https://headlessui.com/react/listbox)


src/app/_components/ListboxButton.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
 
-import { Listbox } from '@headlessui/react'
+import { ListboxButton as HeadlessListboxButton } from '@headlessui/react'
```
mirhamasala, line 3: I would be specific with naming.

```suggestion
import { ListboxButton as HeadlessUIListboxButton } from '@headlessui/react'
```

src/app/_components/ListboxOption.tsx
```diff
@@ -1,6 +1,6 @@
 import { Fragment } from 'react'
 
-import { Listbox } from '@headlessui/react'
+import { ListboxOption as HeadlessListboxOption } from '@headlessui/react'
```
mirhamasala, line 3: CleanShot 2024-10-11 at 06 34 11@2x

mirhamasala (commented):
Looking good! 🙌🏼 Nit: I would rename everything to include the full name by adding `UI` so `HeadlessUIListboxButton`, etc.

I think we can also run `npm install @headlessui/react@latest` at this point.

Tagging @CharlyMartin for a second look. 🙏🏼

src/app/_components/ListboxOptions.tsx
```diff
@@ -13,14 +13,14 @@ export function ListboxOptions({
   const positionClass = position === 'right' ? 'right-6 md:right-auto' : ''
 
   return (
-    <Listbox.Options
+    <HeadlessUIListboxOptions
       aria-labelledby="listbox-button"
```
CharlyMartin, line 17: As for the listbox button, they probably manage the aria labels for us now.

src/app/_components/ListboxOption.tsx
```diff
@@ -31,12 +31,12 @@ function OptionContent({ option, counts }: ListboxOptionProps) {
 
 export function ListboxOption({ option, counts }: ListboxOptionProps) {
   return (
-    <Listbox.Option value={option.id} as={Fragment}>
-      {({ active, selected }) => (
+    <HeadlessUIListboxOption value={option.id} as={Fragment}>
+      {({ focus, selected }) => (
```
CharlyMartin, line 35: Instead of using render props, how about we use `data-selected` and `data-focus` pseudo classes?

https://headlessui.com/react/listbox#listbox-option

src/app/_components/ListboxButton.tsx
```diff
@@ -14,13 +14,13 @@ export function ListboxButton({
...
 }: ListboxButtonProps) {
   return (
-    <Listbox.Button
+    <HeadlessUIListboxButton
       aria-haspopup="listbox"
       aria-expanded={open}
       aria-label={ariaLabel}
-      className="focus:brand-outline inline-flex w-full items-center justify-between gap-2 rounded-lg border border-brand-300 p-3 text-brand-300 hover:border-current hover:text-brand-400 md:min-w-40"
+      className="inline-flex w-full items-center justify-between gap-2 rounded-lg border border-brand-300 p-3 text-brand-300 focus:brand-outline hover:border-current hover:text-brand-400 md:min-w-40"
     >
```
CharlyMartin, line 22: I'm not sure we still need all these props.

Looking at the type definition `ListboxButtonProps`, they seem to manage the aria attributes for us:
```ts
type ButtonPropsWeControl = 'aria-controls' | 'aria-expanded' | 'aria-haspopup' | 'aria-labelledby' | 'disabled';
```

Also, `ListboxButtonProps` could extend the type definition exported by `'@headlessui/react'`

https://headlessui.com/react/button

src/app/_components/SlideOver.tsx
```diff
@@ -10,9 +15,9 @@ type SlideOverProps = {
 
 export function SlideOver({ open, setOpen, children }: SlideOverProps) {
   return (
-    <Transition.Root show={open} as={Fragment}>
+    <Transition show={open} as={Fragment}>
       <Dialog className="relative z-10" onClose={setOpen}>
-        <Transition.Child
+        <TransitionChild
```
CharlyMartin, line 20: Do we still need Transition? Looking at the [docs](https://headlessui.com/react/dialog#adding-transitions), we can now use a transition `prop` and use `data-closed`, `data-enter` and `data-leave` with Tailwind classes.

I haven't tried it so maybe we still need it, just sharing :)

CharlyMartin (changes requested):
@barbaraperic I think we can remove most aria labels as they are managed by Headless UI now. Also, in some instances, we should import their type definition to be closer to reality.

Let me know what you think :) Thanks!

---

As a more general comment, we might want to find a way to use `FormListbox` here. The implementation of both components is very similar, but that's probably for another PR!

src/app/_components/SlideOver.tsx
barbaraperic: Hmm yea I remember we tried that, but we came back to this solution, don't remember the reason tho

CharlyMartin:
@mirhamasala I got rid of the `Transition` component in the `Popover` but couldn't manage to do the same for the `Dialog`.

I didn't want to spend too much time on this so I believe this is good to go for now!

CharlyMartin (approved):

mirhamasala (approved):
@barbaraperic @CharlyMartin - Love it! So much cleaner! ❤️ LGTM!

PR #704, opened 2024-10-09
[UXIT-1539] Events Page · Schedule

mirhamasala:
## 📝 Description

This pull request introduces a new feature to handle event schedules, including the addition of a schema for schedules, updates to the event data structure, and UI components to display the schedule on event pages. The most important changes include adding a new schedule schema, updating the event data structure, and creating new components for displaying the schedule.

- **Type:** New feature

## Note
For the sake of speed, we have not implemented the mobile dropdown at this stage, instead opting for a sticky horizontal scrollbar. With Filipa unavailable, we don’t yet have a solid solution for handling the x-scrollbar effectively. This is also a pending task for the Allocators table.

The mobile dropdown will also be required for the Security Maturity Model Table of Contents, and at that point, we can refactor both implementations.

### Schema and Data Structure Updates:
* Added a new `ScheduleSchema` to define the structure of event schedules, including days and events (`src/app/_schemas/event/ScheduleSchema.ts`).
* Updated the `EventFrontMatterSchema` to include the new `schedule` field (`src/app/_schemas/event/FrontMatterSchema.ts`).
* Modified the `convertMarkdownToEventData` function to handle the new `schedule` field (`src/app/_utils/convertMarkdownToEventData.ts`).

### UI Components:
* Created a `ScheduleSection` component to display the schedule section on event pages (`src/app/events/[slug]/components/ScheduleSection.tsx`). ([src/app/events/[slug]/components/ScheduleSection.tsxR1-R17](diffhunk://#diff-12077aeda983adab3b92d95a7156148c8e7a587376d353cf1c9b8bcc7fa4ea4aR1-R17))
* Added a `ScheduleTabs` component to display the schedule in a tabbed format (`src/app/events/[slug]/components/ScheduleTabs.tsx`). ([src/app/events/[slug]/components/ScheduleTabs.tsxR1-R65](diffhunk://#diff-dd2e0f521001f0ac1f13251d226a62dbfeb4fac97a834c5f24a84cf10bb83859R1-R65))

### Event Page Integration:
* Integrated the `ScheduleSection` component into the event page (`src/app/events/[slug]/page.tsx`). ([src/app/events/[slug]/page.tsxR18](diffhunk://#diff-815f182831c560a8fc9dc4acaab794fe338311e80251f17c976d8ec33931e8c1R18), [src/app/events/[slug]/page.tsxR51](diffhunk://#diff-815f182831c560a8fc9dc4acaab794fe338311e80251f17c976d8ec33931e8c1R51), [src/app/events/[slug]/page.tsxR76-R79](diffhunk://#diff-815f182831c560a8fc9dc4acaab794fe338311e80251f17c976d8ec33931e8c1R76-R79))
* Updated the event markdown file to include sample schedule data (`src/content/events/fil-bangkok-2024.md`).

### Utility Functions:
* Added utility functions for formatting dates and times (`src/app/events/[slug]/utils/dateUtils.ts`). ([src/app/events/[slug]/utils/dateUtils.tsR1-R10](diffhunk://#diff-0e33f005e2d49b45dbf3b8595fb7106b1d4164370c030c25894c00716b41a113R1-R10))

### Configuration Update:
* Updated the admin configuration to include the new schedule field (`public/admin/config.yml`).

## 📌 To-Do Before Merging

- [ ] Remove test data

## 🧪 How to Test

- Go to http://localhost:3000/events/fil-bangkok-2024 and verify styling
- Go to `http://localhost:3000/admin/index.html` and try to add a new event with an Event Day schedule.

## 📸 Screenshots

CleanShot 2024-10-10 at 10 14 15@2x

### DIFF
--- a/src/app/events/[slug]/components/ScheduleSection.tsx
+++ b/src/app/events/[slug]/components/ScheduleSection.tsx
@@ -5,12 +5,12 @@ import { PageSection } from '@/components/PageSection'
 import { ScheduleTabs } from './ScheduleTabs'
 
 type ScheduleSectionProps = {
-  schedule: Event['schedule']
+  schedule: NonNullable<Event['schedule']>
 }
 
 export function ScheduleSection({ schedule }: ScheduleSectionProps) {
   return (
-    <PageSection kicker="Join Us" title={schedule?.title || 'Schedule'}>
+    <PageSection kicker="Join Us" title={schedule.title || 'Schedule'}>
       <ScheduleTabs schedule={schedule} />
     </PageSection>
   )
--- a/src/app/events/[slug]/components/ScheduleSection.tsx
+++ b/src/app/events/[slug]/components/ScheduleSection.tsx
@@ -1,0 +1,17 @@
+import type { Event } from '@/types/eventType'
+
+import { PageSection } from '@/components/PageSection'
+
+import { ScheduleTabs } from './ScheduleTabs'
+
+type ScheduleSectionProps = {
+  schedule: Event['schedule']
+}
+
+export function ScheduleSection({ schedule }: ScheduleSectionProps) {
+  return (
+    <PageSection kicker="Join Us" title={schedule!.title || 'Schedule'}>
+      <ScheduleTabs schedule={schedule} />
+    </PageSection>
+  )
+}
--- a/src/app/events/[slug]/components/ScheduleSection.tsx
+++ b/src/app/events/[slug]/components/ScheduleSection.tsx
@@ -10,7 +10,7 @@ type ScheduleSectionProps = {
 
 export function ScheduleSection({ schedule }: ScheduleSectionProps) {
   return (
-    <PageSection kicker="Join Us" title={schedule!.title || 'Schedule'}>
+    <PageSection kicker="Join Us" title={schedule?.title || 'Schedule'}>
       <ScheduleTabs schedule={schedule} />
     </PageSection>
   )
--- a/src/app/events/[slug]/components/ScheduleTabs.tsx
+++ b/src/app/events/[slug]/components/ScheduleTabs.tsx
@@ -8,14 +8,14 @@ import { TextLink } from '@/components/TextLink'
 
 import { formatDate, formatTime } from '../utils/dateUtils'
 
-type ScheduleProps = {
+type ScheduleTabsProps = {
   schedule: Event['schedule']
 }
 
-export function Schedule({ schedule }: ScheduleProps) {
+export function ScheduleTabs({ schedule }: ScheduleTabsProps) {
   return (
-    <TabGroup className="grid gap-6">
-      <TabList className="-m-2 flex gap-4 overflow-auto p-2">
+    <TabGroup className="relative grid gap-6">
+      <TabList className="sticky top-0 -m-2 flex gap-4 overflow-auto bg-brand-800 p-2 lg:static">
         {schedule!.days.map((day) => (
           <Tab
             key={formatDate(day.date)}
--- a/src/app/events/[slug]/components/ScheduleTabs.tsx
+++ b/src/app/events/[slug]/components/ScheduleTabs.tsx
@@ -1,4 +1,9 @@
+'use client'
+
+import { useEffect, useRef } from 'react'
+
 import { Tab, TabGroup, TabList, TabPanel, TabPanels } from '@headlessui/react'
+import theme from 'tailwindcss/defaultTheme'
 
 import type { Event } from '@/types/eventType'
 
@@ -12,11 +17,33 @@ type ScheduleTabsProps = {
   schedule: Event['schedule']
 }
 
+const { screens } = theme
+
 export function ScheduleTabs({ schedule }: ScheduleTabsProps) {
+  const tabGroupRef = useRef<HTMLDivElement>(null)
+  const hasMounted = useRef(false)
+
   const validDays = schedule!.days.filter((day) => day.events.length > 0)
 
+  useEffect(() => {
+    hasMounted.current = true
+  }, [])
+
+  function scrollToTabGroup() {
+    if (hasMounted.current && isScreenBelowLg() && tabGroupRef.current) {
+      tabGroupRef.current.scrollIntoView({
+        behavior: 'smooth',
+        block: 'start',
+      })
+    }
+  }
+
   return (
-    <TabGroup className="relative grid gap-6">
+    <TabGroup
+      ref={tabGroupRef}
+      className="relative grid gap-6"
+      onChange={scrollToTabGroup}
+    >
       <TabList className="sticky top-0 -m-2 flex gap-4 overflow-auto bg-brand-800 p-2 lg:static">
         {validDays.map((day) => (
           <Tab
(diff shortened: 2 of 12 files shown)


barbaraperic:
Maybe we can say something like "No events on this day", or not show the day at all if it doesn't have events?

Screenshot 2024-10-10 at 11 39 01

barbaraperic:
On mobile looks good! Tho there's a slight glitch when you scroll down in the first tab and click on the second tab. It doesn't take you back in the view - maybe something we can implement if not too complicated.

https://github.com/user-attachments/assets/c06d8aed-97b1-484c-875f-63ac0ec06795

src/app/events/[slug]/components/ScheduleSection.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import { ScheduleTabs } from './ScheduleTabs'
+
+type ScheduleSectionProps = {
+  schedule: Event['schedule']
+}
+
+export function ScheduleSection({ schedule }: ScheduleSectionProps) {
+  return (
+    <PageSection kicker="Join Us" title={schedule!.title || 'Schedule'}>
```
Comment (barbaraperic) [inline src/app/events/[slug]/components/ScheduleSection.tsx:13]: schedule?.title ?

```suggestion
    <PageSection kicker="Join Us" title={schedule?.title || 'Schedule'}>
```

src/app/events/[slug]/page.tsx
```diff
@@ -71,6 +72,10 @@ export default function EventEntry({ params }: EventProps) {
         />
       </div>
 
+      {schedule && schedule.days && schedule.days.length > 0 && (
+        <ScheduleSection schedule={schedule} />
+      )}
+
```
Comment (barbaraperic) [inline src/app/events/[slug]/page.tsx:78]: what do you think of extracting conditions in variables? So we have something like this 

```
  const eventHasSchedule = schedule && schedule.days.length > 0
  const eventHasSpeakers = speakers && speakers.length > 0
  const eventHasSponsors = sponsors && Object.keys(sponsors).length > 0
```

...

`      {eventHasSchedule && <ScheduleSection schedule={schedule} />}
`

Comment (mirhamasala) [reply src/app/events/[slug]/page.tsx:78]: I like it a lot.

src/app/events/[slug]/components/ScheduleTabs.tsx
```diff
@@ -0,0 +1,99 @@
...
+  schedule: Event['schedule']
+}
+
+const { screens } = theme
+
+export function ScheduleTabs({ schedule }: ScheduleTabsProps) {
+  const tabGroupRef = useRef<HTMLDivElement>(null)
+  const hasMounted = useRef(false)
+
+  const validDays = schedule!.days.filter((day) => day.events.length > 0)
```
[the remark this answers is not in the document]
Comment (mirhamasala) [reply src/app/events/[slug]/components/ScheduleTabs.tsx:26]: TypeScript think that `schedule` can be `undefined` but, at this point, we know for sure that `schedule` isn't `undefined` since we check for it with `eventHasSchedule` - and only then render the component. 

> You can postfix an expression with ! to tell TypeScript that you know it's not null or undefined. This works the same as an 'as' assertion.

barbaraperic (approved):
Looking good, well done! Left a few non-blocking comments

src/app/events/[slug]/components/ScheduleTabs.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  useEffect(() => {
+    hasMounted.current = true
+  }, [])
```
Comment (CharlyMartin) [inline src/app/events/[slug]/components/ScheduleTabs.tsx:30]: Nice little optimisation - we can import `useIsMounted` from [usehooks-ts](https://usehooks-ts.com/react-hook/use-is-mounted), which is already part of the project :)

src/app/events/[slug]/components/ScheduleTabs.tsx
```diff
@@ -0,0 +1,99 @@
...
+        ))}
+      </TabPanels>
+    </TabGroup>
+  )
+}
+
+function isScreenBelowLg() {
+  return window.matchMedia(`(max-width: ${parseInt(screens.md, 10) - 1}px)`)
+    .matches
+}
```
Comment (CharlyMartin) [inline src/app/events/[slug]/components/ScheduleTabs.tsx:99]: We already rely on `useMediaQuery` from [usehooks-ts](https://usehooks-ts.com/react-hook/use-media-query) in the project. Maybe it can be useful here as well?

src/app/events/[slug]/components/ScheduleTabs.tsx
```diff
@@ -0,0 +1,99 @@
...
+  useEffect(() => {
+    hasMounted.current = true
+  }, [])
+
+  function scrollToTabGroup() {
+    if (hasMounted.current && isScreenBelowLg() && tabGroupRef.current) {
+      tabGroupRef.current.scrollIntoView({
+        behavior: 'smooth',
+        block: 'start',
+      })
```
Comment (CharlyMartin) [inline src/app/events/[slug]/components/ScheduleTabs.tsx:37]: We already have a function `scrollToSection` in the maturity model `utils` folder, in case it can be extracted in the global `utils` reused here. Not sure, I haven't tried.

Super cool you added this by the way, smooth UX!

CharlyMartin (changes requested):

CharlyMartin:
@mirhamasala I added a blocking comment about the prop type of the section components. I believe we should use `NonNullable` instead of `!` to be more accurate / on the safe side. I added more context in the comments, I hope it makes sense.

I also added an event via the CMS, and it worked great. The event I created was added at the end of the list, even though it happened sooner. How about sorting `validDays` by soonest to furthest?

CleanShot 2024-10-11 at 11 15 21

The page looks really good and professional, it will be great for Bangkok! Good job ✨👏

src/app/events/[slug]/components/SpeakersSection.tsx
```diff
@@ -0,0 +1,28 @@
...
+            image={{
+              ...speaker.image,
+              alt: `Photo of ${speaker.name}`,
+            }}
+          />
+        ))}
+      </CardGrid>
+    </PageSection>
+  )
+}
```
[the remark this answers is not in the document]
Comment (mirhamasala) [reply src/app/events/[slug]/components/SpeakersSection.tsx:28]: Ahhhh TS Master 🙇🏽‍♀️

src/app/events/[slug]/components/ScheduleTabs.tsx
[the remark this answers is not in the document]
Comment (mirhamasala) [reply src/app/events/[slug]/components/ScheduleTabs.tsx:26]: I think this is amazing.

src/app/events/[slug]/components/ScheduleTabs.tsx
Comment (mirhamasala) [reply src/app/events/[slug]/components/ScheduleTabs.tsx:37]: Decided not to extract the `scrollToSection` function for now because it is tightly coupled to the concept of a `SectionHash`.

By the way, this made me notice that the function `scrollToSection` uses `document.querySelector`, which in React is considered an anti-pattern because it directly manipulates the DOM outside of React’s control. Instead, React encourages the use of `refs` to interact with DOM elements.

Something like

```
import { useRef } from 'react'

export function scrollToSection(sectionRef: React.RefObject<HTMLElement>) {
  if (!sectionRef.current) {
    console.error(`The referenced element does not exist`)
    return
  }

  sectionRef.current.scrollIntoView({
    block: 'start',
    behavior: 'smooth',
  })
}
```

mirhamasala:
@CharlyMartin 

>How about sorting validDays by soonest to furthest?

Oh, good catch! Thanks so much for the review, Charly! It was super helpful.

src/app/events/[slug]/components/ScheduleTabs.tsx
Comment (CharlyMartin) [reply src/app/events/[slug]/components/ScheduleTabs.tsx:39]: Indeed!

CharlyMartin:
> Oh, good catch! Thanks so much for the review, Charly! It was super helpful.

Anytime 🫡

PR #700, opened 2024-10-07
[UXIT-1415] Refactor Dynamic Images

mirhamasala:
## 📝 Description

This pull request focuses on replacing the `DynamicImage` and `StaticImage` components with the `next/image` component to simplify image handling across the codebase. The most important changes include updating various components to use `next/image`, removing the obsolete `DynamicImage` and `StaticImage` components, and adjusting image properties accordingly.

- **Type:**  Refactor

### Image Handling Updates:

* [`src/app/_components/ArticleHeader.tsx`](diffhunk://#diff-c15186b78b7f8f46d278aabab963f8f02c1bb1e1ba41829d1d2342cfb4b6a0f5R1-L7): Replaced `DynamicImage` with `next/image` and updated image properties to handle fallback scenarios. [[1]](diffhunk://#diff-c15186b78b7f8f46d278aabab963f8f02c1bb1e1ba41829d1d2342cfb4b6a0f5R1-L7) [[2]](diffhunk://#diff-c15186b78b7f8f46d278aabab963f8f02c1bb1e1ba41829d1d2342cfb4b6a0f5L23-L33)
* [`src/app/_components/Card.tsx`](diffhunk://#diff-a83b1c27dc4eef1160c0cd54e260643a1ee188bd35dfcca1fc77cc5c3e0886c8R1-R38): Updated the `Card` component to use `next/image` and adjusted image properties for better handling of dynamic and static images. [[1]](diffhunk://#diff-a83b1c27dc4eef1160c0cd54e260643a1ee188bd35dfcca1fc77cc5c3e0886c8R1-R38) [[2]](diffhunk://#diff-a83b1c27dc4eef1160c0cd54e260643a1ee188bd35dfcca1fc77cc5c3e0886c8L84-L103)
[...]

PR #712, opened 2024-10-11
[UXIT-1596] Fix Card image Fit

CharlyMartin:
## 📝 Description

This PR improves the fit of dynamic images by adding `objectFit: cover`. The `DynamicImage` component previously handled this automatically, but it was removed during the refactor in [#700.](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/700)

## 🛠️ Key Changes
- Adds `objectFit: cover` to `Card` on blog & event pages
- Adds `objectFit: cover` to `PageHeader` on blog & event pages
- Cleans up the prop definitions in the `Card` sub-components by making all props required.

## 📸 Screenshots

Before
CleanShot 2024-10-11 at 16 08 14@2x

After
CleanShot 2024-10-11 at 16 07 55@2x

### DIFF
--- a/src/app/_components/PageHeader.tsx
+++ b/src/app/_components/PageHeader.tsx
@@ -73,8 +73,9 @@ PageHeader.Title = function Title({ children }: TitleProps) {
 
 PageHeader.Image = function PageHeaderImage({
   image,
-}: Required<Pick<PageHeaderProps, 'image'>>) {
+}: Pick<PageHeaderProps, 'image'>) {
   const isStaticImage = 'data' in image
+
   const commonProps = {
     alt: image.alt,
     priority: true,
--- a/src/app/_components/PageHeader.tsx
+++ b/src/app/_components/PageHeader.tsx
@@ -2,7 +2,7 @@ import Image, { type ImageProps } from 'next/image'
 
 import { clsx } from 'clsx'
 
-import type { StaticImageProps } from '@/types/imageType'
+import type { ImageObjectFit, StaticImageProps } from '@/types/imageType'
 
 import { buildImageSizeProp } from '@/utils/buildImageSizeProp'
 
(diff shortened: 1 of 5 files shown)


src/app/_components/PageHeader.tsx
```diff
@@ -22,9 +22,13 @@ type TitleProps = {
   children: string
 }
 
+type PageHeaderImageProps = (StaticImageProps | ImageProps) & {
+  objectFit?: ImageObjectFit
```
mirhamasala, line 26: Is it optional?

src/app/_components/PageHeader.tsx
```diff
@@ -71,12 +75,17 @@ PageHeader.Image = function PageHeaderImage({
...
   const commonProps = {
     alt: image.alt,
     priority: true,
     quality: 100,
     sizes: buildImageSizeProp({ startSize: '100vw', lg: '490px' }),
-    className: 'rounded-lg border border-brand-100',
+    className: clsx(
+      'rounded-lg border border-brand-100',
+      image.objectFit === 'cover' && 'object-cover',
+      image.objectFit === 'contain' && 'object-contain',
```
mirhamasala, line 87: Should we have it default to `cover`?

mirhamasala (approved):
Nice clean up, @CharlyMartin 🙌🏼 Only left a couple of questions. 🙏🏼

src/app/_components/PageHeader.tsx
CharlyMartin: Yes, passing nothing should be possible. Most `PageHeader` don't need `objectFit` for their image.

src/app/_components/PageHeader.tsx
CharlyMartin: In the spirit of our Image refactoring, I made everything explicit so that we know what is going on :)

PR #737, opened 2024-10-16
UXIT-1578 - Security Maturity Model: mobile Table of Contents

barbaraperic:
## 📝 Description

This PR includes mobile version of Table Of Contents for Security Maturity Model page.

- **Type:** New feature

## 🛠️ Key Changes

- new component `MobileTableOfContents`
- I've extracted list items from `DesktopTableOfContents` in its own component, `TableOfContentsList`. To simplify and make it reusable.
- created new `coreFunctionsTableOfContents` function where we extract only `slug` and `title`. Before we had `Content`, which was not necessary for the `TableOfContents` component.

## 🧪 How to Test

- Check mobile and desktop version

## 📸 Screenshots

https://github.com/user-attachments/assets/df248f90-d14b-4e85-b114-5499afb0e264


mirhamasala:
@barbaraperic @CharlyMartin - I didn't check the code - Only the preview - As an early review:

- We probably only want full width on the smaller breakpoints
- I think we'd want the `Table of Contents` text to change as we scroll into different sections - similar to the Desktop implementation

CharlyMartin:
@mirhamasala I implemented the changes you mentioned. I haven't fully reviewed the code, I will do that tomorrow :)

CharlyMartin (approved):
@mirhamasala, I’ve reworked the PR, and it looks good to go from my side! Tagging you for a second review.

As said above: `FormListbox` will most likely need to be broken up into sub-components, as `Card`, but it can be done in a different PR.

src/app/_components/Form/FormError.tsx
```diff
@@ -4,9 +4,14 @@ import { Description } from '@headlessui/react'
...
   error?: string
+  hideError?: boolean
 }
 
-export function FormError({ error }: FormErrorProps) {
+export function FormError({ error, hideError }: FormErrorProps) {
+  if (hideError) {
+    return null
+  }
+
```
mirhamasala, line 14: I still find this an anti-pattern. The `FormError` component should be fully responsible for rendering the error message, while the parent (like `FormField`) could decide whether to pass an error or not, keeping responsibilities separate.

Can you please remind me again why we need this?

src/app/security/maturity-model/components/MobileTableOfContents.tsx
```diff
@@ -0,0 +1,46 @@
...
+    const sectionHash = getHashFromSlug(slug)
+    scrollToSection(sectionHash)
+  }
+
+  const selectedItem = coreFunctionsData.find((item) =>
+    isSectionActive(item.slug),
+  )
+
+  return (
+    <div className="flex justify-center">
```
mirhamasala, line 29: This is making it center aligned when it hits `sm`?
 
CleanShot 2024-10-22 at 14 16 14@2x

mirhamasala (changes requested):
@CharlyMartin - Let's use our custom `Listbox` components if possible instead of `FormListbox`. We can refactor `FormListbox` another time. Thank you.

src/app/_components/Form/FormError.tsx
mirhamasala: Resolving - As we already discussed this! 🙏🏼

CharlyMartin:
@mirhamasala, as discussed, I've reverted the changes to `FormListbox` and re-used the previous implementation.

If we have time in the coming days or weeks, it would be nice to spend some time unifying our use of Listboxes :)

src/app/_components/ListboxOption.tsx
```diff
@@ -20,7 +20,7 @@ type ListboxOptionProps = {
 function OptionContent({ option, counts }: ListboxOptionProps) {
   return (
     <span className="inline-flex items-baseline gap-2">
-      <span className="text-nowrap">{option.name}</span>
+      <span>{option.name}</span>
```
mirhamasala, line 23: This would affect all other instances as well - Do we want that?

src/app/_components/ListboxOptions.tsx
```diff
@@ -15,7 +15,7 @@ export function ListboxOptions({
   return (
     <HeadlessUIListboxOptions
       className={clsx(
-        'absolute z-10 mt-2 overflow-hidden rounded-lg border border-brand-100 bg-brand-800 py-2 text-brand-100 focus:brand-outline focus-within:outline-2',
+        'absolute z-10 mt-2 max-w-sm overflow-hidden rounded-lg border border-brand-100 bg-brand-800 py-2 text-brand-100 focus:brand-outline focus-within:outline-2',
```
mirhamasala, line 18: What's the addition for?

src/app/security/maturity-model/components/MobileTableOfContents.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  return (
+    <nav aria-label="Table of Contents" className="w-full max-w-sm">
+      <HeadlessUIListbox
+        value={selectedOption}
+        onChange={(value) => handleChange(value.id)}
+      >
+        <ListboxButton>
+          <div className="inline-flex w-full items-center gap-2">
+            <Icon component={ArrowsDownUp} />
```
mirhamasala, line 33: It doesn't make a lot of sense to me to have an `ArrowsDownUp` icon here?

CleanShot 2024-10-22 at 20 17 30@2x

src/app/security/maturity-model/components/MobileTableOfContents.tsx
```diff
@@ -0,0 +1,56 @@
...
+            <span className="grow truncate text-left">
+              {selectedOption?.name || 'Select a section'}
+            </span>
+            <Icon component={CaretDown} size={16} weight="bold" />
+          </div>
```
mirhamasala, line 38: It doesn't make a lot of sense to me to have an `ArrowsDownUp` icon here?

CleanShot 2024-10-22 at 20 17 30@2x


```suggestion
          <span className="grow truncate text-left">
            {selectedOption?.name || 'Select a section'}
          </span>
          <Icon component={CaretDown} size={16} weight="bold" />
```

src/app/security/maturity-model/components/MobileTableOfContents.tsx
```diff
@@ -0,0 +1,56 @@
...
+        </ListboxButton>
+
+        <ListboxOptions position="left">
```
mirhamasala, line 41: It's already the default.

```suggestion
        <ListboxOptions>
```

mirhamasala (changes requested):
@CharlyMartin - This doesn't seem to be working as expected. Please check the comments. 🙏🏼

src/app/security/maturity-model/utils/useUrlHash.ts
mirhamasala, line 1: First step scout CR ~ Leaving it here for consideration.

```
'use client'

import { useEffect, useState } from 'react'

import { useRouter, usePathname } from 'next/navigation'

import type { Route } from 'next'

const HASH_SIGN = '#'

type SectionHash = `${typeof HASH_SIGN}${string}`
type NewHashEvent = HashChangeEvent | CustomEvent<SectionHash>

export function useUrlHash() {
  const pathname = usePathname()
  const router = useRouter()

  const [hash, setHash] = useState(getCurrentHash)

  useEffect(() => {
    if (windowIsAvailable()) {
      window.addEventListener('hashchange', onHashChange)
      return () => window.removeEventListener('hashchange', onHashChange)
    }
  }, [hash])

  function updateHash(sectionId: string) {
    const sectionHash = getHashFromSlug(sectionId)

    if (sectionHash !== hash) {
      router.replace(`${pathname}${sectionHash}` as Route, { scroll: false })
      triggerCustomHashChange(sectionHash)
    }
  }

  function clearHash() {
    router.replace(pathname as Route, { scroll: false })
    triggerCustomHashChange(HASH_SIGN)
  }

  function clearHashIfPresent(sectionId: string) {
    if (getHashFromSlug(sectionId) === hash) {
      clearHash()
    }
  }

  function isSectionActive(sectionId: string) {
    return getHashFromSlug(sectionId) === hash
  }

  function onHashChange(event: NewHashEvent) {
    setHash(getHashFromEvent(event))
  }

  return {
    updateHash,
    clearHash,
    clearHashIfPresent,
    isSectionActive,
    getHashFromSlug,
  }
}

function getCurrentHash() {
  return windowIsAvailable()
    ? ((window.location.hash || HASH_SIGN) as SectionHash)
    : HASH_SIGN
}

function getHashFromEvent(event: NewHashEvent) {
  if (event instanceof HashChangeEvent) {
    const slug = event.newURL.split(HASH_SIGN)[1] || ''
    return `${HASH_SIGN}${slug}` as SectionHash
  }
  return event.detail
}

function getHashFromSlug(slug: string) {
  return `${HASH_SIGN}${slug}` as SectionHash
}

function triggerCustomHashChange(newHash: SectionHash) {
  if (windowIsAvailable()) {
    window.dispatchEvent(new CustomEvent('hashchange', { detail: newHash }))
  }
}

function windowIsAvailable() {
  return typeof window !== 'undefined'
}
```

src/app/security/maturity-model/components/MobileTableOfContents.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+const options = coreFunctionsData.map(({ slug, title }) => ({
+  id: slug,
+  name: title,
+}))
+
+export function MobileTableOfContents() {
+  const { updateHash, isSectionActive, getHashFromSlug } = useUrlHash()
+
+  const selectedOption = options.find((option) => isSectionActive(option.id))
```
mirhamasala, line 23: Your message is mostly clear, but it can be refined for better flow and clarity. Here’s a revised version:

This isn’t working as expected. Please refer to the console errors and browser behavior.

An easy fix would be:

```
const defaultOption = options[0]
const selectedOption = options.find((option) => isSectionActive(option.id)) || defaultOption
```

However, as mentioned in the other PR, it would be ideal if we didn’t have to do this workaround.

We can try using:

`const defaultOption = coreFunctionsData.find((item) => item.order === 1)!`

…with the non-null assertion (!) to enforce a default, but it seems like coreFunctionsBaseData isn’t enforcing the order type. Could we ensure that the order type is properly handled in coreFunctionsBaseData?

CharlyMartin: If we want `.find()` to always return a non-null value in places where we know the value exists for sure, we can create this helper function:

```ts
function findOrFail<T>(array: T[], fn: (item: T) => boolean): T {
  const found = array.find(fn);
  if (!found) {
    throw new Error('Item not found');
  }
  return found;
}
```

If someday the value doesn't exist anymore, the build will crash :)

CharlyMartin:
WIO

src/app/security/maturity-model/components/MobileTableOfContents.tsx
mirhamasala: >Regarding having a default, coreFunctionsBaseData is properly typed; 

When I changed `order` to `order: 0` in `/src/app/security/maturity-model/content/leadership-and-organizational-principles.md` - Nothing yelled at me.

Maybe `findOrFail` is in this case a good solution. It seems important to me that the first chapter would be the default? And not the second, etc. Though, I don't know if this is ever an issue, since it seems to already populate correctly by the time the user scrolls to the beginning?

CharlyMartin: I added a zod validation and a sort function so that the first chapter will always be first! If you change the order to zero -> fails

CharlyMartin:
@mirhamasala I'm Sorry about this lengthy review process on this one... The `Listbox` confusion could be a candidate for a mini-postmortem.

As for this PR:
- I reverted the changes to `ListboxOptions` / `ListboxOption` and duplicated some of the code to make `MobileTableOfContents` work.
- I also added a zod validation step to the maturity model content to be on the safe side.
- Finally, I reworked the `useUrlHash` using your suggestion

I hope this works!

PR #740, opened 2024-10-16
[UXIT-1301] Rework Event Sort

CharlyMartin:
## 📝 Description

This pull request reworks the entry sorting mechanism in the app to implement new custom sort options for the Events page.

_New feature & Refactor_

## 🛠️ Key Changes

Previously, `useSort` would receive object properties, such as `startDate`, to handle sorting. This made the hook too generic for our new needs, which now involve a combination of filtering and sorting.

```ts
const { sortQuery, sortedResults } = useSort({
  searchParams,
  entries: search-results,
  sortBy: 'startDate',
  defaultSortId: DEFAULT_SORT_OPTION.chronological,
})
```

Now, `useSort` accepts a list of `SortConfig` with a sort key, a sort label, and a sort function, which makes it much more customizable.

```ts
// app/events/data/sortData.ts
export const eventsSortData = [
  {
    key: 'upcoming-events',
    label: 'Upcoming Events',
    sortFn: getUpcomingEvents,
  },
  {
    key: 'past-events',
    label: 'Past Events',
    sortFn: getPastEvents,
  },
  {
    key: 'all-events',
    label: 'All Events',
    sortFn: sortEventsDesc,
  },
]

// app/events/page.tsx
const { sortQuery, sortedResults, defaultSortQuery } = useSort({
  searchParams,
  entries: searchResults,
  configs: eventsSortData,
})

```
⚠️ Side note: Previously, `useSort` would explicitly accept a `defaultSortId` to initialize sorting. With this new implementation, this is implicit; `useSort` treats the first `SortConfig` object as the default. It can also be made explicit by passing a `defaultConfig` to the hook if you prefer.

## 🧪 How to Test

1. Go to `/events` and make sure the sorting works.
2. Also test the blog and ecosystem pages, as they were impacted by the rework.

## 📸 Screenshots

CleanShot 2024-10-18 at 10 45 11


src/app/_hooks/useSort.ts
```diff
@@ -1,71 +1,49 @@
...
-  const sortedResults = useMemo(() => {
-    if (ALPHABETICAL_SORT_IDS.includes(validatedSortId)) {
-      return sortEntriesAlphabetically({
-        entries,
-        sortId: validatedSortId as AlphabeticalSortId,
-      })
-    }
+  if (!defaultConfig) {
+    throw new Error('At least one sort config must be provided')
+  }
```
mirhamasala, line 27: I feel like at this point in the code - it shouldn't be possible for `defaultConfig` to be `undefined`? Can't TS magic help us here?

```
type UseSortProps<Entry extends Object> = {
  entries: Array<Entry>
  configs: readonly [SortConfig<Entry>, ...Array<SortConfig<Entry>>]
}
```

```
export const blogSortData: readonly [
  SortConfig<BlogPost>,
  ...Array<SortConfig<BlogPost>>,
] = [
  {
    key: 'newest',
    label: 'Newest',
    sortFn: sortPostsByDateDesc,
  },
  {
    key: 'oldest',
    label: 'Oldest',
    sortFn: sortPostsByDateAsc,
  },
] as const
```

src/app/_utils/sortBlogPosts.ts
mirhamasala, line 1: What do you think we continue to move these functions closer to their pages? Into `pages/blog/utils` in this case?

src/app/ecosystem-explorer/data/sortData.ts
mirhamasala, line 1: Isn't this more `constants/sortConstants`?

mirhamasala (commented):
@CharlyMartin - Could this be your best PR to date? I really like what you did here. Some naming and organization nits; a [TS suggestion](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/740/files#r1806653382), and one that I find more important [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/740/files#r1806643772). 🙌🏼🧀

src/app/_hooks/useSort.ts
```diff
@@ -1,71 +1,49 @@
...
 export function useSort<Entry extends Object>({
   searchParams,
   entries,
-  sortBy,
-  defaultSortId,
+  configs,
 }: UseSortProps<Entry>) {
-  const normalizedQuery = normalizeQueryParam(searchParams, SORT_KEY)
-  const validatedSortId = getValidatedSortId(normalizedQuery, defaultSortId)
+  const defaultConfig = configs[0]
```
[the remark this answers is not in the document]
CharlyMartin: Yes you're right, that was one of my concerns.

Your solution works. The potential drawback is that nothing prevents two objects from being `isDefault: true`, but maybe it's a minor problem.

Another solution would be to pass the default key to the hook like we used to do:

```ts
  const { sortQuery, sortedResults, defaultSortQuery } = useSort({
    searchParams,
    entries: searchResults,
    configs: eventsSortData,
    defaultSort: 'upcoming-events',
  })
```

Let me know which one you prefer :) I have a slight preference for the second one.

src/app/_hooks/useSort.ts
CharlyMartin: Yes absolutely! For some reason, I couldn't make it work. Let me try that again :)

src/app/ecosystem-explorer/data/sortData.ts
CharlyMartin: I like `constants/sortConfigs` 👍

mirhamasala:
@CharlyMartin LGTM! ❤️

PR #743, opened 2024-10-18
[UXIT-1587] Allocators Table Search

CharlyMartin:
## 📝 Description

This PR adds a search feature to the allocators table, fixes the allocators data fetching, and reactors some table-related code.

## 🛠️ Key Changes

1. Implemented table-wide search using TanStack Table's built-in [global filter function](https://tanstack.com/table/v8/docs/guide/global-filtering#global-filter-function)
2. Added a [fixed layout](https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout#fixed) to the allocators table so that columns keep their size when the data changes
3. Fixed allocators' data fetching by adding a missing token
4. Created a `Table` component to encapsulate table-related code
5. Moved column definitions from the `components` folder and into the `data` one
6. Used Headless UI components for `SearchInput`. They generate unique IDs automatically, so passing a custom ID should no longer be necessary. To be tested.

## 🧪 How to Test

1. Go to `/filecoin-plus/allocators` and test the search feature. The search query will match data from all columns.
2. Test whether ids generated by Headless UI pass the accessibility problem we used to run into (duplicate ids.)

## 📸 Screenshots

CleanShot 2024-10-22 at 13 20 02@2x


src/app/security/bug-bounty/data/leaderboardTableColumnDef.tsx
```diff
@@ -5,13 +5,13 @@ import { createColumnHelper } from '@tanstack/react-table'
...
 
-import { leaderboardData } from '../data/leaderboard'
+import { leaderboardData } from './leaderboard'
 
 type WhiteHat = (typeof leaderboardData)[number]
 
 const columnHelper = createColumnHelper<WhiteHat>()
 
-export const leaderboardTableData = [
+export const leaderboardTableColumnDef = [
```
mirhamasala, line 14: This code is closely tied to UI rendering, as it defines table columns and handles icon rendering for profile links, which are part of the component’s structure. Since it’s specific to how the table and icons are displayed, I believe it belongs in the `components` folder rather than the `data` folder to keep UI logic organized. I would also rename it to `leaderboardColumns`. I also believe we already made this move once..

src/app/_components/SearchInput.tsx
```diff
@@ -1,46 +1,38 @@
+import { Input, Label, Field, Button } from '@headlessui/react'
```
mirhamasala, line 1: Nice! 👌🏽

src/app/filecoin-plus/allocators/data/allocatorsTableColumnDef.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { createColumnHelper } from '@tanstack/react-table'
+
+import { CustomLink } from '@/components/CustomLink'
+import { Icon } from '@/components/Icon'
+
+import type { Allocator } from '../schemas/AllocatorSchema'
+
+const columnHelper = createColumnHelper<Allocator>()
+
+export const allocatorsTableColumnDef = [
```
mirhamasala, line 11: Same comment [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/743/files#r1810767152)

`components/allocatorsColumns`

src/app/filecoin-plus/allocators/utils/getAllocatorUrlList.ts
```diff
@@ -1,20 +1,15 @@
+import { Octokit } from '@octokit/rest'
```
mirhamasala, line 1: I remember you mentioned trying Octokit - I didn't get if it was necessary to keep? If so - Why do we need this?

src/app/security/bug-bounty/data/leaderboardTableColumnDef.tsx
mirhamasala, line 1: Scout CR: We should use `CustomLink` if possible. Same as in the current `allocatorsTableColumnDef`.

```        <CustomLink
          href={link}
          className="inline-flex items-center justify-center gap-2 focus:brand-outline"
          aria-label={`Apply for ${name} allocator`}
        >
          <span>Apply</span>
          <Icon component={ArrowUpRight} size={20} />
        </CustomLink>
      )
```

Can possibly be refactored into a reusable shared component under `components/Table`?

src/app/filecoin-plus/allocators/utils/getAllocatorUrlList.ts
```diff
@@ -27,3 +22,13 @@ function isValidAllocatorFile(file: AllocatorFileMetaDataBase) {
   const allocatorFilePattern = /^\d{3,4}\.json$/
   return allocatorFilePattern.test(file.name)
 }
+
+async function getAllocatorList() {
```
mirhamasala, line 26: Move before `extractAllocatorUrls`?

src/app/filecoin-plus/allocators/components/FilterableAllocatorsTable.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { type Allocator } from '../schemas/AllocatorSchema'
+
+import { AllocatorsTable } from './AllocatorsTable'
+import { NoFilterResultsMessage } from './NoFilterResultsMessage'
+
+type FilterableAllocatorsTableProps = {
+  data: Array<Allocator>
+}
+
+export function FilterableAllocatorsTable({
```
mirhamasala, line 22: What if we call this component `FilterContainer` like we do for the main shared `FilterContainer` layout?

src/app/filecoin-plus/allocators/components/AllocatorsTableSection.tsx
```diff
@@ -2,7 +2,7 @@ import * as Sentry from '@sentry/nextjs'
 
 import { getAllocators } from '../utils/getAllocators'
 
-import { AllocatorsTable } from './AllocatorsTable'
+import { FilterableAllocatorsTable } from './FilterableAllocatorsTable'
 import { NoAllocatorsMessage } from './NoAllocatorsMessage'
```
mirhamasala, line 6: Scout CR Nit: `NoDataAvailableMessage`?

mirhamasala (changes requested):
This is looking really good @CharlyMartin - Mostly nits. And so happy I could actually review it! Tagging @barbaraperic for a second review! 🙌🏼

src/app/security/bug-bounty/data/leaderboardTableColumnDef.tsx
CharlyMartin: Well, this is hybrid; it contains both data and UI code.

But, if I had to choose, I’d say it looks more like data than components. It’s an array of config objects, not a function returning JSX. We already do something similar in a few places, like `grants/data/applicationProcessData.tsx`, for instance.

That’s just my two cents on the matter, your call :)

src/app/filecoin-plus/allocators/utils/getAllocatorUrlList.ts
CharlyMartin: I also tried to use Octokit in the other fetch function but couldn't figure out a quick implementation. We might want to do it at some point though :)

src/app/filecoin-plus/allocators/components/FilterableAllocatorsTable.tsx
```diff
@@ -0,0 +1,57 @@
...
+  return (
+    <>
+      <div className="mb-6 max-w-readable">
+        <SearchInput query={searchQuery} onChange={setGlobalFilter} />
+      </div>
+
+      {hasSearchResults ? (
+        <AllocatorsTable headerGroups={headerGroups} rowModel={rowModel} />
+      ) : (
+        <NoFilterResultsMessage />
```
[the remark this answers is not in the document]
CharlyMartin: Ah nice! I forgot about this component :)

CharlyMartin:
Thanks @mirhamasala! I believe I implemented all your comments ✨

I also replied to your first [comment](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/743#discussion_r1810962981). I'll update depending on what you think :)

src/app/filecoin-plus/allocators/utils/getAllocatorUrlList.ts
mirhamasala: >I also tried to use Octokit in the other fetch function but couldn't figure out a quick implementation. We might want to do it at some point though :)

From what I understand, using Octokit for the second case (fetching file contents directly from a URL) isn’t a good use case. 

> Octokit is ideal for interacting with the GitHub API, such as fetching repository metadata or directory contents, but it doesn’t handle raw file downloads. Since the raw file URLs are direct HTTP requests, the fetch function is more appropriate in this scenario. Octokit simplifies API interactions, while fetch is best suited for retrieving raw file content. Therefore, keeping both methods makes sense for these distinct use cases.

src/app/security/bug-bounty/data/leaderboardTableColumnDef.tsx
mirhamasala: Yes ~ Better like that! ❤️

src/app/security/bug-bounty/data/leaderboardTableColumnDef.tsx
```diff
@@ -51,14 +52,14 @@ export const leaderboardTableData = [
       const Logo = getProfileLinkLogo(profileLink)
 
       return (
-        <a
+        <CustomLink
           href={profileLink}
           rel="noopener noreferrer"
```
mirhamasala, line 57: Already part of `CustomLink`
```suggestion
```

src/app/filecoin-plus/allocators/components/FilterableAllocatorsTable.tsx
mirhamasala: That's true. What about `AllocatorTableWithFilters`?

src/app/security/bug-bounty/data/leaderboardTableColumnDef.tsx
CharlyMartin: Ok, let's do that then :)

src/app/filecoin-plus/allocators/utils/getAllocatorUrlList.ts
CharlyMartin: Good to know!

src/app/filecoin-plus/allocators/components/FilterableAllocatorsTable.tsx
CharlyMartin: Let's do `AllocatorTableWithFilters`!

src/app/security/bug-bounty/data/leaderboardTableColumnDef.tsx
CharlyMartin: I've extracted both table links into their own local components. I'm not sure there's a lot of value in creating another link component in `components/Table`.

src/app/filecoin-plus/allocators/components/FilterableAllocatorsTable.tsx
CharlyMartin: > It might end up looking very similar to FilterContainer

The key difference here is that `AllocatorTableWithFilters` holds the state for the table and filters, while `FilterContainer` is a real stateless container that can contain anything.

So, we may need to keep `AllocatorsTableWithFilters` and create another proper container component for the layout of our table filters. We can do that in the next PR when we add sorting.

src/app/security/bug-bounty/components/SocialIconLink.tsx
```diff
@@ -0,0 +1,31 @@
...
+type ProfileLinkValueType = string | null
+type SocialIconLinkProps = CellContext<WhiteHat, ProfileLinkValueType>
+
+export function SocialIconLink({ getValue, row }: SocialIconLinkProps) {
+  const profileLink = getValue()
+  const reporter = row.original.reporter
+
+  if (!profileLink) {
+    return null
+  }
```
mirhamasala, line 18: `SocialIconLink` should only be responsible for rendering.

src/app/filecoin-plus/allocators/data/allocatorsTableColumnsData.ts
```diff
@@ -0,0 +1,54 @@
...
+    },
+  }),
+  columnHelper.accessor('application.allocation_bookkeeping', {
+    header: 'Apply',
+    meta: {
+      headerCellStyle: 'w-32',
+      bodyCellStyle: 'text-brand-300',
+    },
+    enableGlobalFilter: false,
+    cell: ApplyLink,
```
mirhamasala, line 52: Same comment [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/743/files#r1812345674)

CharlyMartin: @mirhamasala I'm confused, I thought we agreed on this pattern - `cell: ApplyLink` - in a [comment](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/743#discussion_r1810967008) above?

I can definitely update it, but then we will be using a UI component in the data table, which you didn't like?

Let me know :)

mirhamasala (approved):
Thank you, Charly! ☀️ Let's get this merged! 🙌🏼

PR #748, opened 2024-10-22
UXIT-1501 - Testing: remove trim and truncate text

barbaraperic:
## 📝 Description

Removed `truncateText` and `cleanAndTrim` as we have `textIsClamped` in `Description` component that shortens the text. This might be cause the mismatch between server and client side HTML. 

- **Type:** Bug fix


mirhamasala (approved):
Wow! So this is interesting. I actually think I remember adding this to try to fix the error. Since it didn't fix the error, we might as well remove it. I did push a tiny CR since we wouldn't need `generatePreviewDescription` anymore.

PR #754, opened 2024-10-24
[UXIT-1588] Allocators Filters

CharlyMartin:
## 📝 Description

This PR adds two select filters to the allocator table: type and region.

I originally tried to use the table's internal to implement it, but that approach didn’t work out as expected. So, I switched to using React's `useState` to manage the filter and sort states.

## 🛠️ Key Changes

1. Stopped using `setGlobalFilter` in favor of column filters: the search input now filters only the “name” column
2. Created `SelectFilter` to manage type and location filters state
3. Added `AllocatorsTableFiltersLayout` to handle the positioning of the search and select inputs
4. Created `getColumnOrThrow` util to throw an error if `table.getColumn` returns `undefined`. This util function follows the same principle as `findOrThrow` to make the code simpler.
5. Created `createUniqueOptionsFromData` to dynamically build a list of options from the allocators' data

## 🧪 How to Test

1. Go to `/filecoin-plus/allocators`
2. Test the search input, it should only search through the allocators' name
3. Test the region and type filters

## 📸 Screenshots

CleanShot 2024-10-28 at 10 29 19

CleanShot 2024-10-28 at 10 29 45

## 🔖 Resources

- https://tanstack.com/table/v8/docs/guide/column-filtering
- https://tanstack.com/table/latest/docs/api/core/table#getcolumn


src/app/filecoin-plus/allocators/components/AllocatorsTableWithFilters.tsx
```diff
@@ -22,30 +30,55 @@ type AllocatorsTableWithFiltersProps = {
...
+          />
+        }
+        locationFilter={
+          <SelectFilter
+            column={locationColumn}
+            options={locationOptions}
+            resetOptionName="All Regions"
+          />
+        }
+      />
```
mirhamasala, line 81: From [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/754/files#r1819025319)

```suggestion
          <AllocatorsTableFiltersLayout data={data} table={table} />
```

src/app/filecoin-plus/allocators/components/SelectFilter.tsx
```diff
@@ -0,0 +1,78 @@
...
+  name: string
+}
+
+type SortListboxProps = {
+  column: Column<Allocator>
+  options: ReadonlyArray<FilterOption>
+  resetOptionName: string
+}
+
+const EMPTY_FILTER_VALUE = ''
```
mirhamasala, line 27:
```suggestion
const DEFAULT_FILTER_ID = ''
```

src/app/filecoin-plus/allocators/components/SelectFilter.tsx
```diff
@@ -0,0 +1,78 @@
...
+  options,
+  resetOptionName,
+}: SortListboxProps) {
+  const currentFilterValue =
+    column.getFilterValue()?.toString() || EMPTY_FILTER_VALUE
+
+  const resetOption: FilterOption = {
+    id: EMPTY_FILTER_VALUE,
+    name: resetOptionName,
+  }
```
mirhamasala, line 40: `resetOption` and `resetOptionName` throws me off a bit. What about something like..

```suggestion
    const defaultOption: FilterOption = {
      id: DEFAULT_FILTER_ID,
      name: defaultOptionLabel,
    }
```

And then

```
<SelectFilter
  column={typeColumn}
  options={typeOptions}
  defaultOptionLabel="All Types"
/>
```

src/app/filecoin-plus/allocators/components/SelectFilter.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+const EMPTY_FILTER_VALUE = ''
+
+export function SelectFilter({
+  column,
+  options,
+  resetOptionName,
+}: SortListboxProps) {
+  const currentFilterValue =
+    column.getFilterValue()?.toString() || EMPTY_FILTER_VALUE
```
mirhamasala, line 35: Do we need `|| EMPTY_FILTER_VALUE`?

src/app/filecoin-plus/allocators/utils/createUniqueOptionsFromData.ts
```diff
@@ -0,0 +1,34 @@
...
+function countOccurrences(array: Array<string>) {
+  return array.reduce(
+    (object, item) => {
+      const curentCount = object[item] || 0
+      object[item] = curentCount + 1
+      return object
+    },
+    {} as Record<string, number>,
+  )
+}
```
mirhamasala, line 34:
```suggestion
  const valuesArray = data.map((item) => item[key])
  const valueCounts = countOccurrences(valuesArray)

  const sortedUniqueOptions = Object.keys(valueCounts)
    .sort((a, b) => valueCounts[b] - valueCounts[a])
    .map((value) => ({
      id: value,
      name: value,
    }))

  return sortedUniqueOptions
}

function countOccurrences(array: Array<string>) {
  return array.reduce((counts, item) => {
    counts[item] = (counts[item] || 0) + 1
    return counts
  }, {} as Record<string, number>)
}
```

mirhamasala: Sidenote - I'm happy to keep `reduce` but I do think it tends to throw off people. `forEach` can do the same thing.

```
function countOccurrences(array: Array<string>) {
  const counts: Record<string, number> = {}

  array.forEach((item) => {
    if (counts[item]) {
      counts[item] += 1
    } else {
      counts[item] = 1
    }
  })

  return counts
}
```

mirhamasala: Lastly, what about `generateSortedOptions`?

src/app/filecoin-plus/allocators/components/AllocatorsTableFiltersLayout.tsx
```diff
@@ -0,0 +1,24 @@
+import type { ReactNode } from 'react'
+
+type AllocatorsTableFiltersLayoutProps = {
+  nameSearch: ReactNode
+  typeFilter: ReactNode
+  locationFilter: ReactNode
```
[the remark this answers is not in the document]
CharlyMartin: I used the same logic as in `FilterContainer` and created a stateless container component in order to keep the state in the parent.

src/app/filecoin-plus/allocators/components/SelectFilter.tsx
CharlyMartin: The `toString()` part is to make TypeScript happy as the return type of `getFilterValue` is `unknown`.

https://tanstack.com/table/v8/docs/api/features/column-filtering#getfiltervalue

CharlyMartin: But it looks like we don't need it anymore in this scenario! 👌

src/app/filecoin-plus/allocators/components/SelectFilter.tsx
CharlyMartin: I'm happy with the renaming, good suggestions. Not a big fan of of declaring a `DEFAULT_OPTION` constant and then overwriting it with a prop.

I've reworked the component, let me know what you think of this new version :)

src/app/filecoin-plus/allocators/utils/createUniqueOptionsFromData.ts
CharlyMartin: Love it :)

src/app/filecoin-plus/allocators/components/AllocatorsTableWithFilters.tsx
CharlyMartin: I reworked it as per your suggestion

CharlyMartin:
Thanks for the suggestions @mirhamasala!

- I've used more descriptive names in `SelectFilter`
- I've reworked `generateSortedOptions` as you suggested
- Finally, I've encapsulated the filter components in `AllocatorsTableFilters`

mirhamasala (approved):
@CharlyMartin - Thank you for the rework. This looks great! ❤️🚀

PR #765, opened 2024-10-28
[UXIT-1638] Allocators Sort

CharlyMartin:
## 📝 Description

This PR adds a sorting feature to the allocators' table.

## 🛠️ Key Changes

- Introduced `SelectSort` to interact with the column's sorting state
- Added a local `constants/sorting` file to manage the sort defaults, and highlight the dependency between the table state and the component's default sort option.

## 🧪 How to Test

- Go to `/filecoin-plus/allocators` and try sorting the table
- Change the default table sort in `constants/sorting` and make sure nothing breaks. Switching `desc: false` to `true` should automatically update the component's default sort option.

## 📸 Screenshots

CleanShot 2024-10-29 at 13 39 17@2x

## 🔖 Resources

- About [satisfies](https://www.youtube.com/watch?v=r1L35zxZQPE)


src/app/filecoin-plus/allocators/components/SelectSort.tsx
```diff
@@ -0,0 +1,78 @@
...
+        ))}
+      </HeadlessUIListboxOptions>
+    </HeadlessUIListbox>
+  )
+
+  function handleSortChange(newOption: TableSortOption) {
+    if (newOption.id !== currentSortId) {
+      column.toggleSorting()
+    }
+  }
```
mirhamasala, line 77: This implementation of `toggleSorting()` might not be intended. You can reproduce it by trying A-Z, Z-A, and then A-Z again - The latter, you'll have to hit twice.

The problem seem to be that `column.toggleSorting()` without any arguments cycles through the sorting states: asc, desc, and none. This means:

- If the current state is asc and newOption.id is desc, calling toggleSorting() once will set it to desc, but another call would set it to none.
- Because we’re not explicitly passing a direction, it often takes multiple clicks to reach the desired state, especially when switching between asc and desc.

So, this version leads to unpredictable behavior where it can take multiple clicks to reach the intended sort order. It effectively relies on the component’s internal state and cycles through it, which is inconsistent.

In the suggested version, column.toggleSorting(newOption.id === 'desc') directly sets the sort direction based on newOption.id:

- Passing newOption.id === 'desc' tells toggleSorting exactly which direction to sort (ascending or descending) without cycling through states.
- When newOption.id is 'asc', it sorts in ascending order; when it’s 'desc', it sorts in descending order immediately, eliminating the need for multiple clicks.

```suggestion
    function handleSortChange(newOption: TableSortOption) {
      column.toggleSorting(newOption.id === 'desc')
    }
```

src/app/filecoin-plus/allocators/constants/sorting.ts
```diff
@@ -0,0 +1,28 @@
...
+] as const satisfies Array<TableSortOption>
+
+export const NAME_SORT_CONFIG = {
+  INITIAL_TABLE_STATE: NAME_INITIAL_TABLE_STATE,
+  OPTIONS: NAME_SORT_OPTIONS,
+  DEFAULT_OPTION: findOrThrow(
+    NAME_SORT_OPTIONS,
+    (option) => option.id === DEFAULT_SORT_OPTION_ID,
+  ),
+} as const
```
mirhamasala, line 28: I'm finding this quite hard to read. And even though I saw your comment about `satisfies` - I'm failing to grasp the usefulness here. Let's Huddle later?

```suggestion
import type { TableSortOption } from '../components/SelectSort'

const nameSortOptions: Array<TableSortOption> = [
  { id: 'asc', name: 'Organization (A-Z)' },
  { id: 'desc', name: 'Organization (Z-A)' },
] as const

export const nameSortConfig = {
  initialTableState: {
    id: 'name',
    desc: false,
  },
  options: nameSortOptions,
  defaultOption: nameSortOptions[0],
} as const
```

src/app/filecoin-plus/allocators/components/SelectSort.tsx
CharlyMartin: You are absolutely right @mirhamasala, great catch! I don't know why I didn't catch it... Anyway, fixed it!

src/app/filecoin-plus/allocators/constants/sorting.ts
CharlyMartin: `satisfies` is a looser way to type values, I did my last skill stretch on it 😅 Anyway, I'm not attached to it, it can go!

`satisfies` aside, what I'm trying to do here is ensure that the default option is the same one as defined in the table's initial state.

CharlyMartin: I reworked it to make it closer to your suggestions. Let me know what you think :)

src/app/filecoin-plus/allocators/components/SelectSort.tsx
```diff
@@ -0,0 +1,77 @@
...
+            <span className="invisible mb-px group-data-[selected]:visible">
+              <Icon component={Check} size={20} />
+            </span>
+          </HeadlessUIListboxOption>
+        ))}
+      </HeadlessUIListboxOptions>
+    </HeadlessUIListbox>
+  )
+
+  function handleSortChange(newOption: TableSortOption) {
```
CharlyMartin, line 73: Move at the top

src/app/filecoin-plus/allocators/constants/sorting.ts
```diff
@@ -0,0 +1,20 @@
...
+  id: 'name',
+  desc: false,
+} as const
+
+const defaultSortId = initialTableState.desc ? 'desc' : 'asc'
+
+const ascSortOption = { id: 'asc', name: 'Organization (A-Z)' } as const
+const descSortOption = { id: 'desc', name: 'Organization (Z-A)' } as const
+
+const NAME_COLUMN_SORT_OPTIONS =
```
CharlyMartin, line 11: Remove the column sorting

src/app/filecoin-plus/allocators/constants/sorting.ts
```diff
@@ -0,0 +1,20 @@
...
+  defaultSortId === 'asc'
+    ? [ascSortOption, descSortOption]
+    : [descSortOption, ascSortOption]
+
+export const NAME_COLUMN_SORT_CONFIG = {
+  initialTableState,
```
CharlyMartin, line 17: Rename `initialReactTableState`

mirhamasala (approved):
Naise, Charly-🧀. Let's go! ❤️

PR #768, opened 2024-10-29
[UXIT-1577] New Design Sponsor Section

barbaraperic:
## 📝 Description

Implementation of the [new events section design](https://www.figma.com/design/0pBfQs5tKI6bx5ypPIfVRA/FIL.org-V4-Prototypes?node-id=7070-80392&t=i5a8hUuz0fM7gSp5-4).

- **Type:**  Refactor

## 🛠️ Key Changes

- Layout change

## 🧪 How to Test

- Check the Figma designs

## 📸 Screenshot
Screenshot 2024-10-29 at 16 08 16
Screenshot 2024-10-29 at 16 08 59


src/app/events/[slug]/utils/sponsorTierConfigs.ts
```diff
@@ -0,0 +1,28 @@
+import { buildImageSizeProp } from '@/utils/buildImageSizeProp'
+
+export const sponsorTierConfigs = [
```
Comment (mirhamasala) [inline src/app/events/[slug]/utils/sponsorTierConfigs.ts:3]:
```suggestion

type SponsorTierConfig = {
  tier: 'firstTier' | 'secondTier' | 'thirdTier'
  gap: string
  logoImageConfig: {
    containerStyles: string
    imageSizes: string
  }
}

export const sponsorTierConfigs: Array<SponsorTierConfig> = [
```

src/app/events/[slug]/utils/sponsorTierConfigs.ts
```diff
@@ -0,0 +1,28 @@
...
+    tier: 'firstTier',
+    gap: 'gap-8 sm:gap-20',
+    logoImageConfig: {
+      containerStyles: 'h-16 w-44 sm:h-20',
+      imageSizes: buildImageSizeProp({ startSize: '140px', sm: '180px' }),
+    },
+  },
+  {
+    tier: 'secondTier',
+    gap: 'gap-6 sm:gap-16 ',
```
Comment (mirhamasala) [inline src/app/events/[slug]/utils/sponsorTierConfigs.ts:14]:
```suggestion
    gap: 'gap-6 sm:gap-16',
```

src/app/events/[slug]/utils/sponsorTierConfigs.ts
Comment (mirhamasala) [inline src/app/events/[slug]/utils/sponsorTierConfigs.ts:1]: If this is a constant, should it be under `constants`?

src/app/events/[slug]/components/SponsorContainer.tsx
```diff
@@ -3,21 +3,26 @@ import { clsx } from 'clsx'
...
+  gap = 'gap-8',
+}: SponsorContainerProps) {
   return (
-    <div className={clsx('grid gap-8', gridClassName)}>
+    <div
+      className={clsx(
+        gap,
+        'flex flex-col flex-wrap justify-start sm:flex-row sm:justify-center sm:gap-8',
+      )}
+    >
```
Comment (mirhamasala) [inline src/app/events/[slug]/components/SponsorContainer.tsx:25]: Charly suggested the other day to define the known styles first .. I think it makes sense!

```suggestion
    <div
      className={clsx(
        'flex flex-col flex-wrap justify-start sm:flex-row sm:justify-center sm:gap-8',
        'gap'
      )}
    >
```

Comment (mirhamasala) [reply src/app/events/[slug]/components/SponsorContainer.tsx:25]: Also - Shall we move `sm:gap-8` to be part of the default?

src/app/events/[slug]/components/SponsorContainer.tsx
```diff
@@ -3,21 +3,26 @@ import { clsx } from 'clsx'
 import { type SponsorLogoProps, SponsorLogo } from './SponsorLogo'
 import type { Sponsor } from './SponsorSection'
 
-type SponsorGridProps = {
+type SponsorContainerProps = {
   sponsors: Array<Sponsor>
   tier: string
-  gridClassName: string
   logoImageConfig: SponsorLogoProps['logoImageConfig']
+  gap?: string
```
Comment (mirhamasala) [inline src/app/events/[slug]/components/SponsorContainer.tsx:10]: Is there a reason for the `gap` to be optional? It seems like something we want to control.

mirhamasala (changes requested):
I think it looks good on the front. Will have another look once some of the changes are in! 🙏🏼

src/app/events/[slug]/components/SponsorLogo.tsx
```diff
@@ -25,13 +20,13 @@ export function SponsorLogo({
...
       <Image
         fill
         src={sponsor.image.src}
         alt={`${sponsor.name} Logo - ${formattedSponsorTier} sponsor`}
         sizes={logoImageConfig.imageSizes}
-        className="object-contain object-left-bottom"
+        className="object-contain"
       />
       {sponsor.website && (
         <a
```
Comment (CharlyMartin) [inline src/app/events/[slug]/components/SponsorLogo.tsx:31]: Just wondering why we have `target="_blank"` here? We usually don't. Just a question.

Love the link as an absolute top layer :)

CharlyMartin (approved):
It's a lot of boxes on boxes indeed :) Works well on my side!

I pushed a small commit to re-organize the sponsor types, but it doesn't change any functionality.

src/app/events/[slug]/components/SponsorLogo.tsx
Comment (mirhamasala) [reply src/app/events/[slug]/components/SponsorLogo.tsx:31]: @CharlyMartin is right - Will remove `target="_blank"`  which brings me to [Create External Link Component](https://www.notion.so/filecoin/FF-Create-ExternalLink-component-1307631f282580109a6fef9e98f174eb?pvs=4)

Comment (mirhamasala) [reply src/app/events/[slug]/components/SponsorLogo.tsx:31]: Pushed a CR!

PR #772, opened 2024-10-30
[UXIT-1663] Extend Share Article Feature with Copy Link

barbaraperic:
## 📝 Description

Add CopyToClipboard to Share Article Feature

🚨 NOTE: Icons should have tooltips, like Filipa [pointed here](https://filecoinfoundation.slack.com/archives/C06MTEQ3SP6/p1730286048043609). I'll add in a separate PR.

- **Type:** New feature

## 🛠️ Key Changes

- new component `CopyToClipboard`

## 📌 To-Do Before Merging

- [x] Check with Filipa if the button should have a color change/animation/text display after the user clicks on it → [Slack convo](https://filecoinfoundation.slack.com/archives/C06MTEQ3SP6/p1730286048043609)

## 🧪 How to Test

- Go to a blog post
- Click copy link
- Check if it is copied

## 📸 Screenshots

https://github.com/user-attachments/assets/6f4d8a95-629a-460c-9588-9f27ca645cb9


src/app/_components/CopyToClipboard.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import { Icon } from '@/components/Icon'
+import { NotificationDialog } from '@/components/NotificationDialog'
+
+type CopyToClipboardProps = {
+  text: string
+}
+
+export function CopyToClipboard({ text }: CopyToClipboardProps) {
+  const [_, copy] = useCopyToClipboard()
```
CharlyMartin, line 19: The `_` part can be removed if not used :)

```suggestion
  const [, copy] = useCopyToClipboard()
```

src/app/_components/CopyToClipboard.tsx
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  return (
+    <>
+      <NotificationDialog
+        isOpen={isCopied}
+        setIsOpen={setIsCopied}
+        title="Link copied to clipboard!"
+      />
+      <button
```
CharlyMartin, line 40: Nit: Just sharing what I also told Mirha, I try to use Headless UI component whenever possible. But not a blocker, up to you!

https://headlessui.com/react/button

src/app/_components/CopyToClipboard.tsx
```diff
@@ -0,0 +1,48 @@
...
+      Sentry.captureException(error)
+    }
+  }
+
+  return (
+    <>
+      <NotificationDialog
+        isOpen={isCopied}
+        setIsOpen={setIsCopied}
+        title="Link copied to clipboard!"
```
CharlyMartin, line 38: The notification message should probably also be passed as a prop. What is the thing to copy is not a link?

tailwind.config.js
```diff
@@ -127,6 +127,12 @@ module.exports = {
             paddingRight: theme('padding.10'),
           },
         },
+        '.touch-target': {
+          padding: theme('padding.2'),
```
CharlyMartin, line 131: Nice addition :)

Just wondering, will that always be `padding.2`? If I remember correctly, some touch targets need more than `2`.

If so, we could have multiple touch target classes such as `touch-target-2` or `touch-target-xs`. Just a tought.

CharlyMartin (changes requested):
Nice clean PR :) I've suggested some light edits.

Two side comments:
1. We should spend some time refactoring the `NotificationDialog` in another PR, I've added a comment in [Notion](https://www.notion.so/filecoin/FF-Followup-Dev-Work-d6616163f43c4e7db78edf218cebcc8d?pvs=4#1307631f282580e39331cbc81a55f0c3). This is definitely for another PR, I'm starting to love these small PRs too much 😄
2. I'm working on adding tooltips to the table on my current PR, so I'll make sure the code is re-usable for you

tailwind.config.js
mirhamasala: This is a great thought - but this keeps changing and is not something we can extract yet. I would place it back as it was for now.

mirhamasala (changes requested):
@barbaraperic Super cool! 🥰 Few nits, and agree with the messages that @CharlyMartin left.

Can we also make sure to add a ticket to the Notion board about adding a tooltip? And can we share this addition with the team in #ff-website? Can be under the same [thread](https://filecoinfoundation.slack.com/archives/C06MTEQ3SP6/p1729678126971119) but helpful to tick the "also send to channel" box. Thank you!

src/app/_components/CopyToClipboard.tsx
barbaraperic: Sounds good! Just for me to understand, what is the benefit in this case?

tailwind.config.js
barbaraperic: Reverted back to constants 👍 Just for me to understand @mirhamasala - why are those changing? Are they not defined in the designs?

barbaraperic:
> @barbaraperic Super cool! 🥰 Few nits, and agree with the messages that @CharlyMartin left.
> 
> Can we also make sure to add a ticket to the Notion board about adding a tooltip? And can we share this addition with the team in #ff-website? Can be under the same [thread](https://filecoinfoundation.slack.com/archives/C06MTEQ3SP6/p1729678126971119) but helpful to tick the "also send to channel" box. Thank you!

Yes, added [[FF] Add tooltips to Share Post icons](https://www.notion.so/filecoin/FF-Add-tooltips-to-Share-Post-icons-12f7631f28258030a5f1c889c0032208?pvs=4) plus an extra fix [[BUG FF] Fix animation for NotificationDialog](https://www.notion.so/filecoin/BUG-FF-Fix-animation-for-NotificationDialog-12f7631f28258038bbe6cb92235c3e39?pvs=4)

src/app/_components/CopyToClipboard.tsx
CharlyMartin: In this case I'm not sure, but they usually take care of some a11y things for us.

tailwind.config.js
CharlyMartin: These exist to "augment" small HTML elements to be at least 48x48px, the minimum accessible touch target. How much they should be augmented depends on the original size of the element, so it's hard to know in advance. Does that make sense?

barbaraperic: Interesting issue, makes sense. Tho I'm wondering if there's a workaround where we have .touch-target class with min-width and min-height 48px which can be applied to elements which don't meet the requirement. But I guess that's for another discussion. Thanks @CharlyMartin !

mirhamasala: We might not be talking about the same – We want to maintain the elements’ defined sizes from the design while expanding their clickable areas for improved touch targets. Rather than adjusting width and height, we use padding to increase the interactive space around each element without visually altering its dimensions. To counter the shift caused by the padding, we apply negative margins to realign the element within its intended layout. This approach ensures that touch targets are larger, enhancing usability, while preserving the visual integrity of the design.

barbaraperic: Make sense, thank you!

PR #777, opened 2024-10-31
[UXIT-1589] Include Datacap Stats to Allocator page

barbaraperic:
## 📝 Description

Add Datacap Stats to Allocators Page

Screenshot 2024-10-31 at 10 33 44

### DIFF
--- a/src/app/filecoin-plus/allocators/components/AllocatorsTableSection.tsx
+++ b/src/app/filecoin-plus/allocators/components/AllocatorsTableSection.tsx
@@ -1,6 +1,7 @@
 import * as Sentry from '@sentry/nextjs'
 
 import { getAllocators } from '../utils/getAllocators'
+import { getDatacapStats } from '../utils/getDatacapStats'
 
 import { AllocatorsTableWithFilters } from './AllocatorsTableWithFilters'
 import { NoDataAvailableMessage } from './NoDataAvailableMessage'
@@ -9,11 +10,25 @@ export async function AllocatorsTableSection() {
   try {
     const allocators = await getAllocators()
 
+    const datacapStats = await getDatacapStats()
+
+    const allocatorsWithDatacap = allocators.map((allocator) => {
+      const stats = datacapStats.find(
+        (stats) => stats.address === allocator.address,
+      )
+
+      return {
+        ...allocator,
+        remainingDatacap: stats ? stats.remainingDatacap : null,
+        allowance: stats ? stats.allowance : null,
+      }
+    })
+
     if (!allocators.length) {
       return <NoDataAvailableMessage />
     }
 
-    return <AllocatorsTableWithFilters data={allocators} />
+    return <AllocatorsTableWithFilters data={allocatorsWithDatacap} />
   } catch (error) {
     console.error('Error fetching or validating allocators:', error)
 
--- a/src/app/filecoin-plus/allocators/data/allocatorsTableColumnsData.tsx
+++ b/src/app/filecoin-plus/allocators/data/allocatorsTableColumnsData.tsx
@@ -1,9 +1,9 @@
 import { createColumnHelper } from '@tanstack/react-table'
 
 import { ApplyLink } from '../components/ApplyLink'
-import type { Allocator } from '../schemas/AllocatorSchema'
+import type { AllocatorWithDatacap } from '../schemas/AllocatorSchema'
 
-const columnHelper = createColumnHelper<Allocator>()
+const columnHelper = createColumnHelper<AllocatorWithDatacap>()
 
 export const allocatorsTableColumnsData = [
   columnHelper.accessor('name', {
@@ -31,8 +31,31 @@ export const allocatorsTableColumnsData = [
     },
     filterFn: 'includesString',
   }),
+  columnHelper.accessor('remainingDatacap', {
+    header: 'Remaining Datacap',
+    cell: (info) => info.getValue(),
+    meta: {
+      headerCellStyle: 'w-44',
+    },
+    filterFn: 'includesString',
+  }),
+  columnHelper.accessor('allowance', {
+    header: 'Used Datacap',
+    cell: (info) => info.getValue(),
+    meta: {
+      headerCellStyle: 'w-44',
+    },
+    filterFn: 'includesString',
+  }),
   columnHelper.accessor('application.required_replicas', {
     header: 'Required Replicas',
+    cell: (info) => info.getValue(),
+    meta: {
+      headerCellStyle: 'w-44',
+    },
+  }),
+  columnHelper.accessor('application.required_sps', {
+    header: 'Required SPs',
     cell: (info) => info.getValue(),
     meta: {
       headerCellStyle: 'w-44',
--- a/src/app/filecoin-plus/allocators/schemas/AllocatorSchema.ts
+++ b/src/app/filecoin-plus/allocators/schemas/AllocatorSchema.ts
@@ -1,4 +1,5 @@
 import { z } from 'zod'
+import { DatacapSchema } from './DatacapSchema'
 
 const ApplicationSchema = z.object({
   required_sps: z.string(),
@@ -8,6 +9,7 @@ const ApplicationSchema = z.object({
 
 export const AllocatorSchema = z.object({
   name: z.string(),
+  address: z.string(),
   metapathway_type: z.string(),
   location: z.string(),
   application: ApplicationSchema,
@@ -42,8 +44,11 @@ export const AllocatorFileListMetaDataBaseSchema = z.array(
   AllocatorFileMetaDataBaseSchema,
 )
 
+export const AllocatorWithDatacapSchema = AllocatorSchema.merge(DatacapSchema)
+
 export type Allocator = z.infer<typeof AllocatorSchema>
 export type AllocatorFileMetaDataBase = z.infer<
   typeof AllocatorFileMetaDataBaseSchema
 >
 export type AllocatorFileMetaData = z.infer<typeof AllocatorFileMetaDataSchema>
+export type AllocatorWithDatacap = z.infer<typeof AllocatorWithDatacapSchema>
--- a/src/app/filecoin-plus/allocators/schemas/DatacapSchema.ts
+++ b/src/app/filecoin-plus/allocators/schemas/DatacapSchema.ts
@@ -1,0 +1,9 @@
+import { z } from 'zod'
+
+export const DatacapSchema = z.object({
+  address: z.string(),
+  remainingDatacap: z.string().nullable(),
+  allowance: z.string().nullable(),
+})
+
+export type Datacap = z.infer<typeof DatacapSchema>
--- a/src/app/filecoin-plus/allocators/utils/getDatacapStats.ts
+++ b/src/app/filecoin-plus/allocators/utils/getDatacapStats.ts
@@ -1,0 +1,18 @@
+import { DatacapSchema, type Datacap } from '../schemas/DatacapSchema'
+
+const api = 'https://api.datacapstats.io/api/getVerifiers'
+
+export async function getDatacapStats(): Promise<Datacap[]> {
+  try {
+    const response = await fetch(api)
+    const { data } = await response.json()
+    const datacapStats = data.map((item: any) => {
+      return DatacapSchema.parse(item)
+    })
+
+    return datacapStats
+  } catch (error) {
+    console.error(error)
+    return []
+  }
+}
--- a/src/app/filecoin-plus/allocators/components/AllocatorsTable.tsx
+++ b/src/app/filecoin-plus/allocators/components/AllocatorsTable.tsx
@@ -6,11 +6,11 @@ import { Table } from '@/components/Table/Table'
 import { TableBody } from '@/components/Table/TableBody'
 import { TableHead } from '@/components/Table/TableHead'
 
-import { type Allocator } from '../schemas/AllocatorSchema'
+import { type AllocatorWithDatacap } from '../schemas/AllocatorSchema'
 
 type AllocatorsTableProps = {
-  headerGroups: Array<HeaderGroup<Allocator>>
-  rowModel: RowModel<Allocator>
+  headerGroups: Array<HeaderGroup<AllocatorWithDatacap>>
+  rowModel: RowModel<AllocatorWithDatacap>
 }
 
 export function AllocatorsTable({
--- a/src/app/filecoin-plus/allocators/components/AllocatorsTableFilters.tsx
+++ b/src/app/filecoin-plus/allocators/components/AllocatorsTableFilters.tsx
@@ -1,7 +1,7 @@
 import type { Table } from '@tanstack/react-table'
 
 import { NAME_COLUMN_SORT_CONFIG } from '../constants/sorting'
-import type { Allocator } from '../schemas/AllocatorSchema'
+import type { AllocatorWithDatacap } from '../schemas/AllocatorSchema'
 import { generateSortedOptions } from '../utils/generateSortedOptions'
 import { getColumnOrThrow } from '../utils/getColumnOrThrow'
 
@@ -10,7 +10,7 @@ import { SelectFilter } from './SelectFilter'
 import { SelectSort } from './SelectSort'
 
 type AllocatorsTableFiltersProps = {
-  table: Table<Allocator>
+  table: Table<AllocatorWithDatacap>
 }
 
 export function AllocatorsTableFilters({ table }: AllocatorsTableFiltersProps) {
--- a/src/app/filecoin-plus/allocators/components/AllocatorsTableWithFilters.tsx
+++ b/src/app/filecoin-plus/allocators/components/AllocatorsTableWithFilters.tsx
@@ -15,13 +15,13 @@ import { NoSearchResultsMessage } from '@/components/NoSearchResultsMessage'
 
 import { NAME_COLUMN_SORT_CONFIG } from '../constants/sorting'
 import { allocatorsTableColumnsData } from '../data/allocatorsTableColumnsData'
-import { type Allocator } from '../schemas/AllocatorSchema'
+import { type AllocatorWithDatacap } from '../schemas/AllocatorSchema'
 
 import { AllocatorsTable } from './AllocatorsTable'
 import { AllocatorsTableFilters } from './AllocatorsTableFilters'
 
 type AllocatorsTableWithFiltersProps = {
-  data: Array<Allocator>
+  data: Array<AllocatorWithDatacap>
 }
 
 export function AllocatorsTableWithFilters({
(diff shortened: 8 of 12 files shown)


src/app/filecoin-plus/allocators/components/AllocatorsTableSection.tsx
```diff
@@ -1,19 +1,40 @@
 import * as Sentry from '@sentry/nextjs'
 
+import { convertDatacapToPiB } from '../utils/convertDatacapToPiB'
 import { getAllocators } from '../utils/getAllocators'
+import { getDatacapStats } from '../utils/getDatacapStats'
 
 import { AllocatorsTableWithFilters } from './AllocatorsTableWithFilters'
 import { NoDataAvailableMessage } from './NoDataAvailableMessage'
 
+async function fetchAllocatorsWithDatacap() {
```
mirhamasala, line 10: Should this be moved to `utils`?

src/app/filecoin-plus/allocators/utils/convertDatacapToPiB.ts
mirhamasala, line 1: What if we used something like https://github.com/sindresorhus/pretty-bytes?

src/app/filecoin-plus/allocators/utils/getDatacapStats.ts
```diff
@@ -0,0 +1,27 @@
+import { API_URLS } from '@/constants/apiUrls'
+
+import { DatacapSchema, type Datacap } from '../schemas/DatacapSchema'
+
+async function fetchDatacapStats(): Promise<any> {
```
mirhamasala, line 5: No need for return types.

src/app/filecoin-plus/allocators/utils/getDatacapStats.ts
```diff
@@ -0,0 +1,27 @@
...
+  return response.json()
+}
+
+function parseDatacapData(data: any): Array<Datacap> {
+  return data.map((item: any) => {
+    return DatacapSchema.parse(item)
+  })
+}
+
+export async function getDatacapStats(): Promise<Array<Datacap>> {
```
mirhamasala, line 19: Should be placed at the top.

src/app/filecoin-plus/allocators/utils/getDatacapStats.ts
```diff
@@ -0,0 +1,27 @@
...
+
+async function fetchDatacapStats(): Promise<any> {
+  const response = await fetch(API_URLS.datacapStats)
+  if (!response.ok) {
+    throw new Error(`Failed to fetch datacap stats, status: ${response.status}`)
+  }
+  return response.json()
+}
+
+function parseDatacapData(data: any): Array<Datacap> {
```
mirhamasala, line 13: No any.

src/app/filecoin-plus/allocators/schemas/DatacapSchema.ts
```diff
@@ -0,0 +1,9 @@
+import { z } from 'zod'
+
+export const DatacapSchema = z.object({
+  address: z.string(),
+  remainingDatacap: z.string().nullable(),
+  allowance: z.string().nullable(),
```
mirhamasala, line 6: Are these actually `nullable`?

src/app/filecoin-plus/allocators/schemas/AllocatorSchema.ts
```diff
@@ -42,8 +45,11 @@ export const AllocatorFileListMetaDataBaseSchema = z.array(
   AllocatorFileMetaDataBaseSchema,
 )
 
+export const AllocatorWithDatacapSchema = AllocatorSchema.merge(DatacapSchema)
+
 export type Allocator = z.infer<typeof AllocatorSchema>
```
CharlyMartin, line 50: Are we still using this type?

src/app/filecoin-plus/allocators/schemas/AllocatorSchema.ts
```diff
@@ -42,8 +45,11 @@ export const AllocatorFileListMetaDataBaseSchema = z.array(
   AllocatorFileMetaDataBaseSchema,
 )
 
+export const AllocatorWithDatacapSchema = AllocatorSchema.merge(DatacapSchema)
```
CharlyMartin, line 48: Nit: We usually use [extend](https://zod.dev/README?id=extend) for these operations.

Is there a reason for `merge` here?

src/app/filecoin-plus/allocators/utils/getDatacapStats.ts
```diff
@@ -0,0 +1,31 @@
...
+    throw new Error(`Failed to fetch datacap stats, status: ${response.status}`)
+  }
+  return response.json()
+}
+
+function parseDatacapData(data: Array<BaseDatacap>) {
+  return data.map((item: BaseDatacap) => {
+    return DatacapSchema.parse(item)
+  })
+}
```
CharlyMartin, line 31: How about something like that? It's a bit simpler.


```suggestion
import { API_URLS } from '@/constants/apiUrls'

import {
  GetDatacapStatsResponseSchema,
  type Datacap,
} from '../schemas/DatacapSchema'

export async function getDatacapStats(): Promise<Array<Datacap>> {
  const response = await fetch(API_URLS.datacapStats)
  const json = await response.json()
  const parsedJson = GetDatacapStatsResponseSchema.parse(json)
  return parsedJson.data
}
```

If you like it, you would have to create this schema:

```
export const GetDatacapStatsResponseSchema = z.object({
  data: z.array(BaseDatacapSchema),
})
```

src/app/filecoin-plus/allocators/utils/getAllocatorsWithDatacap.ts
```diff
@@ -0,0 +1,27 @@
...
+      allowance: formatDatacap(stats?.allowance),
+    }
+  })
+
+  return allocatorsWithDatacap
+}
+
+function formatDatacap(value?: string) {
+  return value ? prettyBytes(Number(value)) : 'No data available'
+}
```
CharlyMartin, line 27: If feel like this formatting/rendering logic should go in `allocatorsTableColumnsData`. `getAllocatorsWithDatacap` should only be concerned about returning the data.

Something like

```ts
cell: (info) => info.getValue() || 'No data available'
```

What do you think?

package.json
```diff
@@ -46,6 +46,7 @@
     "next-plausible": "^3.12.1",
     "nuqs": "^2.1.1",
     "pluralize": "^8.0.0",
+    "pretty-bytes": "^6.1.1",
```
CharlyMartin, line 49: I didn't know this one, it's cool :)

CharlyMartin (changes requested):
I made some suggestions but I couldn't test all of them because I hit the API rate limit 🥲 I hope they make sense!

src/app/filecoin-plus/allocators/data/allocatorsTableColumnsData.tsx
```diff
@@ -46,6 +62,13 @@ export const allocatorsTableColumnsData = [
     ),
     cell: (info) => info.getValue(),
   }),
+  columnHelper.accessor('application.required_sps', {
+    header: 'Required SPs',
```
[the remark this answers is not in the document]
barbaraperic: Updated the branch with main, it's there 👍

src/app/filecoin-plus/allocators/utils/getDatacapStats.ts
barbaraperic: I kind of prefer the modular approach with error handling, but if you think it's clearer, don't mind going with it - @mirhamasala any thoughts?

src/app/filecoin-plus/allocators/utils/getAllocatorsWithDatacap.ts
barbaraperic: I like that!

barbaraperic: Actually on this one, we need to provide fallback here 

```
    return {
      ...allocator,
      remainingDatacap: stats?.remainingDatacap ?? 'No data available',
      allowance: stats?.allowance ?? 'No data available',
    }
```
Wondering if it would make sense to leave the format function rather than provide an empty string or 'No data available' - any thoughts?

src/app/filecoin-plus/allocators/schemas/AllocatorSchema.ts
CharlyMartin: Ah, my bad! I thought they were interchangeable, but they have a slight difference.

Your first implementation was better than what I suggested, sorry!

```ts
A.merge(B)
// is equivalent to
A.extend(B.shape)
```

CharlyMartin: Both seem to behave the same, though

> If the two schemas share keys, the properties of B overrides the property of A. The returned schema also inherits the "unknownKeys" policy (strip/strict/passthrough) and the catchall schema of B.

src/app/filecoin-plus/allocators/utils/getDatacapStats.ts
CharlyMartin: In my opinion, all the error handling we need is in `allocators/components/AllocatorsTableSection.tsx`.

But I couldn't test due to the API rate limits so, maybe this is needed.

src/app/filecoin-plus/allocators/data/allocatorsTableColumnsData.tsx
```diff
@@ -63,3 +72,7 @@ export const allocatorsTableColumnsData = [
     },
   }),
 ]
+
+function formatDatacap(value?: string) {
+  return value ? prettyBytes(Number(value)) : 'No data available'
+}
```
mirhamasala, line 78: Shall we move this out to `utils`?

src/app/filecoin-plus/allocators/schemas/DatacapSchema.ts
mirhamasala, line 1: I'm finding the naming here slightly confusing. Here are some suggestions but I'm not sure if they're the right ones, since I'm not entirely sure based on the current naming what the schemas represent.

- Is `DatacapSchema` too generic? Is this the minimal/basic version? Is this `MinimalDatacapSchema`?
- I find `BaseDatacapSchema` confusing because "Base" typically implies a simpler, foundational version, but this schema seems rather comprehensive? And I also don't see it extended anywhere? Is this `DetailedDatacapSchema`?
- `AllowanceArrayItemSchema` could better describe what these items represent. Would `DatacapAllowanceHistorySchema` be better?

src/app/filecoin-plus/allocators/utils/getAllocatorsWithDatacap.ts
```diff
@@ -0,0 +1,21 @@
+import { getAllocators } from '../utils/getAllocators'
+import { getDatacapStats } from '../utils/getDatacapStats'
+
+export async function getAllocatorsWithDatacap() {
+  const allocators = await getAllocators()
+  const datacapStats = await getDatacapStats()
```
mirhamasala, line 6: This change runs both API calls in parallel using `Promise.all()` rather than sequentially. Running them in parallel is more efficient since we don't need to wait for `getAllocators()` to finish before starting `getDatacapStats()`. This can significantly reduce the total execution time, especially when the API calls are slow.

```suggestion
  const [allocators, datacapStats] = await Promise.all([
    getAllocators(),
    getDatacapStats()
  ]);
```

src/app/filecoin-plus/allocators/utils/getAllocatorsWithDatacap.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  const allocatorsWithDatacap = allocators.map((allocator) => {
+    const stats = datacapStats.find(
+      (stats) => stats.address === allocator.address,
+    )
```
mirhamasala, line 11: The change replaces `find()` with a Map lookup. While `find()` has to search through the entire array each time, Map provides instant lookups by address. This makes the code more efficient, especially with larger datasets, since we only build the Map once instead of searching repeatedly.

```suggestion
  const statsMap = new Map(datacapStats.map((stats) => [stats.address, stats]))
```

src/app/filecoin-plus/allocators/data/allocatorsTableColumnsData.tsx
mirhamasala: Asked @filipagr and crew to clarify `No data available` [>>](https://filecoinfoundation.slack.com/archives/C06MTEQ3SP6/p1731241180071039)

src/app/filecoin-plus/allocators/utils/getAllocatorsWithDatacap.ts
mirhamasala: Done.

src/app/filecoin-plus/allocators/utils/getAllocatorsWithDatacap.ts
mirhamasala: Done.

src/app/filecoin-plus/allocators/schemas/DatacapSchema.ts
mirhamasala: Done.

src/app/filecoin-plus/allocators/data/allocatorsTableColumnsData.tsx
mirhamasala: Done. Still pending response from Filipa and crew.

src/app/filecoin-plus/allocators/utils/getDatacapStats.ts
mirhamasala: You might be talking about two things? Charly's approach could still be reworked to be modular? 

```
import { API_URLS } from '@/constants/apiUrls'
import { GetDatacapStatsResponseSchema, type Datacap } from '../schemas/DatacapSchema'

export async function getDatacapStats() {
 const json = await fetchDatacapStats()
 return parseDatacapStats(json)
}

async function fetchDatacapStats() {
 const response = await fetch(API_URLS.datacapStats)
 return response.json()
}

function parseDatacapStats(json: unknown) {
 return GetDatacapStatsResponseSchema.parse(json).data
}
```

mirhamasala: What do you all think about using some `revalidate`? Something like

```
 const response = await fetch(API_URLS.datacapStats, {
    next: { revalidate: 3600 } // Cache for 1 hour 
  })
```

src/app/filecoin-plus/allocators/data/allocatorsTableColumnsData.tsx
mirhamasala: As just discussed - We figured out this is about inconsistencies between addresses in the two different datasets. Will follow up with the Filecoin Plus team.

src/app/filecoin-plus/allocators/utils/formatDatacap.ts
```diff
@@ -0,0 +1,8 @@
+import prettyBytes from 'pretty-bytes'
+
+export function formatDatacap(bytes?: string) {
+  if (bytes === undefined || bytes === '') return 'N/A'
```
CharlyMartin, line 4: Nit: Both `undefined` and `''` are falsy values, so it could simplified as:


```suggestion
  if (!bytes) return 'N/A'
```

But what you have also works fine :)

src/app/filecoin-plus/allocators/utils/formatDatacap.ts
```diff
@@ -0,0 +1,8 @@
...
+  const num = Number(bytes)
+  if (isNaN(num) || !Number.isFinite(num) || num === 0) return 'N/A'
```
CharlyMartin, line 6: We should use `Number.isNaN` here, instead of the legacy `isNaN`.

As per [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN): The `isNaN()` function determines whether a value is `NaN`, **first converting the value to a number if necessary.**

```ts
isNaN("Charly")
// true
Number.isNaN("Charly")
// false
```


```suggestion
  if (Number.isNaN(num) || !Number.isFinite(num) || num === 0) return 'N/A'
```

CharlyMartin (changes requested):

src/app/filecoin-plus/allocators/schemas/DatacapSchema.ts
```diff
@@ -0,0 +1,49 @@
...
+})
+
+const APIDatacapSchema = z.object({
+  id: z.number(),
+  addressId: z.string(),
+  address: z.string(),
+  auditTrail: z.string(),
+  retries: z.number(),
+  name: z.string().nullish(),
+  orgName: z.string().nullish(),
```
barbaraperic, line 25: Screenshot 2024-11-13 at 07 57 23

src/app/filecoin-plus/allocators/utils/formatDatacap.ts
barbaraperic: In this case, I think we want to be explicit, because we want to show 0B if the value is 0 (also falsy)

src/app/filecoin-plus/allocators/schemas/AllocatorSchema.ts
```diff
@@ -45,9 +45,11 @@ export const AllocatorFileListMetaDataBaseSchema = z.array(
...
 )
 
-export const AllocatorWithInternalDatacapSchema = AllocatorSchema.extend(
-  InternalDatacapSchema.shape,
-)
+export const AllocatorWithInternalDatacapSchema = AllocatorSchema.extend({
+  address: InternalDatacapSchema.shape.address,
+  remainingDatacap: InternalDatacapSchema.shape.allowance,
+  usedDatacap: InternalDatacapSchema.shape.remainingDatacap,
+})
```
mirhamasala, line 52: Isn't this still weird? 

`remainingDatacap` is `allowance` and `usedDatacap` is `remainingDatacap`?

barbaraperic: It's confusing, but yes. Here's the screenshot of comparison between Filecoin Plus table and Filecoin Allocators. 

Screenshot 2024-11-13 at 08 38 07
Screenshot 2024-11-13 at 08 38 16

src/app/filecoin-plus/allocators/utils/formatDatacap.ts
CharlyMartin: Ok, works for me!

src/app/filecoin-plus/allocators/schemas/AllocatorSchema.ts
barbaraperic: We renamed the `remainingDatacap` to `availableDatacap` to avoid confusion.export const 

```
AllocatorWithInternalDatacapSchema = AllocatorSchema.extend({
  address: InternalDatacapSchema.shape.address,
  availableDatacap: InternalDatacapSchema.shape.allowance,
  usedDatacap: InternalDatacapSchema.shape.remainingDatacap, // remainingDatacap shows how much datacap allocator has used
})
```

PR #786, opened 2024-11-04
[UXIT-1699] - Save Event Schedule Tabs’ State in URL

CharlyMartin:
## 📝 Description

This PR syncs the state of `Tabs` (on the event's individual page) with the URL query: `/events/fil-bangkok-2024?tab=1`

This will persist the active schedule tab, so when users click on “View Details” and then return to the page, they land on the same tab.

## 🛠️ Key Changes

- Installs `nuqs` and uses `useQueryState` to sync the state with the URL query

## 🧪 How to Test

- Go to `/events/fil-bangkok-2024`
- Select a day in the schedule -> the URL should update
- Click on `View Details` on one of the events and then go back -> you should land on the same tab

⚠️ One limitation is that, on `xs` devices, the last tab can be active and out of the viewport simultaneously.

## 📸 Screenshots

CleanShot 2024-11-05 at 12 04 33

## 🔖 Resources

- https://nuqs.47ng.com/docs/adapters#nextjs-app-router
- https://nuqs.47ng.com/docs/basic-usage
- https://headlessui.com/react/tabs#controlling-the-selected-tab

### DIFF
--- a/src/app/events/[slug]/components/ScheduleSection/Tabs.tsx
+++ b/src/app/events/[slug]/components/ScheduleSection/Tabs.tsx
@@ -10,7 +10,7 @@ import { useIsMounted, useMediaQuery } from 'usehooks-ts'
 import type { Event } from '../../../types/eventType'
 import { formatDate } from '../../utils/dateUtils'
 import { filterAndSortScheduleDays } from '../../utils/filterAndSortScheduleDays'
-import { scrollRefIntoView } from '../../utils/scrollRefIntoView'
+import { scrollTabContentIntoView } from '../../utils/scrollTabContentIntoView'
 
 import { EventDetails } from './EventDetails'
 
@@ -43,7 +43,7 @@ export function Tabs({ schedule }: TabsProps) {
     setActiveTabIndex(index)
 
     if (isMounted() && isScreenBelowLg) {
-      scrollRefIntoView(tabGroupRef)
+      scrollTabContentIntoView(tabGroupRef)
     }
   }
 
--- a/src/app/events/[slug]/components/ScheduleSection/Tabs.tsx
+++ b/src/app/events/[slug]/components/ScheduleSection/Tabs.tsx
@@ -10,6 +10,7 @@ import { useIsMounted, useMediaQuery } from 'usehooks-ts'
 import type { Event } from '../../../types/eventType'
 import { formatDate } from '../../utils/dateUtils'
 import { filterAndSortScheduleDays } from '../../utils/filterAndSortScheduleDays'
+import { scrollRefIntoView } from '../../utils/scrollRefIntoView'
 
 import { EventDetails } from './EventDetails'
 
@@ -41,11 +42,8 @@ export function Tabs({ schedule }: TabsProps) {
   function handleIndexChange(index: number) {
     setActiveTabIndex(index)
 
-    if (isMounted() && isScreenBelowLg && tabGroupRef.current) {
-      tabGroupRef.current.scrollIntoView({
-        behavior: 'smooth',
-        block: 'start',
-      })
+    if (isMounted() && isScreenBelowLg) {
+      scrollRefIntoView(tabGroupRef)
     }
   }
 
--- a/src/app/events/[slug]/components/ScheduleSection/Tabs.tsx
+++ b/src/app/events/[slug]/components/ScheduleSection/Tabs.tsx
@@ -42,8 +42,8 @@ export function Tabs({ schedule }: TabsProps) {
   function handleTabIndexChange(index: number) {
     setActiveTabIndex(index)
 
-    if (isMounted() && isScreenBelowLg) {
-      scrollTabContentIntoView(tabGroupRef)
+    if (isMounted() && isScreenBelowLg && tabGroupRef.current) {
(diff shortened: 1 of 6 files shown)


src/app/events/[slug]/components/ScheduleSection/Tabs.tsx
```diff
@@ -18,29 +20,39 @@ type TabsProps = {
...
-  function scrollToTabGroup() {
-    if (isMounted() && isScreenBelowLg && tabGroupRef.current) {
-      tabGroupRef.current.scrollIntoView({
-        behavior: 'smooth',
-        block: 'start',
-      })
+  function handleIndexChange(index: number) {
+    setActiveTabIndex(index)
+
+    if (isMounted() && isScreenBelowLg) {
```
Comment (mirhamasala) [inline src/app/events/[slug]/components/ScheduleSection/Tabs.tsx:45]: Wouldn’t it be better to let `scrollRefIntoView` handle its core functionality without adding optional chaining or guards?

```suggestion
    if (isMounted() && isScreenBelowLg && tabGroupRef.current) {
```

src/app/events/[slug]/components/ScheduleSection/Tabs.tsx
```diff
@@ -18,29 +20,39 @@ type TabsProps = {
...
     `(max-width: ${parseInt(screens.md, 10) - 1}px)`,
   )
 
-  function scrollToTabGroup() {
-    if (isMounted() && isScreenBelowLg && tabGroupRef.current) {
-      tabGroupRef.current.scrollIntoView({
-        behavior: 'smooth',
-        block: 'start',
-      })
+  function handleIndexChange(index: number) {
```
Comment (mirhamasala) [inline src/app/events/[slug]/components/ScheduleSection/Tabs.tsx:42]:
```suggestion
  function handleTabIndexChange(index: number) {
```

src/app/events/[slug]/utils/scrollRefIntoView.ts
```diff
@@ -0,0 +1,3 @@
+export function scrollRefIntoView(ref: React.RefObject<HTMLElement>) {
```
Comment (mirhamasala) [inline src/app/events/[slug]/utils/scrollRefIntoView.ts:1]: It wasn't long ago that we implemented `scrollToTabGroup` and yet both `scrollToTabGroup` and `scrollRefIntoView` made me wonder what this was for. What about `scrollToTabContentStart`? Or something like that?

mirhamasala (commented):
@CharlyMartin - Quite awesome, Nuqs. Just some nits - And one question:

What should happen if a user directly goes to `fil.org/events/fil-bangkok-2024?tab=2`

src/app/events/[slug]/components/ScheduleSection/Tabs.tsx
Comment (CharlyMartin) [reply src/app/events/[slug]/components/ScheduleSection/Tabs.tsx:45]: Yes, I agree with that statement.

However, ref objects are always typed as a value or `null`.

```ts
interface RefObject<T> {
  readonly current: T | null;
}
```

So even if we check that `tabGroupRef.current` exists before calling the function, we still have to use optional chaining in the function body.

But I'm fine with adding `&& tabGroupRef.current`

src/app/events/[slug]/utils/scrollRefIntoView.ts
Comment (CharlyMartin) [reply src/app/events/[slug]/utils/scrollRefIntoView.ts:1]: Sure, that works :)

CharlyMartin:
> What should happen if a user directly goes to fil.org/events/fil-bangkok-2024?tab=2?

Good question. I wonder if we should scroll to the schedule section if a tab is active in the URL.

The platform already provides this behavior "for free" as most browsers keep the scroll position of the previous page when navigating back. So, if someone clicks on "View Details" and then comes back, they end up in the same spot on the page.

Now, if a user were to share `fil.org/events/fil-bangkok-2024?tab=2` with someone, they would land at the top of the page.

But would sharing a URL with `?tab=2` be intentional? Once the tab parameter is in the URL, it stays there, even if users scroll back up the page.

What do you think?

src/app/events/[slug]/components/ScheduleSection/Tabs.tsx
Comment (mirhamasala) [reply src/app/events/[slug]/components/ScheduleSection/Tabs.tsx:45]: We do? Can't we use the NonNullable magic?

Comment (CharlyMartin) [reply src/app/events/[slug]/components/ScheduleSection/Tabs.tsx:45]: I suppose we can do something like that if we reach into the `ref` and mark `current` as non-nullable.

```ts
export function scrollTabContentIntoView(
  current: NonNullable<React.RefObject<HTMLElement>['current']>,
) {
  current.scrollIntoView({ behavior: 'smooth', block: 'start' })
}
```

And then

```ts
if (isMounted() && isScreenBelowLg && tabGroupRef.current) {
  scrollTabContentIntoView(tabGroupRef.current)
}
```

What do you think about that? I'm not used to doing it but it works :)

CharlyMartin:
> It might not be intentional. Let's keep it as is for now?

Let's!

And thanks for sharing the SEO tip :)

src/app/events/[slug]/components/ScheduleSection/Tabs.tsx
Comment (mirhamasala) [reply src/app/events/[slug]/components/ScheduleSection/Tabs.tsx:45]: Works for me! LGTM!

mirhamasala (approved):
Love it! Thanks, Charly.

PR #787, opened 2024-11-04
[UXIT-1710] Fix speaker images on events page

CharlyMartin:
## 📝 Description

This PR fixes the speaker images on the FIL Bangkok event page. Currently not all images have the same width on mobile devices.

## 🛠️ Key Changes

- Reworks `KeyMemberCard` to treat dynamic and static images differently

## 🧪 How to Test

- Inspect the About and FIL Bangkok event page on a mobile device and make sure image sizes are consistent


mirhamasala (commented):
@CharlyMartin I think when we push a "fix PR," it would be nice to include a sentence about what was wrong in the description. Pushed a CR - still testing. 🙏🏼

PR #801, opened 2024-11-06
[UXIT-1666] - Zod Validation for Page Attributes

CharlyMartin:
> [!NOTE]
> To keep this PR small, I only tackled the home, about, and blog pages as examples. If you like this new set-up, I'll add the rest of the pages to this PR or a open second one.

## 📝 Description

This PR adds zod validation to data coming from the page Markdown entries, located in `src/content/pages`. Instead of importing the file, we now call `getPageMarkdownData` to fetch and validate the page information by passing the relevant zod parser.

```ts
// Before
import { attributes } from '@/content/pages/home.md'
attributes.header

// After
const homePageData = getPageMarkdownData({
  slug: 'home',
  zodParser: HomePageDataSchema.parse,
})
homePageData.header
```

## 🛠️ Key Changes

1. Creates `getPageMarkdownData`, which is a thin wrapper around `getMarkdownData`
2. Creates 3 types of page data schemas: `GenericPageDataSchema` and `PageDataWithFeaturedEntrySchema`, which are generic, and `HomePageDataSchema`, which is a custom one just for the home page.

## 🧪 How to Test

- Make sure the pages affected by the changes have the same information as before.


src/app/about/page.tsx
```diff
@@ -26,11 +27,14 @@ import { focusAreasData } from './data/focusAreasData'
...
-const { header, seo } = attributes
+const aboutPageData = getPageMarkdownData({
+  slug: 'about',
+  zodParser: GenericPageDataSchema.parse,
+})
 
 export const metadata = createMetadata({
   seo: {
-    ...seo,
+    ...aboutPageData.seo,
```
barbaraperic, line 37: It would be nice to extract it? `const { seo, header } = aboutPageData`

src/app/_schemas/PageDataSchema.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const PageDataWithFeaturedEntrySchema = GenericPageDataSchema.extend({
+  featured_entry: MarkdownEntryPathSchema,
+})
+
+export const HomePageDataSchema = GenericPageDataSchema.extend({
+  featured_ecosystem_projects: z.array(MarkdownEntryPathSchema),
+})
+
+function validateMarkdownEntryPathFormat(value: string) {
```
barbaraperic, line 34: why do we need validation?

src/app/about/page.tsx
```diff
@@ -26,11 +27,14 @@ import { focusAreasData } from './data/focusAreasData'
 import { reportsData } from './data/reportsData'
 import { generateStructuredData } from './utils/generateStructuredData'
 
-const { header, seo } = attributes
+const aboutPageData = getPageMarkdownData({
```
barbaraperic, line 30: So the `aboutPageData` type has `header` and `seo`, but also the `slug`? Why do we need `slug`? I might be missing something
Screenshot 2024-11-07 at 12 33 01

src/app/about/page.tsx
```diff
@@ -26,11 +27,14 @@ import { focusAreasData } from './data/focusAreasData'
...
+  slug: 'about',
+  zodParser: GenericPageDataSchema.parse,
+})
```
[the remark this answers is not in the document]
CharlyMartin: Good thinking.

`PATHS.ABOUT.path` gives us `/about`. We want `about`, which is the file's name. I'm not sure we can rely entirely on `PATHS` here because the URL paths don't match the file paths 1:1:
- `home.md` is `/`
- `maturity-model.ms` is `/security/maturity-model`

src/app/_schemas/PageDataSchema.ts
CharlyMartin: If a a file path is broken, it would be to catch it, no? `z.string` is a bit weak.

src/app/about/page.tsx
CharlyMartin: We don't need `slug`, but because `getPageMarkdownData` uses `getMarkdownData` internally, it gets added to the response. I'm fine with it.

Would you prefer `getPageMarkdownData` to implement its own logic? It will look 95% similar to `getMarkdownData`, but maybe it's better this way? What do you think?

src/app/about/page.tsx
CharlyMartin: Usually I'm all for it. But in this case I think it's best not to, because of conflict variable names.

On the home page, for instance, `homePageData.featuredEcosystemProjects` would conflict with the already declared `featuredEcosystemProjects`.

So, to be consistent, I didn't destructure any objects.

src/app/about/page.tsx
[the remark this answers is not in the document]
mirhamasala: I do think @barbaraperic has a point here. I'm still figuring out this PR, but why don't we have something like:


```
function getSlugFromPath(path: PathValues) {
  const pathConfig = Object.values(PATHS).find((config) => config.path === path)
  if (!pathConfig) {
    throw new Error(`Path config not found for path: ${path}`)
  }

  return (pathConfig.mainContentPath ?? path).replace(
    `${CONTENT_PAGES_ROOT}`,
    '',
  )
}

export function getPageMarkdownData<T>({ slug, zodParser }: GetPageData<T>) {
  const mainContentPath = getSlugFromPath(slug as PathValues)
  return getMarkdownData({
    slug: mainContentPath,
    directoryPath: CONTENT_PAGES_ROOT,
    zodParser,
  })
}
```

This was we can do:

```
const aboutPageData = getPageMarkdownData({
  slug: PATHS.ABOUT.path,
  zodParser: GenericPageDataSchema.parse,
})
```

src/app/about/page.tsx
mirhamasala: I agree with @barbaraperic - Most of the times, destructuring will work. And it reads much better. (If we really wanted, we could change things like `homePageData.featuredEcosystemProjects` to be `featuredEcosystemProjectSlugs` because that is what they actually are.)

src/app/_schemas/PageDataSchema.ts
mirhamasala: I think @barbaraperic raises a good point here. This current validation might give us a false sense of security. Even if we extend it to something like this:

```typescript
function validateEntryPath(customEntriesContentPath: string) {
  return (value: string) => {
    return value.startsWith(`${CONTENT_ROOT}${customEntriesContentPath}`) && 
           value.endsWith('.md');
  };
}

function createEntryPathSchema(customEntriesContentPath: string) {
  return z.string().refine(
    validateEntryPath(customEntriesContentPath),
    {
      message: `Path must be in ${CONTENT_ROOT}${customEntriesContentPath} and end with .md`,
    }
  );
}

export const EcosystemProjectPathSchema = createEntryPathSchema(
  PATHS.ECOSYSTEM_EXPLORER.customEntriesContentPath!
);

export const PageDataWithFeaturedEntrySchema = GenericPageDataSchema.extend({
  featured_ecosystem_projects: z.array(EcosystemProjectPathSchema),
});

However, there are still some limitations:

1. **Only validates path structure**: It checks the path pattern but doesn’t confirm if the file actually exists.
2. **Misses typos**: It won’t catch filename typos (e.g., `"nuklai.md"` vs. `"nuklei.md"`).
3. **Limited validation**: It’s essentially just verifying string prefixes and suffixes, which doesn’t add much value over a simple `z.string()`.

For instance, both of these would pass validation:

- `"src/content/ecosystem-explorer/projects/nuklai.md"` (actual file)
- `"src/content/ecosystem-explorer/projects/not-real.md"` (typo/non-existent)

### Given that:

- The path structure is already enforced by our directory structure.
- We’re not verifying if the file actually exists.
- Most typos would be caught at runtime when trying to read the file.
- This validation adds complexity without providing meaningful safety.

### Would it make sense to:

- Simplify this to just using `z.string()`?
- Be more transparent about what we’re actually validating?
- Keep it simple since it would be just as effective in practice?

What do you think?

File: src/app/about/page.tsx (hunk 3 above)
Comment (mirhamasala) [reply src/app/about/page.tsx:30]: I would probably say yes to `getPageMarkdownData` implementing its own logic.

**Review by mirhamasala — CHANGES_REQUESTED** [review]
@charly - I think you took an interesting approach here. @barbaraperic raised some great questions—many of which I didn’t even realize I had, but they articulated what felt off in the code.

I’ve left a few comments and proposals. Please take a look and let me know your thoughts. Once those are addressed and we have a clearer path forward, I'll do a second review. We’ll likely need to rename some functions after the restructuring, but that should be simpler once the implementation is more solid.

Thank you!

File: src/app/about/page.tsx (hunk 4 above)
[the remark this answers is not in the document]
Comment (CharlyMartin) [reply src/app/about/page.tsx]: ✅

File: src/app/about/page.tsx (hunk 1 above)
Comment (CharlyMartin) [reply src/app/about/page.tsx]: ✅

File: src/app/_schemas/PageDataSchema.ts (hunk 2 above)
Comment (CharlyMartin) [reply src/app/_schemas/PageDataSchema.ts:39]: What do you think about this? Now, it does verify that the file exists.

https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/801/commits/5c5fe5fdbf46addc59c9559f2b9784cc80566956#diff-c125a91e0f476ae7c84c450daa9050680bf46190395d0d8a17e9332b7fcdaca6R22-R24

If you don't like it, `z.string()` is an easy fallback.

File: src/app/about/page.tsx (hunk 4 above)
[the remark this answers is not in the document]
Comment (CharlyMartin) [reply src/app/about/page.tsx]: `getPageMarkdownData` now accepts `PATHS.ABOUT` + a Zod parser function.

File: src/app/about/page.tsx (hunk 3 above)
Comment (CharlyMartin) [reply src/app/about/page.tsx]: `getPageMarkdownData` implement its own logic now!

File: src/app/about/page.tsx [hunk 5]
```diff
@@ -26,11 +27,14 @@ import { focusAreasData } from './data/focusAreasData'
 import { reportsData } from './data/reportsData'
 import { generateStructuredData } from './utils/generateStructuredData'
 
-const { header, seo } = attributes
+const aboutPageData = getPageMarkdownData({
+  slug: 'about',
+  zodParser: GenericPageDataSchema.parse,
```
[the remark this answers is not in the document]
CharlyMartin: I tried but TS didn't like it

src/app/_schemas/PageDataSchema.ts
mirhamasala: Love it.

CharlyMartin:
> Also, there’s one area we haven’t discussed yet—are there any potential drawbacks to reading from the filesystem instead of using direct imports?

That's a great question @mirhamasala!

GPT tells me that "_Direct imports are typically faster since they’re cached by Node.js, avoiding repeated I/O operations. Reading from the filesystem involves accessing the disk every time, which is slower and can impact performance in high-demand applications._"

Given that we have caching, it may not be a problem?

mirhamasala:
@CharlyMartin ~ In this case, you've written twice the same. But yes, the filenames still need to be renamed.

>
Also, we should rename getPageMarkdownData.ts to getPageMarkdownData.ts no?

Also...

> Given that we have caching, it may not be a problem?

Do we?

CharlyMartin:
@mirhamasala This is what `export const revalidate = 86400`, no?

src/app/_schemas/PageDataSchema.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const HomePageFrontmatterSchema = BaseFrontmatterSchema.extend({
+  featured_ecosystem_projects: z.array(MarkdownPathSchema),
+})
+
+export const GrantsPageFrontmatterSchema = BaseFrontmatterSchema.extend({
+  featured_grant_graduates: z.array(MarkdownPathSchema),
+})
+
+function isValidMarkdownPath(path: string): boolean {
```
mirhamasala, line 43: Shall we move this to `utils`?

mirhamasala:
> This is what export const revalidate = 86400, no?

@CharlyMartin Based on the latest message from Vercel support, I'm not sure..

```
It appears that you are trying to cache the results from a function, in which case setting the revalidation within the layout file won't actually control this, as that is intended for caching pages.

In this case, you will want to look at caching your fetches: https://nextjs.org/docs/app/building-your-application/data-fetching/fetching-caching-and-revalidating#caching-data

You can debug your fetch caching locally by adding the following to your next.config.js:

module.exports = {
  logging: {
    fetches: {
      fullUrl: true,
    },
  },
}

This will provide additional logging for fetch requests, helping you understand which requests are cached or uncached. Additionally, you will likely need to set up logic to get cached data when necessary. 
```

We have a ticket open to look into this [>>](https://www.notion.so/filecoin/FF-Look-into-caching-fetches-11f7631f282580b9a156cef97a025155?pvs=4). Since we haven't, and we still don't know what's causing the initial trouble, my worry is that this PR might increase the bandwidth and invocations limits.

Nonetheless, I think we can go ahead and merge this PR.

CharlyMartin:
@mirhamasala Thanks a lot for sharing! I thought this whole caching Vercel issue was resolved; my bad 🥲

If we don't have caching, then using `import` may be better indeed!

I don't know how it will impact performance as we don't really measure it, but I will implement the changes to be on the safe side.

And about SEO - yes, both are server-side rendered, so there's no difference.

Thank you 🙏

PR #802, opened 2024-11-06
[UXIT-1704] Refactor Link components

mirhamasala:
## 📝 Description
Refactoring links across components to standardize link handling.

**Type:** Refactor

## 🛠️ Key Changes
- Standardized external link handling using `CustomLink` and `TextLink` components
- Added proper `aria-labels` and `rel` attributes for external links
- Removed redundant `CustomLink` usage where simpler `a` tags suffice
- Fixed missing key prop warning

## 🧪 How to Test
**Steps to Test:**
Navigate site checking checking the links


src/app/_components/CustomLink.tsx
```diff
@@ -25,13 +27,11 @@ export function CustomLink({
...
   }
 
-  const rel = href.startsWith('mailto:') ? undefined : 'noopener noreferrer'
-
   return (
     <a
-      href={href}
       rel={rel}
-      aria-label="Opens external link"
+      title={title || 'External link'}
```
barbaraperic, line 33: As I understand, the role of the `title` is to clarify the purpose of the link, like where the link will send the user who clicks on that link? [Example code from W3C ](https://www.w3.org/TR/WCAG20-TECHS/H33.html)

```
<a href="http://example.com/WORLD/africa/kenya.elephants.ap/index.html" 
   title="Read more about failed elephant evacuation">
   Evacuation Crumbles Under Jumbo load
</a>
```
So not sure if "External link" is helpful / servers the purpose?

barbaraperic: Also seems redundant when we have aria-label?

Screenshot 2024-11-07 at 13 17 58

src/app/_components/TextLink.tsx
```diff
@@ -3,40 +3,33 @@
...
 
-type TextLinkProps = {
-  href: string
-  children: React.ReactNode
-} & React.ComponentProps<'a'>
-
 export const linkBaseStyles =
   'text-brand-300 focus:brand-outline hover:underline'
 
 export function TextLink({
```
barbaraperic, line 14: I find it a bit confusing that we have `TextLink` and `CustomLink` 😕 they seem to be doing similar things?

mirhamasala: They are not - `CustomLink` is just a shallow wrapper. `TextLink` includes styles and the arrow out icon.

src/app/_components/CustomLink.tsx
mirhamasala: The HTML `title` and `aria-label` attributes serve different purposes and it's fine to use both:

- `title` creates a visual tooltip that appears when users hover over the link with their mouse - this helps sighted users understand where the link goes before clicking
- `aria-label` is specifically for screen readers and doesn't have any visual effect - this helps users who can't see the tooltip get the same contextual information.

(For example, for the copy to clipboard link, and social icons, the `title` attribute is quite helpful.`)
But you are right that we shouldn't fallback to `External link` - We can leave it empty if it's not provided.

src/app/_components/TextLink.tsx
barbaraperic: I feel like having one component would make maintenance easier...But I haven't tried consolidating, not sure what effect would that have on the code base.

src/app/_components/Navigation/DesktopNavigation/ExternalLink.tsx
```diff
@@ -13,10 +13,10 @@ export function ExternalLink({
 }: LinkProps) {
   return (
     <a
```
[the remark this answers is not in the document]
mirhamasala: At the moment, I couldn't find a good solution for this.

src/app/_components/TextLink.tsx
mirhamasala: Have you checked the use cases for `CustomLink`? I've expanded my initial comment. Maybe that helps explain it.

mirhamasala: At some point, we might also add `OverlayLink` etc. Which would use `CustomLink` internally.

CharlyMartin (changes requested):
Nice refactor :)

One part that was not super obvious at first is: When do we use `a` over `TextLink`. 

As you said, we know that `href={linkedin}` is external in `KeyMemberCard`, so using an `a` tag is enough. It makes sense.

But we also know that `href={`${FILECOIN_FOUNDATION_GITHUB_PR_URL}/${prNumber}`}` is external in `SuccessMessage`, yet we use `TextLink`.

I understand now it's because we want to have the arrow icon automatically included (right?) but that might be confusing us in the future.

src/app/_components/CustomLink.tsx
CharlyMartin: Nice a11y refresher :)

src/app/_components/TextLink.tsx
barbaraperic: Thanks for clarifying! It order wasn't clear to me at first

mirhamasala: Thank you for keeping asking. It's important that it's clear.

src/app/_components/TextLink.tsx
```diff
@@ -3,40 +3,33 @@
...
-  href,
   className,
   children,
   ...rest
-}: TextLinkProps) {
-  const isExternal = isExternalLink(href)
+}: Omit<CustomLinkProps, 'children'> & { children: React.ReactNode }) {
```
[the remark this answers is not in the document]
mirhamasala: Yes, I agree. Apologies - Refactoring craze mistakes.

src/app/_components/TextLink.tsx
```diff
@@ -3,40 +3,33 @@
...
+  const isExternal = !isInternalLink(rest.href)
```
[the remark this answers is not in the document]
mirhamasala: Refactoring craze mistakes. 🙈

src/app/_components/CustomLink.tsx
```diff
@@ -5,9 +5,9 @@ import type { Route } from 'next'
 import { isInternalLink } from '@/utils/linkUtils'
 
 export type CustomLinkProps = {
-  href: string
+  href: string | Route
   children: React.ReactNode
```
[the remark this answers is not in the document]
mirhamasala: Done

src/app/_components/CustomLink.tsx
```diff
@@ -25,16 +26,8 @@ export function CustomLink({
...
-
   return (
-    <a
-      href={href}
-      rel={rel}
-      aria-label="Opens external link"
-      className={className}
-      {...rest}
-    >
+    <a rel={rel} href={href} className={className} {...rest}>
```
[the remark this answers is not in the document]
mirhamasala: Done.

src/app/_components/TextLink.tsx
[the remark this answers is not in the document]
mirhamasala: Done.

mirhamasala:
> One part that was not super obvious at first is: When do we use a over TextLink.
> As you said, we know that href={linkedin} is external in KeyMemberCard, so using an a tag is enough. It makes sense.
> But we also know that href={${FILECOIN_FOUNDATION_GITHUB_PR_URL}/${prNumber}} is external in SuccessMessage, yet we use TextLink.
> I understand now it's because we want to have the arrow icon automatically included (right?) but that might be confusing us in the future.


This was a great question and it led me to refactor the code into

- `SmartTextLink` - For cases where we don’t know ahead of time if a link is external or internal.
- `ExternalTextLink` - Used internally by SmartTextLink but can also be used directly when we know a link is external.
- `InternalTextLink` - Also used by SmartTextLink internally but can be used directly when we know a link is internal.
- Renamed `CustomLink` to `BaseLink` -  his is for components that may not know whether a link is external or internal but aren’t strictly TextLinks.

src/app/_components/TextLink/TextLink.module.scss
```diff
@@ -0,0 +1,3 @@
+.base {
+  @apply text-brand-300 focus:brand-outline hover:underline;
+}
```
CharlyMartin, line 3: Love the idea of moving the style into separate CSS files!

I'm not sure we need CSS modules here, though.

Since we use `@apply`, we get the Tailwind optimization out of the box, and we don't need to import the CSS in components that need it.

src/app/ecosystem-explorer/project-form/components/EcosystemProjectForm.tsx
```diff
@@ -169,15 +170,9 @@ export function EcosystemProjectForm({
               For best quality, please submit a white logo with a transparent
               background, at least 1000px by 1000px, and under 100KB. You can
               use tools like{' '}
-              <a
-                href="https://squoosh.app/"
-                className="text-brand-300 underline hover:text-brand-400"
-                rel="noopener noreferrer"
-                target="_blank"
```
CharlyMartin, line 176: It would nice to keep `target="_blank"` here. If you click on the link then come back, the form state resets.

CharlyMartin: And I suppose `rel="noopener noreferrer"` too

mirhamasala: Yes! Good catch.

src/app/_components/TextLink/TextLink.module.scss
mirhamasala: I'm not entirely sure what the suggestion is, to move them to a regular CSS file?

If we do that, we lose the scoping benefits of CSS Modules. I also think that React/Next.js best practices usually favor scoped styling solutions.

Add them to `tailwind.config.js`?

We can talk about this a bit more, but moving forward as I don't think this is a blocker!

PR #806, opened 2024-11-06
[UXIT-1714] Refactor and reuse ListboxOption

barbaraperic:
## 📝 Description

This PR refactors the ListboxOption component and replaces the Headless UI ListboxOption with our custom implementation.

- **Type:** Refactor

- I've used HeadlessUIListboxOptions instead of our custom ListboxOptions - it was easier to see where's the overlap with components. This will be refactored in a separate PR.
- Missing: ecosystem project form components

## 🧪 How to Test

- Check Filter and Sort components in `blog`, `events`, `ecosystem-explorer,` `filcoin-plus/allocators` and `security/maturity-model`


mirhamasala:
@barbaraperic - Is this ready for review? Asking because you tagged me but it's in draft still?

CharlyMartin (commented):
Hey @barbaraperic,

I've reworked the PR to update only `ListboxOption`, as there were some changes related to `ListboxOptions`.

I saw your comment in the description, but we should wrap it before the new sprint, so I rebased and cleaned the PR.

One thing we should rework is how we pass `count` to the listbox, as it's currently a bit messy. But it's outside the scope of this PR.

src/app/_components/ListboxOption.tsx
```diff
@@ -1,42 +1,41 @@
...
-      <span className="text-nowrap">{option.name}</span>
-      {counts && (
-        <span className="text-sm font-light">({counts[option.id]})</span>
+    <HeadlessUIListboxOption
+      as="li"
+      value={option.id}
+      disabled={option.disabled}
+      className={clsx(
+        'group flex cursor-default items-center justify-between gap-12 px-5 py-2 ui-active:bg-brand-500',
+        option.disabled && 'cursor-not-allowed',
```
CharlyMartin, line 26: Replace by `data-[disabled]:cursor-not-allowed`

CharlyMartin:
@barbaraperic I've reworked the `ListboxOption` component as per the [V2 HeadlessUI docs](https://headlessui.com/react/listbox#basic-example), meaning we pass the whole `option` object to the component instead of just the `id`.

```ts
<ListboxOption key={option.name} value={option}>
```

We can have a final look together tomorrow!

CharlyMartin (approved):
Live reviewed with @barbaraperic

PR #818, opened 2024-11-11
UXIT-1713 - Fix ListboxButton responsive behavior for compactBelow prop

mirhamasala:
This PR fixes the responsive behavior (#794) of the `ListboxButton` component and improves its type naming to be more semantic.

### Bug Fix
Fixed incorrect responsive behavior when using `compactBelow` prop:

**Before:**
When `compactBelow="sm"`:
- Below sm: Incorrectly showed full version
- Above sm: Incorrectly showed compact version

**After:**
When `compactBelow="sm"`:
- Below sm: Correctly shows compact version
- Above sm: Correctly shows full version

### Code Improvements
- Renamed `ViewTransition` type to `LayoutVariant` to better reflect its purpose
- Renamed properties to match their semantic meaning:
 - `hide` → `full`: Controls visibility of the full button layout
 - `show` → `compact`: Controls visibility of the compact icon-only layout
- Updated responsive class patterns:
 - full: `hidden {size}:flex`
 - compact: `flex {size}:hidden`

### DIFF
--- a/src/app/_components/ListboxButton.tsx
+++ b/src/app/_components/ListboxButton.tsx
@@ -14,19 +14,19 @@ type ListboxButtonProps = {
   compactBelow?: Breakpoint
 }
 
-type ViewTransition = {
-  hide: string
-  show: string
+type LayoutVariant = {
+  full: string
+  compact: string
 }
 
-type BreakpointStyles = Record<Breakpoint, ViewTransition>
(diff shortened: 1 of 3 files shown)


mirhamasala:
@barbaraperic @CharlyMartin Tagging for visibility. Going to merge this straight away since production is broken. If anything needs changing, we can take it up as a Scout CR in the stil open Listbox PRs.

PR #827, opened 2024-11-13
[UXIT-1755] Refactor and reuse ListboxOptions

barbaraperic:
## 📝 Description

This PR refactors the ListboxOptions component and replaces the Headless UI ListboxOptions with our custom implementation.

Type: Refactor

PR #829, opened 2024-11-14
[UXIT-517] - Event page: add region to config file and populate events [skip percy]

barbaraperic:
## 📝 Description

This is part 1 of the [Event Page Implement Regions Select](https://www.notion.so/filecoin/FF-Event-Page-Implement-Regions-Select-43169a31c65947b3b0dd0e56d56a0fae?pvs=4) where we add region field to config file and populate the events with the field (**except the Virtual events**).

I will wait to implement part 2 when https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/827 and https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/806 get merged.

This PR also includes a slight refactor in the utils files.

Looks like a lot of changes, but that's because of the updates on event pages!

- **Type:** New feature

## 🛠️ Key Changes

- new config field
- new `cmsFieldUtils` file

## 🧪 How to Test

- Check if location and region match


src/app/_utils/cmsFieldUtils.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+/**
+ * Retrieves and formats field settings for a specified collection and field.
+ *
+ * @param {CMSCollectionName} collectionName - The name of the CMS collection.
+ * @param {string} fieldName - The name of the field within the collection.
+ * @returns {{ options: { id: string; name: string }[]; validIds: string[] }}
+ * An object containing formatted options for the field and a list of valid IDs.
+ */
```
CharlyMartin, line 19: Are we supporting JSDoc now?

src/app/_utils/cmsFieldUtils.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export function getFieldSettings({
+  collectionName,
+  fieldName,
+}: {
+  collectionName: CMSCollectionName
+  fieldName: string
+}) {
```
CharlyMartin, line 27: Can we keep the types outside the function?

src/app/events/schemas/FrontMatterSchema.ts
```diff
@@ -9,15 +9,24 @@ import { ScheduleSchema } from './ScheduleSchema'
 import { SpeakersSchema } from './SpeakerSchema'
 import { SponsorsSchema } from './SponsorSchema'
 
-const { validCategoryIds } = getEventsCategorySettings()
+const { validIds: validCategoryIds } = getFieldSettings({
+  collectionName: 'event_entries',
+  fieldName: 'category',
+})
```
CharlyMartin, line 15: Why don't we use `getCategorySettings` here?

src/content/events/the-crypties.md
```diff
@@ -5,6 +5,7 @@ updated-on: 2023-06-27T17:24:48.635Z
 published-on: 2023-06-27T17:24:48.666Z
 category: supported-sponsored
 location:  Miami, Florida
```
mirhamasala, line 7: Scout nit - To stay consistent with the rest of the data.

```suggestion
location:  Miami, FL
```

src/app/events/page.tsx
```diff
@@ -56,7 +54,8 @@ type Props = {
 }
 
 const events = getEventsData()
-const { categoryOptions, validCategoryIds } = getEventsCategorySettings()
+const { options: categoryOptions, validIds: validCategoryIds } =
+  getFieldSettings({ collectionName: 'event_entries', fieldName: 'category' })
```
CharlyMartin, line 58: Same question here. Couldn't we do this?

```ts
const { options: categoryOptions, validIds: validCategoryIds } =
  getCategorySettings({ collectionName: 'event_entries' })
```

CharlyMartin (commented):
@barbaraperic I think I like this refactor of the CMS utils 🙂

The function names usually confuse me, but this version feels a bit more intuitive.

I'm almost wondering if we could get rid of `getCategorySettings` and use `getFieldSettings` instead. It's just an idea! If you agree, then some of my comments are irrelevant.

public/admin/config.yml
mirhamasala, line 1: Should `location` and `region` be merged? Something like the following:

```
- name: "location"
  label: "Location"
  widget: "object"
  fields:
    - name: "primary"
      label: "City, State/Country"
      widget: "string"
    - name: "region"
      label: "Region"
      widget: "select"
      options:
        - label: "Australia/Oceania"
          value: "australia-oceania"
```

public/admin/config.yml
```diff
@@ -663,6 +663,25 @@ collections:
...
+          - label: "South America"
+            value: "south-america"
+          - label: "Europe"
+            value: "europe"
+          - label: "APAC/Asia"
+            value: "apac-asia"
+          - label: "Africa"
+            value: "africa"
+          - label: "Middle East"
+            value: "middle-east"
```
mirhamasala, line 683: Alphabetize, please.

```suggestion
          - label: "Africa"
           value: "africa"
          - label: "APAC/Asia" 
           value: "apac-asia"
          - label: "Australia/Oceania"
           value: "australia-oceania"
          - label: "Europe"
           value: "europe"
          - label: "Middle East"
           value: "middle-east"
          - label: "North America"
           value: "north-america"
          - label: "South America"
           value: "south-america"
```

src/app/_utils/cmsFieldUtils.ts
mirhamasala, line 19: We avoid comments in code for two reasons: they often become outdated during maintenance, and well-written functions should be self-documenting through explicit naming and structure.

src/app/_utils/cmsFieldUtils.ts
mirhamasala, line 1:
```
import { type CollectionName } from '@/types/cms'
import { type FieldOption, type ComponentOption } from '@/types/options'
import { getCMSFieldOptions, getCollectionConfig } from '@/utils/cmsConfigUtils'

type FieldParams = {
 collectionName: CollectionName
 fieldName: string
}

export function getFieldOptionsAndValidIds({ collectionName, fieldName }: FieldParams) {
 const { fields } = getCollectionConfig(collectionName)
 const cmsOptions = getCMSFieldOptions(fields, fieldName)
 const options = mapCMSOptionsToComponentFormat(cmsOptions)
 const validIds = options.map((option) => option.id)

 return { options, validIds }
}

function mapCMSOptionsToComponentFormat(options: Array<FieldOption>) {
 return options.map((option) => ({
   id: option.value,
   name: option.label,
 }))
}
````

File: src/app/_utils/categoryUtils.ts [hunk 7]
```diff
@@ -11,18 +12,6 @@ type GetCategoryLabelParams = {
...
-  return options.map((option) => ({
-    id: option.value,
-    name: option.label,
-  }))
-}
-
-function getCategoryData(collectionName: CMSCollectionName) {
-  const { fields } = getCollectionConfig(collectionName)
-  return getCMSFieldOptions(fields, CATEGORY_FIELD_NAME)
-}
```
mirhamasala, line 24: Curious: Why did we move this out?

src/app/events/page.tsx
mirhamasala: @CharlyMartin - I'm guessing because of `fieldName` - But we could rework `getCategorySettings` to accept `fieldName` ..

src/app/_utils/cmsFieldUtils.ts
mirhamasala: [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/829/files#r1846837186)

src/app/events/schemas/FrontMatterSchema.ts
```diff
@@ -9,15 +9,24 @@ import { ScheduleSchema } from './ScheduleSchema'
...
 
 const CategorySchema = createEnumSchema(validCategoryIds)
+const RegionSchema = createEnumSchema(validRegionIds)
 
 export const EventFrontMatterSchema = DynamicBaseDataSchema.extend({
   title: z.string(),
   category: CategorySchema,
   description: z.string().optional(),
   location: z.string(),
+  region: RegionSchema.optional(),
```
mirhamasala, line 29: @CharlyMartin - What do you think about this? This is now `optional` because an event can be virtual.

src/content/events/fil-dev-summit-singapore.md
mirhamasala, line 1: Says `singapore` in the slug but is FIL Dev Summit Iceland - [Slack Thread >>](https://filecoinfoundation.slack.com/archives/C06MTEQ3SP6/p1731946315659299). I'll open a ticket to fix with redirect. 🙏🏼

mirhamasala (changes requested):
@barbaraperic - Very excited about this one! 🤩 Please have a look at the comments and we can discuss some the suggestions if needed. 🙏🏼

src/app/events/page.tsx
CharlyMartin: I see!

I agree that `getCategorySettings` is likely obsolete now. We can use `getFieldSettings` instead; I find this more intuitive.

Regarding the renaming, I would go with `getCMSFieldOptions` and have that function return only an array of options.

We can create a simple util function to extract ids from the list of options when we need it, such as:

```ts
function getIdsFromOptions<O extends OptionType>(options: Array<O>) {
  return options.map((option) => option.id)
}

const validIds = getIdsFromOptions(options)
```

What do you think?

src/content/events/fil-dev-summit-singapore.md
mirhamasala: Fixed in https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/837

src/app/events/schemas/FrontMatterSchema.ts
CharlyMartin: I think we can enforce `region` to be present if location is not `"Virtual"`, let me have a look.

CharlyMartin: What about something like this?

```ts
const BaseEventFrontMatterSchema = DynamicBaseDataSchema.extend({
  title: z.string(),
  category: CategorySchema,
  description: z.string().optional(),
  'external-link': z
    .object({
      url: z.string().url(),
      text: z.string().optional(),
    })
    .optional(),
  'luma-calendar-link': z.string().url().optional(),
  'luma-events-section': z
    .object({
      title: z.string().optional(),
      'embed-link': z.string().url(),
    })
    .optional(),
  'start-date': z.coerce.date(),
  'end-date': z.coerce.date().optional(),
  schedule: ScheduleSchema.optional(),
  speakers: SpeakersSchema.optional(),
  sponsors: SponsorsSchema.optional(),
  'recap-youtube-embed-url': z.string().url().optional(),
  'recap-youtube-playlist-url': z.string().url().optional(),
}).strict()

const VirtualEventSchema = BaseEventFrontMatterSchema.extend({
  location: z.literal('Virtual'),
  region: z.void(),
}).strict()

const InPersonEventSchema = BaseEventFrontMatterSchema.extend({
  location: z.string(),
  region: RegionSchema,
}).strict()

export const EventFrontMatterSchema = z.union([
  VirtualEventSchema,
  InPersonEventSchema,
])
```

barbaraperic: It doesn't throw an error if the `primary` location is "Virtual" and the user fills out the `region`... Maybe we try with [refine](https://zod.dev/?id=refine)?

barbaraperic:
```ts
const LocationSchema = z
  .object({
    primary: z.string(),
    region: RegionSchema.optional(),
  })
  .refine((data) => !(data.primary === 'Virtual' && data.region), {
    message: 'Region must be empty if the event is virtual.',
    path: ['region'],
  })
  .refine((data) => !(data.primary !== 'Virtual' && !data.region), {
    message: 'Region must be filled out if the event is not virtual.',
    path: ['region'],
  })
```

PR #831, opened 2024-11-14
[UXIT-1100] - Include Nested Checkbox for Ecosystem Subcategories [skip percy]

CharlyMartin:
> [!NOTE]
> The high number of file changes is mostly content files. Only 14 file changes are code-related.

## 📝 Description

This PR adds subcategory filters to the ecosystem page on mobile and desktop.

The first two commits update the `FormCheckbox` and `SlideOver` components to adapt them for this feature:
- 10f0ce9945852fe1be85772cd19f7674f1f136cd (1 file changed)
- 6c0a508255865105aa5e132f25bd53cee542990b (1 file changed)

The third commit is where the implementation of the feature is:
- 44c2e7df7f8730605f479288e29c24c5d48e720e (12 files changed)

The fourth and last commit cleans the ecosystem project Markdown entries. There were many mismatches between the subcategories and their parent category, so I manually updated them. We might want to enforce that at the Zod level going forward, if feasible.
- 16e7d170d883a52c6647e0a79969ee303a9810bb (35 files changed)

## 🛠️ Key Changes
- Created `getCategoriesFromDirectory` and `getSubcategoriesFromDirectory` to fetch the list of categories and subcategories from the CMS config.
- Used `Zod` to validate the data coming from the CMS config, as we are increasingly going in this direction
- Added `SubcategoryFilters` with `nuqs` to sync the selected checkboxes with the URL.
- Created `useEcosystemCategory`, which replaces `useCategory` on the ecosystem explorer page

> [!IMPORTANT]
> The vertical spacing between checkboxes is bigger than on Figma on smaller devices to account for touch targets. @filipagr approved the design with me via DM.  

## 🧪 How to Test

1. Go to `/ecosystem-explorer`
2. Apply some filters and make sure the Results total and selected projects match the selection

## 📸 Screenshots

CleanShot 2024-11-19 at 09 31 47

## 🔖 Resources

- https://nuqs.47ng.com/docs/parsers/built-in#arrays


mirhamasala:
@CharlyMartin - We need @matthendrian to review https://github.com/FilecoinFoundationWeb/filecoin-foundation/commit/16e7d170d883a52c6647e0a79969ee303a9810bb..

barbaraperic:
Started testing the UI — it looks great, and the filters work well! 👌 One thing I noticed: if I click on a filter checkbox and then select a list item, navigating back resets the filter. I would expect the filter to remain checked instead. Is this intentional or something to adjust?

src/app/ecosystem-explorer/components/SubcategoryFilters.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const [subcategory, setSubcategory] = useQueryState(
+    CATEGORY_KEY,
+    parseAsArrayOf(parseAsString, CATEGORY_QUERY_SEPARATOR_SYMBOL)
+      .withDefault([])
+      .withOptions({ shallow: false }),
+  )
+
+  return (
+    <div>
+      <h3 className="pb-7 text-sm font-bold text-blue-300 lg:pb-4">
```
barbaraperic, line 27: Should it be `text-brand-300`?

src/app/_components/Form/FormCheckbox.tsx
```diff
@@ -1,26 +1,57 @@
...
 
 export type FormCheckboxProps = {
   label: string
+  labelSize?: 'sm' | 'md'
 } & Omit<CheckboxProps, ExcludedHeadlessUIProps>
 
-export function FormCheckbox({ label, ...rest }: FormCheckboxProps) {
+const TOUCH_TARGET = {
+  base: 'size-5',
+  augmentingClass: 'p-3.5',
```
barbaraperic, line 16: For me the key name are not easy to understand 😕. `augumentingClass` is the padding to create larger clickable area right? Can we say something like `clickArea`? and `clickAreaOffset`? Something descriptive... Should we also make it consistent with other TOUCH_TARGET (maybe in a separate PR tho)?

src/app/ecosystem-explorer/components/SubcategorySlider.tsx
```diff
@@ -0,0 +1,82 @@
...
+import type { CategoriesAndSubcategoriesWithCount } from '../types/ecosystemCategoryType'
+
+import { SliderCloseButton } from './SliderCloseButton'
+import { SubcategoriesSidebar } from './SubcategoriesSidebar'
+
+type Props = {
+  categories: CategoriesAndSubcategoriesWithCount
+}
+
+export function SubcategorySlider({ categories }: Props) {
```
barbaraperic, line 23: Looks really good on mobile 😊

src/app/ecosystem-explorer/components/SubcategorySlider.tsx
```diff
@@ -0,0 +1,82 @@
...
+            <Button
+              className="basis-1/2"
+              variant="ghost"
+              onClick={resetParamsAndCloseSlider}
+            >
+              Clear all
+            </Button>
+
+            <Button className="basis-1/2" onClick={closeSlider}>
+              See results
```
barbaraperic, line 62: On Figma it says "Apply Filters"?

src/app/ecosystem-explorer/components/SubcategorySlider.tsx
```diff
@@ -0,0 +1,82 @@
...
+        </div>
+
+        <div className="sticky bottom-0 z-10 border-t border-brand-300 bg-brand-800 p-6">
+          <div className="flex items-center gap-3">
+            <Button
+              className="basis-1/2"
+              variant="ghost"
+              onClick={resetParamsAndCloseSlider}
+            >
+              Clear all
```
barbaraperic, line 58: On Figma it says "Clear All"?

src/app/ecosystem-explorer/components/SubcategorySlider.tsx
```diff
@@ -0,0 +1,82 @@
...
+    closeSlider()
+  }
+
+  function openSlider() {
+    setOpen(true)
+  }
+
+  function closeSlider() {
+    setOpen(false)
+  }
```
barbaraperic, line 81: A personal preference - I find it easier to read if the functions are before the return statement 😀 That said, I’m happy to follow the existing pattern if both of you ok with keeping them below.

src/app/ecosystem-explorer/utils/getCategoriesFromDirectory.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export function getCategoriesFromDirectory() {
+  const rawCategories = readAndValidateYamlFiles(
+    ECOSYSTEM_CATEGORIES_DIRECTORY_PATH,
+  )
+
+  const categories = z.array(DirectoryCategorySchema).parse(rawCategories)
+
+  return categories
+}
```
barbaraperic, line 17: Could `getCategoriesFromDirectory` and `getSubcategoriesFromDirectory` be a single function? There are quite similar, but happy to keep it this way if more readable.

src/app/ecosystem-explorer/components/SubcategoryFilters.tsx
CharlyMartin: Definitely! Good catch :)

src/app/ecosystem-explorer/constants/searchParams.ts
```diff
@@ -0,0 +1 @@
+export const CATEGORY_QUERY_SEPARATOR_SYMBOL = ','
```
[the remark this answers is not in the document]
CharlyMartin: [nuqs](https://github.com/47ng/nuqs/blob/next/packages/nuqs/src/parsers.ts#L379) uses comma separators for arrays of values, so I kept it this way.

I am happy to change to something else; it should work the same with a dash or a plus sign.

src/app/_components/SlideOver.tsx
```diff
@@ -8,23 +8,48 @@ import {
...
+    closedPosition: 'translate-x-full',
+  },
+  left: {
+    startingPosition: 'left-0',
+    openPosition: 'translate-x-0',
+    closedPosition: '-translate-x-full',
+  },
+} as const
+
+const animationStyles = 'transition ease-in-out duration-500 sm:duration-700'
```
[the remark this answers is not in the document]
CharlyMartin: > Curious why different animation speeds on mobile and tablet? For me 700 seems a bit too slow...

This is how we currently have it. I think these values came from [Tailwind UI](https://tailwindui.com/components/application-ui/overlays/drawers).

I'll play around with your suggestion and see :)

src/app/ecosystem-explorer/components/SubcategorySlider.tsx
CharlyMartin: Ah yes! We had a chat about it yesterday during standup.

Because filters are applied as users click checkboxes, `Apply Filters` does not really apply any filters; it just closes the slider. So we decided to rename it to `See results`

I will rename it to `See Results`!

src/app/ecosystem-explorer/components/SubcategorySlider.tsx
CharlyMartin: I think we agreed to put them below to follow the Clean Code JS examples.

When function names are descriptive enough, keeping them below the component makes the file more readable.

For instance, `onClick={openSlider}` is readable enough that you don't need to know about the implementation, in my opinion. You know the callback opens the modal.

(But obviously, in the context of PR review, looking at the implementation is important 😅)

src/app/ecosystem-explorer/utils/getCategoriesFromDirectory.ts
CharlyMartin: I initially combined them in one function, then split that function into two so they each had a single responsibility.

These are single-use functions, so to speak, so maybe one function is fine, yes.

CharlyMartin:
> Started testing the UI — it looks great, and the filters work well! 👌 One thing I noticed: if I click on a filter checkbox and then select a list item, navigating back resets the filter. I would expect the filter to remain checked instead. Is this intentional or something to adjust?

@barbaraperic This is how we currently have it on [fil.org](https://www.fil.org/ecosystem-explorer). It's not related to this implementation, per se, but I agree with you: It's not the best UX.

I'm not sure how we can solve this, as the breadcrumb links are hard-coded; they always navigate to a fresh URL `/ecosystem-explorer,` hence resetting the state.

Using the browser's back button does not, though! This is what I primarily use, so I never ran into the issue.

So, is this a bug or a feature? 😅

barbaraperic:
> > Started testing the UI — it looks great, and the filters work well! 👌 One thing I noticed: if I click on a filter checkbox and then select a list item, navigating back resets the filter. I would expect the filter to remain checked instead. Is this intentional or something to adjust?
> 
> @barbaraperic This is how we currently have it on [fil.org](https://www.fil.org/ecosystem-explorer). It's not related to this implementation, per se, but I agree with you: It's not the best UX.
> 
> I'm not sure how we can solve this, as the breadcrumb links are hard-coded; they always navigate to a fresh URL `/ecosystem-explorer,` hence resetting the state.
> 
> Using the browser's back button does not, though! This is what I primarily use, so I never ran into the issue.
> 
> So, is this a bug or a feature? 😅

Haven't tried, but would local storage work? 🧐

CharlyMartin:
> Haven't tried, but would local storage work? 🧐

@barbaraperic, The issue with `localStorage`, is that we now have two sources of authority: the URL and the browser's storage. If there's a discrepancy, which one should we trust?

src/app/_components/SlideOver.tsx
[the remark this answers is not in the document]
CharlyMartin: ✅

src/app/ecosystem-explorer/components/SliderCloseButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+  closeSlider: () => void
+}
+
+const TOUCH_TARGET = {
+  base: 24,
+  augmentingClass: 'p-3',
+  offsettingClass: '-m-3',
+}
+
+export function SliderCloseButton({ closeSlider }: SliderCloseButtonProps) {
```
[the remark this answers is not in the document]
mirhamasala: I think @barbaraperic makes a valid point here.

Our `ToggleButton` doesn't actually toggle states. And our `SliderCloseButton` should have the exact same styles.

src/app/ecosystem-explorer/components/SubcategoriesSidebar.tsx
```diff
@@ -0,0 +1,21 @@
+import type { CategoriesAndSubcategoriesWithCount } from '../types/ecosystemCategoryType'
+
+import { SubcategoryFilters } from './SubcategoryFilters'
+
+type Props = {
```
mirhamasala, line 5: We do simple `Props` now? :-)

mirhamasala:
> @barbaraperic, The issue with localStorage - and any form of browser storage - is that we would have two sources of authority: the URL and the storage. If there's a discrepancy between the two, which one should we trust?

@barbaraperic @CharlyMartin ~ I don't think `localStorage` is the way to go for reasons Charly mentioned. I do think this is good catch, and we can open a ticket to enhance the breadcrumb component to preserve URL parameters when navigating, making it consistent with other navigation methods. Not a blocker for getting this PR merged. Thank you, both. 🙏🏼

mirhamasala:
@CharlyMartin - There are some console errors.

src/app/_components/Form/FormCheckbox.tsx
```diff
@@ -1,26 +1,57 @@
...
   label: string
+  labelSize?: 'sm' | 'md'
 } & Omit<CheckboxProps, ExcludedHeadlessUIProps>
 
-export function FormCheckbox({ label, ...rest }: FormCheckboxProps) {
+const TOUCH_TARGET = {
+  base: 'size-5',
+  clickArea: 'p-3.5',
+  clickAreaOffset: '-m-3.5',
+}
```
mirhamasala, line 18: @barbaraperic @CharlyMartin - We need to tackle this naming once and for all 🤣 I don't like these much either.

What about:

```
  visibleElementSize: 'size-5',      
  touchAreaPadding: 'p-3.5',    
  touchAreaOffset: '-m-3.5',
```  
 
 ```
  elementVisibleSize: 'size-5',  
  additionalTouchPadding: 'p-3.5', 
  alignmentOffset: '-m-3.5',       

```

```
  displaySize: 'size-5',    
  expandTouchArea: 'p-3.5',       
  adjustPositionForPadding: '-m-3.5', 
```

File: src/app/_components/Form/FormCheckbox.tsx [hunk 13]
```diff
@@ -1,26 +1,57 @@
...
+  labelSize = 'md',
+  ...rest
+}: FormCheckboxProps) {
   return (
     <Field className="flex items-center gap-3">
       <Checkbox
         {...rest}
-        className="group size-5 rounded bg-brand-100 p-0.5 text-brand-100 focus:brand-outline data-[disabled]:cursor-not-allowed data-[checked]:bg-brand-400"
+        className={clsx(
+          'group cursor-pointer focus:outline-none',
```
mirhamasala, line 30: Let's work on the focus styles.

CharlyMartin: We should have a medium blue focus style for white background elements.

src/app/ecosystem-explorer/components/SubcategoryFilters.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { useQueryState, parseAsArrayOf, parseAsString } from 'nuqs'
+
+import { CATEGORY_KEY } from '@/constants/searchParams'
+
+import { FormCheckbox } from '@/components/Form/FormCheckbox'
+
+import { CATEGORY_QUERY_SEPARATOR_SYMBOL } from '../constants/searchParams'
+import type { SubcategoriesWithCount } from '../types/ecosystemCategoryType'
+
+type Props = {
```
mirhamasala, line 12: 🎅🏿

src/app/ecosystem-explorer/page.tsx
```diff
@@ -163,13 +160,16 @@ export default function EcosystemExplorer({ searchParams }: Props) {
                         project
 
                       const isFirstTwoImages = i < 2
+                      const categoryName = categories.find(
```
mirhamasala, line 163: `findOrThrow

src/app/ecosystem-explorer/utils/getCategoriesFromDirectory.ts
mirhamasala: Yes - Let's have a single function.

src/app/ecosystem-explorer/utils/getSubcategoriesFromQueryParam.ts
```diff
@@ -0,0 +1,16 @@
+import type { NextServerSearchParams } from '@/types/searchParams'
+
+import { CATEGORY_KEY } from '@/constants/searchParams'
+
+import { CATEGORY_QUERY_SEPARATOR_SYMBOL } from '../constants/searchParams'
+
+type CategoryParams = NextServerSearchParams[typeof CATEGORY_KEY]
+
+export function getSubcategoriesFromQueryParam(categoryParams: CategoryParams) {
```
mirhamasala, line 9: `parseCategoryIdsFromQuery or `parseCategoryQueryParam`

src/app/ecosystem-explorer/components/SliderCloseButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <Button
+      aria-label="Close filters"
```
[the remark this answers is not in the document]
CharlyMartin: ✅

src/app/ecosystem-explorer/components/SliderCloseButton.tsx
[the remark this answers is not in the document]
CharlyMartin: ✅

src/app/ecosystem-explorer/utils/getCategoriesFromDirectory.ts
CharlyMartin: ✅

src/app/ecosystem-explorer/page.tsx
CharlyMartin: ✅

src/app/ecosystem-explorer/components/SubcategoryFilters.tsx
CharlyMartin: ✅

src/app/ecosystem-explorer/components/SubcategoriesSidebar.tsx
CharlyMartin: ✅

src/app/ecosystem-explorer/utils/getSubcategoriesFromQueryParam.ts
CharlyMartin: ✅

src/app/_components/Form/FormCheckbox.tsx
CharlyMartin: ✅

src/app/_components/Form/FormCheckbox.tsx
CharlyMartin: ✅

CharlyMartin:
@mirhamasala I think the console errors we see happen because the labels are paired with span elements (what `Checkbox` renders), and not with an input-related element.

As per the MDN [docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label)
```
Elements that can be associated with a <label> element include <button>, <input> (except for type="hidden"), <meter>, <output>, <progress>, <select> and <textarea>. 
```

I tried doing `<Checkbox as="input" />`, but it creates an error because inputs are self-contained elements.

PR #849, opened 2024-11-21
[UXIT-517] Implement location filtering on Events Page 

barbaraperic:
This PR is currently being split in multiple PRs:
1. https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/873
2. https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/878
3. TODO: Implement location filter

## 📝 Description

This PR implements a location filter functionality in Events page that can be combined with category filtering. 

- **Type:**  New feature / Refactor

## 🛠️ Key Changes

I made comments in the code explaining major changes, but here's an overview:

- removed `count` prop in `Listbox` and similar components, because we can handle count in the parent component
- removed `CategoryResetButton`
- refactored `useCategory` to handle adding count and "All Categories" which is the reset option. 
- `useEventFilter` handles filtering by location and category. 

## 🧪 How to Test

- Test filtering on `blog`, `ecosystem-explorer`, `events` if they work as expected


src/app/_components/FilterListbox.tsx
```diff
@@ -0,0 +1,31 @@
...
+  selected: OptionType
+  options: Array<OptionType>
+  onChange: (selected: OptionType) => void
+}
+
+export function FilterListbox({
+  selected,
+  options,
+  onChange,
+}: FilterListboxProps) {
```
barbaraperic, line 20: Previously named `CategoryListbox` - now a more generic component so we can use it for both Category and Location filtering

src/app/_components/CategoryResetButton.tsx
barbaraperic, line 1: No need for this component, we handle reset with this part in `useCategory`

```
  if (
      !validatedCategoryOption ||
      validatedCategoryOption === ALL_CATEGORIES_OPTION.id
    ) {
      return entries
    }
```

src/app/_components/CategorySidebar.tsx
```diff
@@ -17,10 +13,9 @@ const touchTarget = {
 } as const
 
 type CategorySidebarProps = {
-  selectedId: CategoryId
-  options: Array<CategoryOption>
-  counts?: CategoryCounts
-  onChange: (selected: CategoryOption) => void
+  selectedId: OptionType['id']
+  options: Array<OptionType>
+  onChange: (selected: OptionType) => void
```
barbaraperic, line 18: We remove the `counts` props are we calculate and add it to `options` object in the parent component

src/app/_hooks/useCategory.ts
```diff
@@ -41,7 +48,10 @@ export function useCategory<Entry extends Object>({
   )
 
   const categorizedResults = useMemo(() => {
-    if (!validatedCategoryOption) {
+    if (
+      !validatedCategoryOption ||
+      validatedCategoryOption === ALL_CATEGORIES_OPTION.id
+    ) {
```
barbaraperic, line 54: this handles reset filter case when the user clicks "All Categories"

src/app/_components/FilterContainer.tsx
```diff
@@ -53,11 +56,13 @@ FilterContainer.MobileFiltersAndResults = function MobileFiltersAndResults({
 
 FilterContainer.DesktopFilters = function DesktopFilters({
   search,
+  location,
   sort,
 }: FilterProps) {
   return (
     <div className="hidden justify-end gap-6 lg:flex">
       {search}
+      <div>{location}</div>
```
CharlyMartin, line 65: This is probably where we should set a width, as we do in `MobileFiltersAndResults`, so that the layout doesn't shift as users select options.

Also, why do we check for the existence of `location` above and not here?

```ts
{location && ...}
```

src/app/_components/Category.tsx
```diff
@@ -1,60 +1,61 @@
...
   )
   const { updateSearchParams, resetSearchParams } = useUpdateSearchParams()
 
-  const selectedCategory = options.find((option) => option.id === categoryId)
+  const selectedCategory =
+    options.find((option) => option.id === categoryId) || ALL_CATEGORIES_OPTION
+
+  useEffect(() => {
+    setCategoryId(query || ALL_CATEGORIES_OPTION.id)
+  }, [query])
```
CharlyMartin, line 31: Why is this needed?

`updateCategoryAndParams` already takes care of updating the state no?

src/app/_components/LocationFilter.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+import { FilterListbox } from '@/components/FilterListbox'
+import { type OptionType } from '@/components/ListboxOption'
+
+type LocationFilterProps = {
+  query: ReturnType<typeof useCategory>['categoryQuery']
+  options: Array<OptionType>
+}
+
+export function LocationFilter({ query, options }: LocationFilterProps) {
```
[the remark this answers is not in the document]
CharlyMartin: They are indeed very similar! We might be able to create a wrapper component that handle setting the state of `FilterListbox`.

Then `Category` and `LocationFilter` would render that wrapper component.

Also, we can rename `LocationFilter` to `Location` to be consistent.
Or maybe `Category` -> `CategoryFilter` to be more descriptive.

What do you think?

src/app/_components/LocationFilter.tsx
```diff
@@ -0,0 +1,53 @@
...
+  )
+
+  const { updateSearchParams, resetSearchParams } = useUpdateSearchParams()
+
+  const selectedCategory =
+    options.find((option) => option.id === optionId) || options[0]
+
+  useEffect(() => {
+    setOptionId(query || ALL_LOCATIONS_OPTION.id)
+  }, [query])
```
CharlyMartin, line 31: Same comment as above - is this needed?

src/app/_constants/filterConstants.ts
```diff
@@ -0,0 +1,15 @@
+export const ALL_FILTER_ID = 'all' as const
+
+export type FilterOption = {
+  id: typeof ALL_FILTER_ID
+  name: string
+}
```
CharlyMartin, line 6: This should be somewhere else - maybe under the local `types` folder?

src/app/_hooks/useCategory.ts
```diff
@@ -12,14 +17,15 @@ export type UseCategoryProps<Entry extends Object> = {
   searchParams: NextServerSearchParams
   entries: Array<Entry>
   validCategoryIds: Array<CategoryId>
+  categoryOptions: Array<CategoryOption>
```
CharlyMartin, line 20: Why do we need to pass an array of options with count if we calculate `count` in the hook?

```ts
export type CategoryOption = {
  id: CategoryId
  name: string
  count?: number
}
```

src/app/_hooks/useCategory.ts
```diff
@@ -12,14 +17,15 @@ export type UseCategoryProps<Entry extends Object> = {
   searchParams: NextServerSearchParams
   entries: Array<Entry>
   validCategoryIds: Array<CategoryId>
```
CharlyMartin, line 19: We most likely don't need to pass `validCategoryIds` anymore now we pass `categoryOptions`.

We can derive the validIds from the options. What do you think?

src/app/events/hooks/useEventFilters.tsx
```diff
@@ -0,0 +1,146 @@
...
+const EVENT_COLLECTION_NAME = 'event_entries' as CMSCollectionName
+
+const filters: EventFilters = {
+  fields: {
+    CATEGORY: 'category',
+    LOCATION: 'location.region',
+  },
+} as const
+
+export const useEventFilters = ({
```
CharlyMartin, line 40: I haven't reviewed this fully yet - but I'm wondering if we need it.

My initial thought would be to rework `useCategory` and call it `useFilter`, so that it's a bit more flexible and we don't need to create these custom functions.

Like `useSort`, `useFilter` would take a filter function to handle custom filtering when necessary, like region for instance.

Something like that:

```
const { categoryQuery, categorizedResults } = useFilter({
    searchParams,
    entries: sortedResults,
    categoryOptions,
    filterFn: filterEventsByCategoryAndRegion
})
```

What do you think?

barbaraperic:
> @barbaraperic First feedback/question:
> 
> 1. The width of the Listbox button changes when selecting options. Is it what we want? - [Recording](https://share.cleanshot.com/hMDrBn7t)
> 2. Clicking `All Categories` resets the location filter. I thought we agreed not to do that, meaning `All Categories` only resets the categories - [Recording](https://share.cleanshot.com/WxMFnNV6)

Great catch! :) 

Should be fixed 👍

src/app/_components/FilterContainer.tsx
barbaraperic: So we avoid this in blog layout

Screenshot 2024-11-25 at 21 48 43

src/app/_components/Category.tsx
barbaraperic: you're right! think it was a leftover of something I tried

barbaraperic: ✅

src/app/_components/LocationFilter.tsx
barbaraperic: ✅

src/app/_hooks/useCategory.ts
barbaraperic: Ah yes, we should change the type. The `categoryOptions` we pass are without count, they look something like this
```
 categoryOptions: [
    { id: 'ecosystem', name: 'Ecosystem' },
    { id: 'events', name: 'Events' },
    { id: 'interviews', name: 'Interviews' },
    { id: 'news', name: 'News' },
    { id: 'reports', name: 'Reports' },
    { id: 'use-cases', name: 'Use Cases' }
  ]
```

barbaraperic: ✅

src/app/_hooks/useCategory.ts
barbaraperic: ✅

src/app/events/hooks/useEventFilters.tsx
barbaraperic: Completely agree! I had in mind we would need to refactor `useCategory`, as there's an overlap with `useEventsFilter. To be honest, it felt like there were too many parts to handle in order to create 1 hook that would handle all the cases, so I created a similar one to `useCategory`, having in mind we would need consolidate the two. 

I needed fresh eyes on this code lol.

mirhamasala:
Closing in favor of: 

- https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/878
- https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/888

PR #856, opened 2024-11-25
Fix Davos 2025 event [skip percy]

CharlyMartin:
## 📝 Description

This PR fixes the date in the title of the upcoming Davos event [[Slack](https://filecoinfoundation.slack.com/archives/C06MTEQ3SP6/p1732536070296089)]. It also makes the Location schema `strict()` to catch unwanted keys such as:

```yaml
location:
  primary: Davos, Switzerland
  Switzerland: null
  region: europe
```

## 📸 Screenshots

CleanShot 2024-11-25 at 12 14 14


mirhamasala (changes requested):
@CharlyMartin - The slug still says `2024`. We will also need a redirect.

mirhamasala (approved):

PR #864, opened 2024-11-27
[UXIT-1784] Events · Fixup Davos redirects [skip percy]

mirhamasala:
This PR updates the URL structure for the annual Davos event to adopt an evergreen URL approach. The evergreen URL (https://www.fil.org/davos) should always point to the most recent Davos event, ensuring consistency and strengthening its value over time.


CharlyMartin (approved):
🙌

PR #868, opened 2024-11-28
[UXIT-1782] Make Sponsor link required

barbaraperic:


CharlyMartin (changes requested):
Hi @barbaraperic 

Shouldn’t we also update the `SponsorsSchema` Zod schema to match the CMS configuration?

```ts
const sponsorSchema = z.object({
  name: z.string(),
  website: z.string().url(), // Remove .optional()
  image: ImagePropsSchema,
})
```

And If we do, we might need to backfill the data for sponsors with missing website links (I haven’t verified if any are missing)

barbaraperic:
> Hi @barbaraperic
> 
> Shouldn’t we also update the `SponsorsSchema` Zod schema to match the CMS configuration?
> 
> ```ts
> const sponsorSchema = z.object({
>   name: z.string(),
>   website: z.string().url(), // Remove .optional()
>   image: ImagePropsSchema,
> })
> ```
> 
> And If we do, we might need to backfill the data for sponsors with missing website links (I haven’t verified if any are missing)

Great points, fixed!

public/admin/config.yml
```diff
@@ -93,7 +93,7 @@ event_sponsor_config: &event_sponsor_config
   - name: "website"
     label: "website"
     widget: "string"
-    required: false
+    required: true
```
mirhamasala, line 96: This is the default.

mirhamasala (approved):
@barbaraperic - I removed [the broken link](https://filecoinfoundation.slack.com/archives/C06MTEQ3SP6/p1732859938176049). Also `required: true` is the default so removed that as well. Good to go.

PR #873, opened 2024-12-02
[UXIT-517] - Rework Category Filters

CharlyMartin:
## 📝 Description

This PR reworks how we implement filtering and lays the groundwork for implementing filtering with multiple fields on the events page.

It's the first of two to PR breaking down https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/849.

## 🛠️ Key Changes

- Reworks `useCategory` to only filter entries by their category. Once we introduce filtering by location, it will be renamed `useFilter`
- The "All Categories" option only resets the categories filter
- Introduces `useCategoryOptionsWithCount` to handle adding counts to options
- Use [nuqs](https://nuqs.47ng.com/), instead of our custom methods, in `Category`

## 🧪 How to Test

- Filter by category on the events and blog page and ensure the results and count make sense.

## 📸 Screenshots

CleanShot 2024-12-02 at 12 40 39


src/app/_components/FilterContainer.tsx
```diff
@@ -15,7 +15,7 @@ type FilterProps = {
 
 type ResultsProps = {
   category: ReactNode
-  results: ReactNode
+  results?: ReactNode
```
mirhamasala, line 18: How come this is optional now?

src/app/_hooks/useCategory.ts
```diff
@@ -1,70 +1,35 @@
...
 
 import { normalizeQueryParam } from '@/utils/queryUtils'
 
-export type UseCategoryProps<Entry extends Object> = {
-  searchParams: NextServerSearchParams
-  entries: Array<Entry>
-  validCategoryIds: Array<CategoryId>
+type EntryWithCategory = {
+  category: string
 }
```
mirhamasala, line 11: Why do we need this?

CharlyMartin: This is more accurate (and safer) than `Object`.

We are calling `entry.category` below, so the entries we pass need a category key.

src/app/_constants/categoryConstants.ts
```diff
@@ -1 +1,6 @@
-export const DEFAULT_CATEGORY = 'All'
+export const ALL_CATEGORIES_ID = 'all'
+
+export const ALL_CATEGORIES_OPTION = {
+  id: ALL_CATEGORIES_ID,
+  name: 'All Categories',
+} as const
```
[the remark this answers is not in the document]
CharlyMartin: For this PR, no valid reason, so we can have everything in `ALL_CATEGORIES_OPTION`

In the second PR, they are decoupled as there is also this:

```ts
export const ALL_LOCATIONS_OPTION = {
  id: ALL_FILTER_ID,
  name: 'All Locations',
} as const
```

src/app/_components/FilterContainer.tsx
CharlyMartin: This is because we render `ResultsAndReset` for the ecosystem projects, which is pass via this prop.

For blogs and events, we don't a global filter reset anymore.

src/app/_hooks/useCategory.ts
```diff
@@ -1,70 +1,35 @@
...
 
 import { CATEGORY_KEY } from '@/constants/searchParams'
 
 import { normalizeQueryParam } from '@/utils/queryUtils'
 
-export type UseCategoryProps<Entry extends Object> = {
-  searchParams: NextServerSearchParams
-  entries: Array<Entry>
-  validCategoryIds: Array<CategoryId>
+type EntryWithCategory = {
```
mirhamasala, line 9: `WithCategory` or `HasCategory`

src/app/_hooks/useCategory.ts
mirhamasala: As discussed, let's go with `HasCategory` or `WithCategory`.

mirhamasala (approved):
Yay! One step closer. LGTM! 🙌🏼

PR #878, opened 2024-12-03
[UXIT-517] - Rework filtering to be more flexible

CharlyMartin:
## 📝 Description

This PR is the second PR (of 3) breaking down https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/849.

It renames the category-related code and updates `useCategory` - now `useFilter` - making it more flexible and allowing filtering by any field, not just category.

## 🛠️ Key Changes

- The first 8 commits are mostly renaming: `category` becomes `filter`
- The last [commit](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/878/commits/91984c563b34c40c986461e345cf17515f5d9ae6) reworks the filter hook to accept a key and callback function.

## 🧪 How to Test

- Filtering by category on `/blog` and `/events` should work the same as before


src/app/events/utils/filterEvent.ts
```diff
@@ -0,0 +1,5 @@
+import type { Event } from '@/events/types/eventType'
+
```
CharlyMartin, line 2: The file should be renamed `filterEvents.ts` (plural)

src/app/_hooks/useFilterOptionsWithCount.ts
```diff
@@ -40,11 +40,11 @@ export function useCategoryOptionsWithCount<Entry extends Object>({
...
       ...allOption,
-      count: getTotalCategoryCount(optionsWithCount),
+      count: sumObjectCounts(optionsWithCount),
     }
 
     return [allOptionWithCount, ...optionsWithCount]
   }, [allOption, optionsWithCount])
 
-  return { categoryOptionsWithCount: optionsWithAllAndCount }
+  return optionsWithAllAndCount
```
mirhamasala, line 49: I find quite a few namings in this hook confusing - Perhaps we can pair and find some clarity.

src/app/_utils/sumObjectCounts.ts
```diff
@@ -0,0 +1,14 @@
+type WithCount = {
+  count: number
+}
+
+const COUNT_INITIAL_VALUE = 0
+
+export function sumObjectCounts<AnyObject extends WithCount>(
```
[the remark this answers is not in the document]
CharlyMartin: I'm trying to express that this function sums the count value of all objects in the array.

Maybe `sumCountValues`? Or `sumObjectCountValues`?

`calculateTotalCount` also works, but I find it a bit less descriptive.

src/app/blog/utils/filterBlogPosts.ts
mirhamasala, line 1: What about for `filterBlogPosts` and filterEvents`:

```
type WithCategory = {
  category: string;
}

export function matchesCategory<T extends WithCategory>(
  item: T,
  query?: string
): boolean {
  return item.category === query;
}
```

src/app/_hooks/useFilter.ts
```diff
@@ -0,0 +1,39 @@
...
+  entries: Array<Entry>
+  filterKey: string
+  filterFn: (entry: Entry, query?: string) => boolean
+}
+
+export function useFilter<Entry extends Object>({
+  searchParams,
+  entries,
+  filterKey,
+  filterFn,
```
mirhamasala, line 21: In this PR, `filterKey` and `filterFn` are the same in both use cases. Will this change?

src/app/blog/utils/filterBlogPosts.ts
CharlyMartin: Yes, I like that :)

src/app/_hooks/useFilter.ts
CharlyMartin: Yes, it will change when we add location to the mix.

```ts
const { filteredResults } = useFilter({
  searchParams,
  entries: sortedResults,
  filterKey: LOCATION_KEY,
  filterFn: filterByLocation,
})
```

The goal of this PR is to make `useFilter` more agnostic and allow filtering by any field.

src/app/_hooks/useFilterOptionsWithCount.ts
mirhamasala, line 1: - `useListboxOptions`
- `defaultOption`
- `DEFAULT_FILTER_OPTION` / `DEFAULT_FILTER_ID`

src/app/_hooks/useFilter.ts
```diff
@@ -0,0 +1,39 @@
...
+}: UseFilterProps<Entry>) {
+  const filterQuery = normalizeQueryParam(searchParams, filterKey)
+
+  const filterByQuery = useCallback(
+    (entry: Entry) => filterFn(entry, filterQuery),
+    [filterFn, filterQuery],
+  )
+
+  const filteredResults = useMemo(() => {
```
mirhamasala, line 30: `filteredEntries`? - Can we look into the rest where we use entries vs. results just to make sure it still all makes sense.

src/app/blog/utils/filterBlogPosts.ts
CharlyMartin: ✅

src/app/_hooks/useFilter.ts
CharlyMartin: The other hooks use results instead of entries -> [Ticket](https://www.notion.so/filecoin/FF-Followup-Dev-Work-d6616163f43c4e7db78edf218cebcc8d?pvs=4#1537631f282580d98665c69c4e4b328d)

src/app/_hooks/useFilterOptionsWithCount.ts
CharlyMartin: ✅

src/app/_utils/sumObjectCounts.ts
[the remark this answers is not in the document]
CharlyMartin: I went with `sumCountValues` - feel free to rename if you have a better idea

mirhamasala (approved):
Yay! Looking super nice! 👌🏽

PR #879, opened 2024-12-03
[UXIT-831] Add 404 tracking code for Plausible.io

barbaraperic:
## 📝 Description

This PR adds script to track 404 events in plausible.

- **Type:** New feature

## 🧪 How to Test

Plausible only tracks events from domains that are registered in Plausible account, so we can't test from localhost. 

## 🔖 Resources

- [Plausible - error pages tracking 404](https://plausible.io/docs/error-pages-tracking-404)
- [Nextjs - script](https://nextjs.org/docs/pages/api-reference/components/script)
- [NextJS recipe: Tracking 404s with Plausible and NextJS App Router](https://calebfaruki.me/nextjs-plausible-app-router-404)

### DIFF
--- a/src/app/not-found-analytics.tsx
+++ b/src/app/not-found-analytics.tsx
@@ -1,0 +1,17 @@
+'use client'
+
+import Script from 'next/script'
+
+import { usePlausible } from 'next-plausible'
+
+export default function NotFoundAnalytics() {
+  const plausible = usePlausible()
+
+  return (
+    <Script
+      onReady={() =>
+        plausible('404', { props: { path: window.location.pathname } })
+      }
+    />
+  )
+}
--- a/src/app/not-found.tsx
+++ b/src/app/not-found.tsx
@@ -1,9 +1,14 @@
 import ErrorMessage from '@/components/ErrorMessage'
 
+import NotFoundAnalytics from './not-found-analytics'
+
 export default function NotFound() {
   return (
-    <ErrorMessage kicker="404" title="Page Not Found">
-      We&apos;re sorry, but the page you were looking for is not here.
-    </ErrorMessage>
+    <>
+      <NotFoundAnalytics />
+      <ErrorMessage kicker="404" title="Page Not Found">
+        We&apos;re sorry, but the page you were looking for is not here.
+      </ErrorMessage>
+    </>
   )
 }


mirhamasala (changes requested):
Since we're already using next-plausible in our project, we could use its built-in `usePlausible` hook instead of adding another `Script` component. This keeps our analytics approach consistent across the codebase and lets us easily capture additional context like referrer data. We can update the `NotFoundPage` component to use `useEffect` with the hook to track 404s, capturing both the `path` and `referrer`.
We could

```
'use client'

import { useEffect } from 'react'

import { usePlausible } from 'next-plausible'

import ErrorMessage from '@/components/ErrorMessage'

export default function NotFoundPage() {
  const plausible = usePlausible()

  useEffect(() => {
    plausible('404', {
      props: {
        path: window.location.pathname,
        referrer: document.referrer,
      },
    })
  }, [plausible])

  return (
    <ErrorMessage kicker="404" title="Page Not Found">
      We&apos;re sorry, but the page you were looking for is not here.
    </ErrorMessage>
  )
}

```

barbaraperic:
> Since we're already using next-plausible in our project, we could use its built-in `usePlausible` hook instead of adding another `Script` component. This keeps our analytics approach consistent across the codebase and lets us easily capture additional context like referrer data. We can update the `NotFoundPage` component to use `useEffect` with the hook to track 404s, capturing both the `path` and `referrer`.
> 
> ```
> 'use client'
> 
> import { useEffect } from 'react'
> 
> import { usePlausible } from 'next-plausible'
> 
> import ErrorMessage from '@/components/ErrorMessage'
> 
> export default function NotFoundPage() {
>   const plausible = usePlausible()
> 
>   useEffect(() => {
>     plausible('404', {
>       props: {
>         path: window.location.pathname,
>         referrer: document.referrer,
>       },
>     })
>   }, [plausible])
> 
>   return (
>     <ErrorMessage kicker="404" title="Page Not Found">
>       We&apos;re sorry, but the page you were looking for is not here.
>     </ErrorMessage>
>   )
> }
> ```

Yea that was my initial idea, but wanted to keep not-found a server component. Don't mind this solution!

mirhamasala:
@barbaraperic - What about:

```
'use client'
import { useEffect } from 'react'
import { usePlausible } from 'next-plausible'

export function NotFoundAnalytics() {
  const plausible = usePlausible()

  useEffect(() => {
    plausible('404', {
      props: {
        path: window.location.pathname,
        referrer: document.referrer,
      },
    })
  }, [plausible])

  return null
}
```

mirhamasala:
@barbaraperic - What about:

```
'use client'

import { useEffect } from 'react'

import { usePlausible } from 'next-plausible'

export function NotFoundAnalytics() {
  const plausible = usePlausible()

  useEffect(() => {
    plausible('404', {
      props: {
        path: window.location.pathname,
        referrer: document.referrer,
      },
    })
  }, [plausible])

  return null
}

```

CharlyMartin (commented):
I think it's a good idea to keep the main component a server component.

I would roll with what you suggested last:

```ts
'use client'

import { useEffect } from 'react'

import { usePlausible } from 'next-plausible'

export function NotFoundAnalytics() {
  const plausible = usePlausible()

  useEffect(() => {
    plausible('404', {
      props: {
        path: window.location.pathname,
        referrer: document.referrer,
      },
    })
  }, [plausible])

  return null
}
```

mirhamasala (approved):
🙌🏼 LGTM!

PR #888, opened 2024-12-05
[UXIT-517] - Add Location Filter to Events

CharlyMartin:
## 📝 Description

This PR adds location-based filtering to the events page.

## 🛠️ Key Changes

To do so, this PR:
- Renames `Category` to `CategoryFilter` 
- Creates a `LocationFilter` component
- Adds a location prop to `FilterContainer`
- Adds an `entryMatchesLocationQuery` filter function 
- Creates `getCMSOptionsWithDefault` when the count is not needed for listbox options, such as `LocationFilter`

## 🧪 How to Test

- Go to /events and try filtering by location: Events and counts in the category sidebar should update accordingly.

## 📌 To-Do Before Merging

- [x] Review UI changes Filipa

## 📸 Screenshots

CleanShot 2024-12-09 at 11 24 58

CleanShot 2024-12-09 at 11 25 18


src/app/_components/FilterContainer.tsx
```diff
@@ -6,10 +6,17 @@ type LayoutProps = {
   children: ReactNode
 }
 
-type FilterProps = {
+type DesktopFilterProps = {
   search: ReactNode
+  location?: ReactNode
```
mirhamasala, line 11: Oei - I'm passing something as specific as `location` here slightly weird, it's like the component knows too much?

src/app/_components/FilterContainer.tsx
```diff
@@ -31,17 +38,27 @@ export function FilterContainer({ children }: LayoutProps) {
...
+
 FilterContainer.MobileFiltersAndResults = function MobileFiltersAndResults({
   search,
   category,
+  location,
   sort,
-}: FilterProps) {
+}: MobileFilterProps) {
+  const filtersCount = location ? 2 : 1
+  const filterStyle = mobileFilterContainerStyle[filtersCount]
```
mirhamasala, line 53: I really don't like this 🙈 Maybe this will help [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/888/files#r1875838345)

src/app/_components/LocationFilter.tsx
mirhamasala, line 1: Can this be refactored with `CategoryFilter` into a custom hook? Something like:

```
function useFilterQuery(key: string, options: Array<OptionType>, defaultOption: OptionType) {
  const [optionId, setOptionId] = useQueryState(
    key,
    parseAsString
      .withDefault(DEFAULT_FILTER_ID)
      .withOptions({ shallow: false })
  )

  const selectedOption = 
    options.find((option) => option.id === optionId) || defaultOption

  const handleChange = (option: OptionType) => setOptionId(option.id)

  return {
    selectedOption,
    handleChange
  }
}
```

src/app/_utils/getCMSOptionsWithDefault.ts
```diff
@@ -0,0 +1,23 @@
+import type { DefaultFilterOptionType } from '@/types/filterTypes'
+
+import {
+  getCMSFieldOptionsAndValidIds,
+  type CMSFieldOptionsAndValidIdsParams,
+} from './getCMSFieldOptionsAndValidIds'
+
+type TODO = CMSFieldOptionsAndValidIdsParams & {
```
mirhamasala, line 8: `TODO`?

mirhamasala: Is there a reason for not merging this with `getCMSFieldOptionsAndValidIds`?

src/app/events/page.tsx
```diff
@@ -76,6 +84,12 @@ export const metadata = createMetadata({
   overrideDefaultTitle: true,
 })
 
+const locationOptions = getCMSOptionsWithDefault({
```
mirhamasala, line 87: Should we abstract some of this config?

```
export const EVENT_CMS_OPTIONS = {
  location: {
    collectionName: 'event_entries',
    fieldName: 'location.region',
    defaultOption: DEFAULT_LOCATION_FILTER_OPTION,
  },
  category: {
    collectionName: 'event_entries',
    fieldName: 'category',
  }
} as const
```

src/app/events/page.tsx
```diff
@@ -94,23 +108,30 @@ export default function Events({ searchParams }: Props) {
...
   })
 
-  const { currentPage, pageCount, paginatedResults } = usePagination({
+  const { filteredEntries: filteredEventsByCategory } = useFilter({
     searchParams,
-    entries: filteredEntries,
+    entries: filteredEventsByLocation,
+    filterKey: CATEGORY_KEY,
+    filterFn: entryMatchesCategoryQuery,
   })
```
mirhamasala, line 123: I wonder if we should have something like:

```
// hooks/useFilteredEvents.ts

import { type Event } from '@/types/events'
import { type NextServerSearchParams } from '@/types/searchParams'
import { CATEGORY_KEY, LOCATION_KEY } from '@/constants/searchParams'
import { entryMatchesCategoryQuery, entryMatchesLocationQuery } from '@/utils/filterUtils'
import { useFilter } from '@/hooks/useFilter'

export function useFilteredEvents({
  searchParams,
  entries,
}: {
  entries: Event[]
}) {
  const { filteredEntries: locationFiltered } = useFilter({
    searchParams,
    entries,
    filterKey: LOCATION_KEY,
    filterFn: entryMatchesLocationQuery,
  })

  const { filteredEntries: filteredEvents } = useFilter({
    searchParams,
    entries: locationFiltered,
    filterKey: CATEGORY_KEY,
    filterFn: entryMatchesCategoryQuery,
  })

  return filteredEvents
}
```

mirhamasala (changes requested):
@CharlyMartin - Left a few comments. I think this should get you going. Will have another look after that! Thank you. 🙏🏼

src/app/_components/FilterContainer.tsx
CharlyMartin: I agree, it could be a nice improvement from what we have.

This means reworking the blog, events and ecosystem explorer pages. How about we keep it for a different PR?

src/app/_utils/getCMSOptionsWithDefault.ts
CharlyMartin: 🙃

src/app/events/page.tsx
CharlyMartin: We use `filteredEventsByLocation` in

```ts
  const { optionsWithCount: categoryOptionsWithCount } = useListboxOptions({
    collectionName: 'event_entries',
    fieldName: 'category',
    defaultOption: DEFAULT_CATEGORY_FILTER_OPTION,
    entries: filteredEventsByLocation,
  })
```

Hence we have to keep both hooks independent for that reason.

src/app/_components/LocationFilter.tsx
CharlyMartin: Yes, that's a nice addition :)

CharlyMartin: I called it `useFilterListboxState` to emphasise the fact that it's the companion hook to `FilterListbox`

mirhamasala: Might be another case of this [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/888#discussion_r1876025442)

src/app/events/page.tsx
CharlyMartin: I went with `EVENT_FILTER_OPTION_SETTINGS`, feel free to rename it if you have a better suggestion

src/app/events/page.tsx
CharlyMartin: Not sure it adds a lot of value. And we end up with the same naming issue as before where `filteredEventsByCategory` includes both filters but its name suggests otherwise.

I would keep it as is, but up to you!

src/app/_components/FilterContainer.tsx
mirhamasala: I think we misunderstood each other - It shouldn't be part of this PR .. this PR can be merged last.

CharlyMartin: I implemented the changes here: https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/903

CharlyMartin: Fixed by https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/903

src/app/_components/FilterContainer.tsx
CharlyMartin: Fixed by https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/903

src/app/events/page.tsx
mirhamasala: I renamed it to `FILTERS_CONFIG` - Shorter and it's scoped under Events either way. 🙏🏼

mirhamasala (changes requested):
@CharlyMartin - This is excellent! ✨ I think we are almost there. (Pushed nit CRs.)

It seems we are missing `Virtual` as an option still? Probably best done in a separate PR branched off from his one?

Once it's all in place, can you please pair with Filipa to walk through all of it visually and make sure we're good? Thank you. 🙏🏼

CharlyMartin:
@mirhamasala Thanks for the review and rework!

> It seems we are missing `Virtual` as an option still? Probably best done in a separate PR branched off from his one?

We said we would exclude virtual events in our original specs, but we might as well add it to be thorough: https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/909

> Once it's all in place, can you please pair with Filipa to walk through all of it visually and make sure we're good? Thank you. 🙏🏼

I booked a meeting with Filipa tomorrow to discuss this.

PR #897, opened 2024-12-09
[UXIT-1807] Ensure External Link Arrow Stays Aligned with Text When Wrapping

barbaraperic:
## 📝 Description

**Type:** Bug fix / Refactor

This PR fixes the issue with have with arrow alignment for external links. I've used [this blog](https://stephango.com/black-pixels) as an inspiration - where arrow icons are background images.

It's not ideal to have arrow icon as svgs and with hardcoded color, but I didn't find an elegant solution where we could use the `Icon` element. This solution seems to work quite nice with text wrapping tho. 

### Before
Screenshot 2024-12-09 at 09 19 38

### After
Screenshot 2024-12-09 at 09 19 26


barbaraperic:
[Saw this ticket](https://www.notion.so/filecoin/FF-Refactor-table-styles-from-addComponents-to-a-dedicated-module-css-file-1347631f28258072b21fdb237c183904?pvs=4) - I'll update so the styles are in tailwind.config

mirhamasala (changes requested):
@barbaraperic - Not a fan of this implementation for the reason you mentioned. What about the following:

```ts
<a className={clsx(styles.base, className)} {...rest}>
  {children}
  <span className="ml-1 inline-flex align-text-top">
    <Icon component={ArrowUpRight} size={16} color="brand-400" />
  </span>
</a>
```

I didn't have time to test the code everywhere but initially it looked good. Can you check if it looks good everywhere?

mirhamasala:
> I like that, yes it seems to work well. 👌

@barbaraperic - It didn't seem to be there yet - See screenshots. This commit should've fixed it [188973a](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/897/commits/188973a13f2f36cbc4e2e43eed4a8853e91f6707)

CleanShot 2024-12-10 at 04 39 53@2x
CleanShot 2024-12-10 at 04 37 50@2x

PR #903, opened 2024-12-09
[UXIT-517] - Rework Filter Container To Accept More Filters

CharlyMartin:
## 📝 Description

This PR refactors `FilterContainer` to use an array of filters, instead of named filter props.

### DIFF
--- a/src/app/_components/FilterContainer.tsx
+++ b/src/app/_components/FilterContainer.tsx
@@ -6,18 +6,18 @@ type LayoutProps = {
   children: ReactNode
 }
 
-type DesktopFilterProps = {
+type FilterComponents = [ReactNode, ReactNode?]
+
+type DesktopFiltersProps = {
   search: ReactNode
-  location?: ReactNode
+  filters?: FilterComponents
   sort: ReactNode
 }
 
-type MobileFilterProps = {
+type MobileFiltersProps = {
   search: ReactNode
+  filters: FilterComponents
   sort: ReactNode
-  category: ReactNode
-  location?: ReactNode
-  results?: ReactNode
 }
 
 type ResultsProps = {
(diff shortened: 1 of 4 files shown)


src/app/_components/FilterContainer.tsx
```diff
@@ -38,44 +38,48 @@ export function FilterContainer({ children }: LayoutProps) {
...
+  filters,
   sort,
-}: MobileFilterProps) {
-  const filtersCount = location ? 2 : 1
-  const filterStyle = mobileFilterContainerStyle[filtersCount]
+}: MobileFiltersProps) {
+  const filtersCount = filters.length
+  const filterContainerWidth = clsx(
+    filtersCount === 1 && 'flex-1 sm:w-64 md:w-44',
+    filtersCount === 2 && 'w-2/5 flex-1 min-w-0 sm:w-52 md:w-44',
```
mirhamasala, line 49: `flex-1` and `md:w-44` seem to be shared..

src/app/_components/FilterContainer.tsx
```diff
@@ -6,18 +6,18 @@ type LayoutProps = {
   children: ReactNode
 }
 
-type DesktopFilterProps = {
+type FilterComponents = [ReactNode, ReactNode?]
+
+type DesktopFiltersProps = {
   search: ReactNode
-  location?: ReactNode
+  filters?: FilterComponents
```
mirhamasala, line 13: @CharlyMartin - I remember we discussed wanting to have clearer naming for `filters` prop since in its current state it could be ambiguous - it might refer to filter values/data rather than the UI components.

Would this be the right place to change it? To something like `filterComponents`?

mirhamasala (approved):
@CharlyMartin - I love this. Just a couple of nits. Have a look and feel free to merge any time.

src/app/_components/FilterContainer.tsx
CharlyMartin: Yes we can have `filterComponents`, `searchComponent` and `sortComponent`

PR #904, opened 2024-12-09
Update Blog Post “filecoin-foundation-s-vision-and-strategic-priorities-for-2025”

matthendrian:
Automatically generated by Decap CMS


mirhamasala:
@matthendrian - The abbreviation FF seems to be used inconsistently. Do we need it? Could we just write out Filecoin Foundation everywhere?

matthendrian:
@mirhamasala I think you're right, we would be totally fine to just write out "Filecoin Foundation" in the headings to keep it consistent. We can leave the FF abbreviation in the body text - otherwise it starts to read a little clunky since it is mentioned so often.

PR #911, opened 2024-12-12
[UXIT-1880] - Rework event filters on XS devices

CharlyMartin:
## 📝 Description

This PR updates the sort and filter Listbox `components` to show different icons on the events page, after a discussion with Filipa this morning.

## 🛠️ Key Changes
- Adds an icon prop to `FilterListbox` and `SortListbox`.
- Creates `EventSort` for the events page, which relies on `SortListbox`
- Updates `SortListbox` to use `OptionType` instead of `SortType`, enabling reuse of `useFilterListboxState` (now renamed `useListboxQueryState`).

## 📌 To-Do Before Merging

- [x] Get the greenlight from Filipa - [Slack](https://filecoinfoundation.slack.com/archives/C07UH9HAURX/p1734016268364299?thread_ts=1733743660.737509&cid=C07UH9HAURX)

## 📸 Screenshots

image

### DIFF
--- a/src/app/_components/FilterListbox.tsx
+++ b/src/app/_components/FilterListbox.tsx
@@ -3,7 +3,10 @@
 import { Listbox } from '@headlessui/react'
 import { FunnelSimple } from '@phosphor-icons/react'
 
-import { ListboxButton } from '@/components/Listbox/ListboxButton'
+import {
+  ListboxButton,
+  type ListboxButtonProps,
+} from '@/components/Listbox/ListboxButton'
 import {
   ListboxOption,
   type OptionType,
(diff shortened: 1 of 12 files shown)


src/app/_components/FilterListbox.tsx
```diff
@@ -3,7 +3,10 @@
 import { Listbox } from '@headlessui/react'
 import { FunnelSimple } from '@phosphor-icons/react'
```
mirhamasala, line 4: What about making this the default?

src/app/_components/Sort.tsx
```diff
@@ -40,6 +42,7 @@ export function Sort({ query, options, defaultQuery }: SortProps) {
     <SortListbox
       options={options}
       selected={selectedOption}
+      icon={ArrowsDownUp}
```
mirhamasala, line 45: Same comments here - `buttonIcon` may be clearer - And We could have `ArrowsDownUp` as the default?

mirhamasala (approved):
@CharlyMartin - Just a couple of nits. Butl ooking good! 😊

One extra observation: Resizing on blog, doesn't make the Filter collapse with `...` while it does on events?

CleanShot 2024-12-16 at 08 59 36@2x
CleanShot 2024-12-16 at 08 59 49@2x

PR #916, opened 2024-12-16
[UXIT-1796] Event Page · Pass Pre-Sorted Program Events to ProgramSection

mirhamasala:
## 📝 Description

This PR refactors the handling of program event sorting by moving the sorting logic from the `ProgramSection` component to its parent. This change centralizes data preparation in the parent, improving separation of concerns and simplifying the child component's responsibilities.

Additional improvements include:
- Adding the `NonEmptyArray<T>` type definition to ensure non-empty array typings for sorted data.
- Introducing the `sortNonEmptyEventsAsc` utility function to handle sorting of `NonEmptyArray<T>` events in ascending order.

Note:

The main issue was preserving the non-empty array type guarantee from our Zod schema through our sorting operations. Rather than trying to handle readonly tuples or complex type conversions, we:

1. Kept the original `sortEventsAsc` for regular arrays
2. Added a new `sortNonEmptyEventsAsc` specifically for `NonEmptyArray<T>`

This simpler approach maintains type safety while avoiding the complexity of readonly tuples and type conversions. It gives us both a general-purpose array sort and a specialized version that preserves the non-empty guarantee when we need it.

- **Type:** Refactor

## 🛠️ Key Changes

- **Move sorting logic to parent:** Sorting of program events is now handled in the parent component before passing the data to `ProgramSection`.
- **Add `NonEmptyArray<T>` type definition:** Ensures type safety for non-empty arrays used in sorting.
- **Introduce `sortNonEmptyEventsAsc`:** A utility function for consistent ascending order sorting of `non-empty` events.

## 🧪 How to Test

- **Setup:** Ensure the project is running locally and updated with the latest changes.
- **Steps to Test:**
  1. Navigate to event pages with program events.
  2. Confirm that program events are displayed in ascending order by their start date.
  3. Check that the `ProgramSection` component does not contain any sorting logic.
- **Expected Results:**
  - Program events should render in ascending order consistently.
  - `ProgramSection` should only display data without additional processing.
- **Additional Notes:** Focus testing on scenarios where program events might have edge cases (e.g., empty or unsorted data).


src/app/events/utils/sortEvents.ts
```diff
@@ -27,6 +29,12 @@ export function sortEventsAsc<T extends DateFields>(events: Array<T>) {
   return sortEvents(events, compareAsc)
 }
 
+export function sortNonEmptyEventsAsc<T extends DateFields>(
+  events: NonEmptyArray<T>,
+) {
+  return sortEvents(events, compareAsc) as NonEmptyArray<T>
+}
```
CharlyMartin, line 36: Making this suggestion so that both types stay in sync.

The "risk" with assertions is forgetting to update them when making other changes that impact them.

Commit if you like it, it's not blocking.

```suggestion
export function sortNonEmptyEventsAsc<
  T extends DateFields,
  Array extends NonEmptyArray<T>,
>(events: Array) {
  return sortEvents(events, compareAsc) as Array
}
```

CharlyMartin (approved):
One suggestion! Otherwise, Let's Get This Merged 🍔

src/app/events/utils/sortEvents.ts
mirhamasala: I love that. Done. Thank you.

PR #919, opened 2024-12-16
[UXIT-1731] Validate Image Src

barbaraperic:
## 📝 Description

**Type:** Refactor

This PR introduces `SmartImage` wrapper component (_name inspired by `SmartTextLink`_). The purpose of the component is to handle image URL validity and conditionally render the fallback.

## 🛠️ Key Changes

Currently it is used in `Card`, `PageHeader` (_app/components_) , `ArticleHeader`, `PostHeader` (_ecosystem_explorer/components_). These components require fallback handling because images in markdown frontmatter are optional.

With use of this component we can remove the "`image || graphicsData.imageFallback.data`" condition and just pass the `image` as a prop.

`SmartImage` will render the fallback (_currently fixed but can be customizable in the future_) in the following cases:
- The image is missing in the markdown frontmatter.
- The src URL in the markdown frontmatter is invalid.
- A remote image URL is passed, but the URL is broken.

## 🧪 How to Test

- Check if the fallback is shown in the cases described above.

### DIFF
--- a/apps/site/src/app/(homepage)/components/FeaturedBlogPosts.tsx
+++ b/apps/site/src/app/(homepage)/components/FeaturedBlogPosts.tsx
@@ -37,16 +37,18 @@ export function FeaturedBlogPosts({
                 href: `${PATHS.BLOG.path}/${slug}`,
                 text: 'Learn More',
               }}
-              image={{
-                ...image,
-                alt: '',
-                objectFit: 'cover',
-                sizes: buildImageSizeProp({
-                  startSize: '100vw',
-                  sm: '350px',
-                  md: '480px',
-                }),
-              }}
+              image={
+                image && {
+                  src: image.src,
+                  alt: image.alt || '',
+                  objectFit: 'cover',
+                  sizes: buildImageSizeProp({
+                    startSize: '100vw',
+                    sm: '350px',
+                    md: '480px',
+                  }),
+                }
+              }
             />
           )
         },
--- a/apps/site/src/app/(homepage)/components/FeaturedBlogPosts.tsx
+++ b/apps/site/src/app/(homepage)/components/FeaturedBlogPosts.tsx
@@ -1,7 +1,5 @@
 import { PATHS } from '@/constants/paths'
 
-import { graphicsData } from '@/data/graphicsData'
-
 import { buildImageSizeProp } from '@/utils/buildImageSizeProp'
 import { getCategoryLabel } from '@/utils/categoryUtils'
 
--- a/apps/site/src/app/(homepage)/components/FeaturedBlogPosts.tsx
+++ b/apps/site/src/app/(homepage)/components/FeaturedBlogPosts.tsx
@@ -40,7 +40,7 @@ export function FeaturedBlogPosts({
                 text: 'Learn More',
               }}
               image={{
-                ...(image || graphicsData.imageFallback.data),
+                ...image,
                 alt: '',
                 objectFit: 'cover',
                 sizes: buildImageSizeProp({
--- a/apps/site/src/app/(homepage)/components/FeaturedEcosystemProjects.tsx
+++ b/apps/site/src/app/(homepage)/components/FeaturedEcosystemProjects.tsx
@@ -30,18 +30,20 @@ export function FeaturedEcosystemProjects({
             text: 'Learn More',
             icon: MagnifyingGlass,
           }}
-          image={{
-            ...image,
-            alt: '',
-            objectFit: 'contain',
-            padding: Boolean(image),
-            sizes: buildImageSizeProp({
-              startSize: '100vw',
-              sm: '320px',
-              md: '440px',
-              lg: '280px',
-            }),
-          }}
+          image={
+            image && {
+              src: image.src,
+              alt: image.alt || '',
+              objectFit: 'contain',
+              padding: Boolean(image),
+              sizes: buildImageSizeProp({
+                startSize: '100vw',
+                sm: '320px',
+                md: '440px',
+                lg: '280px',
+              }),
+            }
+          }
         />
       ))}
     </CardGrid>
--- a/apps/site/src/app/(homepage)/components/FeaturedEcosystemProjects.tsx
+++ b/apps/site/src/app/(homepage)/components/FeaturedEcosystemProjects.tsx
@@ -2,8 +2,6 @@ import { MagnifyingGlass } from '@phosphor-icons/react/dist/ssr'
 
 import { PATHS } from '@/constants/paths'
 
-import { graphicsData } from '@/data/graphicsData'
-
 import { buildImageSizeProp } from '@/utils/buildImageSizeProp'
 
 import { Card } from '@/components/Card'
--- a/apps/site/src/app/(homepage)/components/FeaturedEcosystemProjects.tsx
+++ b/apps/site/src/app/(homepage)/components/FeaturedEcosystemProjects.tsx
@@ -33,7 +33,7 @@ export function FeaturedEcosystemProjects({
             icon: MagnifyingGlass,
           }}
           image={{
-            ...(image || graphicsData.imageFallback.data),
+            ...image,
             alt: '',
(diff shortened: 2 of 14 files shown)


src/app/_components/Card.tsx
```diff
@@ -114,18 +107,18 @@ Card.Image = function ImageComponent({
 
   if (isStaticImage) {
     return (
-      <Image
+      <SmartImage
         {...commonProps}
         className={clsx(commonProps.className, 'aspect-video')}
-        src={image.data}
+        src={(image.data as StaticImageData).src}
```
barbaraperic, line 113: Typescript is complaining here although we have a `isStaticImage` condition (which check `data` in the `image`)

src/app/_components/SmartImage.tsx
```diff
@@ -0,0 +1,86 @@
+'use server'
```
barbaraperic, line 1: we need to be explicit to run on the server because we use fs and path which exclusively run in the node env

src/app/_components/SmartImage.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { graphicsData } from '@/data/graphicsData'
+
+const { data: fallbackSrc, alt: fallbackAlt } = graphicsData.imageFallback
+
+export type SmartImageProps = {
+  src?: string | StaticImageProps['data']
+  alt?: string
+  className?: string
+  objectFit?: ImageObjectFit
+  padding?: boolean
```
CharlyMartin, line 20: What does `padding` do?

src/app/_components/SmartImage.tsx
```diff
@@ -0,0 +1,86 @@
...
+} & Omit<ImageProps, 'src' | 'alt' | 'className'>
+
+export async function SmartImage({
```
CharlyMartin, line 23: What do you think about something like this?

Does it read a bit better to you?


```ts
export async function SmartImage({ src, alt = '', ...props }: SmartImageProps) {
  if (!src) {
    return <Image src={fallbackSrc} alt={fallbackAlt} {...props} />
  }

  const isValid = await checkImageValidity(src)

  return (
    <Image
      src={isValid ? src : fallbackSrc}
      alt={isValid ? alt : fallbackAlt}
      {...props}
    />
  )
}

async function checkImageValidity(src: NonNullable<SmartImageProps['src']>) {
  const isStaticImport = typeof src === 'object'
  const isAssetImage = typeof src === 'string' && src.startsWith('/assets')

  if (isStaticImport) {
    return true
  }

  if (isAssetImage) {
    return await checkAssetImageExists(src)
  }

  return await checkRemoteImageExists(src)
}
```

src/app/_components/Card.tsx
CharlyMartin: It should just be `image.src` no? `image.data` doesn't exist on `SmartImageProps`

CharlyMartin (changes requested):
I did a first pass with quite a few questions / suggestions.

Happy to have another look once the comments are resolved.

Thanks Barbara!

PS: Should statically imported images be part of the smart image component? Since we already know these are valid, it might not be necessary.

src/app/_components/Card.tsx
barbaraperic: Sorry not sure I understand 😕 `StaticImages` have data key, like this: 

```
{
  data: {
    src: '/_next/static/media/022624-ff-anualreport.b5deae8f.png',
    height: 1536,
    width: 3072,
    blurDataURL: '/_next/image?url=%2F_next%2Fstatic%2Fmedia%2F022624-ff-anualreport.b5deae8f.png&w=8&q=70',
    blurWidth: 8,
    blurHeight: 4
  },
  alt: '',
  sizes: '(min-width: 640px) 710px, (min-width: 768px) 980px, (min-width: 1024px) 480px, 100vw'
}
```

Ah so this should actually be `image.data`...

src/app/_components/SmartImage.tsx
barbaraperic: Basically makes sure it runs server side. For example, in `GovernanceCalendarCard` (`use client` component) we import `Card` component which uses `SmartImage`. if we don't explicilty write `use server` in the `SmartImage` component, it will run on the client, and in that case we get an error because we use `fs` which can't run in the browser.

barbaraperic:
> PS: Should statically imported images be part of the smart image component? Since we already know these are valid, it might not be necessary.

Good point! Think we can omit that, and leave validation for remote and asset images 👍

src/app/_components/SmartImage.tsx
barbaraperic: Forgot to delete, thanks :)

src/app/_components/SmartImage.tsx
barbaraperic: I see what you mean, personally I kind of like the separation of logic (`getImageSrc`) and ui rendering (`SmartImage`). That said, I’m happy to explore versions if you think this doesn’t work well for the current context.

src/app/_components/SmartImage.tsx
```diff
@@ -0,0 +1,86 @@
...
+  src,
+  alt = '',
```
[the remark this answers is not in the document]
barbaraperic: Good catch! It shouldn't be `undefined` :)

barbaraperic:
> PS: Should statically imported images be part of the smart image component? Since we already know these are valid, it might not be necessary.

Yes, good observation!

src/app/_components/SmartImage.tsx
CharlyMartin: Components run server-side by default right? Unless marked as `'use client'`. So I think in this case we don't need to add this annotation.

The `use server` directive is reserved for functions and it makes them server actions: https://nextjs.org/docs/app/api-reference/directives/use-server

CharlyMartin: Does that make sense?

src/app/_components/SmartImage.tsx
```diff
@@ -0,0 +1,86 @@
...
+import type { ImageObjectFit, StaticImageProps } from '@/types/imageType'
+
+import { graphicsData } from '@/data/graphicsData'
+
+const { data: fallbackSrc, alt: fallbackAlt } = graphicsData.imageFallback
+
+export type SmartImageProps = {
+  src?: string | StaticImageProps['data']
+  alt?: string
+  className?: string
```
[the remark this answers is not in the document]
CharlyMartin: It was omitted from `ImageProps`, so I removed the omission so that we don't have to add it to the prop type ourselves :)

src/app/_components/SmartImage.tsx
CharlyMartin: Yes but they are never passed to the Image component, they exist to style the image. Yhey should not belong to `SmartImage` in my opinion.

src/app/_components/SmartImage.tsx
mirhamasala: We can't call `fs` on client side. But instead of adding `'use server'` to the component itself, we would need to create a separate server action file marked with `'use server'` that contains the `fs` operations, then import and call this action from `SmartImage`. 

Something like

```ts
// actions.ts
'use server'
 
export async function readFileContents(path: string) {
  const fs = require('fs')
  return fs.readFileSync(path, 'utf8')
}

// ClientComponent.tsx
'use client'
 
import { readFileContents } from './actions'
 
export function ClientComponent() {
  return (
    <button onClick={() => readFileContents('/some/path')}>
      Read File
    </button>
  )
}
```

// ServerComponent.tsx
// No directive needed - server component by default
import { readFileSync } from 'fs'

export function ServerComponent() {
  const contents = readFileSync('/some/path', 'utf8')
  return <div>{contents}</div>
}


@CharlyMartin is right .. these are two different things `use client` and `use server`.

However, @CharlyMartin ~ This might be the reason production is failing at the moment:

CleanShot 2025-01-03 at 04 59 43@2x

mirhamasala:
Closing in favor of #979

PR #933, opened 2024-12-18
[UXIT-1906] Event Page · Add optional kicker and tag to schedule section

barbaraperic:
## 📝 Description

**Type:** New feature

This PR introduces `kicker` and `tag` field in Events page Schedule section.

Screenshot 2024-12-18 at 11 46 45

### DIFF
--- a/src/app/events/[slug]/components/ScheduleSection/ScheduleSection.tsx
+++ b/src/app/events/[slug]/components/ScheduleSection/ScheduleSection.tsx
@@ -10,7 +10,7 @@ type ScheduleSectionProps = {
 
 export function ScheduleSection({ schedule }: ScheduleSectionProps) {
   return (
-    <PageSection kicker="Join Us" title={schedule.title || 'Schedule'}>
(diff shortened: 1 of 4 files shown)


src/app/events/[slug]/components/ScheduleSection/ScheduleSection.tsx
```diff
@@ -10,7 +10,7 @@ type ScheduleSectionProps = {
 
 export function ScheduleSection({ schedule }: ScheduleSectionProps) {
   return (
-    <PageSection kicker="Join Us" title={schedule.title || 'Schedule'}>
+    <PageSection kicker={schedule.kicker} title={schedule.title}>
       <Tabs schedule={schedule} />
```
Comment (mirhamasala) [inline src/app/events/[slug]/components/ScheduleSection/ScheduleSection.tsx:14]: Something is wrong here from before (nothing to do with your implementation) - We are passing the entire `schedule` object - Need a bit more time to look into it - Will open a new ticket for it!

mirhamasala (approved):
Naise! ❤️ The missing bit of the puzzle. Let's go! 🙌🏼

PR #934, opened 2024-12-18
[UXIT-1891] Move FeaturedGrantGraduates component to grants directory

barbaraperic:
## 📝 Description

**Type:** Refactor

Refactored featured entries implementation.


mirhamasala (commented):
I pushed a naming nit. Also, I was wrong in the initial spec - We can't ensure that `FeaturedGrantGraduates` won't be empty.

I would like to keep this open still - Because I want to walk through some typing questions with @CharlyMartin. 🙏🏼

mirhamasala (approved):
LGTM! ☀️

PR #937, opened 2024-12-18
[UXIT-1795] - Flatten Subcategories

CharlyMartin:
## 📝 Description

This PR removes the temporary Zod transform function `temporaryForBackwardCompatibility` and updates the codebase to align with the new schema, where `category` is now a singular value.

## 🛠️ Key Changes

- Removed `temporaryForBackwardCompatibility`.
- Updated functions in the ecosystem-explorer project form.
- Removed filtering logic from `useEcosystemCategory`  - now `useEcosystemCategoryTree` - in favor of `useFilter`.
- Reworked `useFilter` to handle arrays of query values instead of only strings.

## 🧪 How to Test

- Verify that filtering works correctly on all pages where entries are filtered.

### DIFF
--- a/src/app/ecosystem-explorer/page.tsx
+++ b/src/app/ecosystem-explorer/page.tsx
@@ -5,17 +5,20 @@ import { BookOpen } from '@phosphor-icons/react/dist/ssr'
 import type { NextServerSearchParams } from '@/types/searchParams'
 
 import { PATHS } from '@/constants/paths'
+import { CATEGORY_KEY } from '@/constants/searchParams'
 
 import { graphicsData } from '@/data/graphicsData'
 
 import { buildImageSizeProp } from '@/utils/buildImageSizeProp'
 import { createMetadata } from '@/utils/createMetadata'
+import { entryMatchesSubcategoryQuery } from '@/utils/filterUtils'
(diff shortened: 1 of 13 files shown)


src/app/ecosystem-explorer/page.tsx
```diff
@@ -155,13 +164,13 @@ export default function EcosystemExplorer({ searchParams }: Props) {
                         title,
                         description,
                         image,
-                        subcategories: [categoryId],
+                        subcategory: subcategoryId,
                       } = project
 
                       const isFirstTwoImages = i < 2
                       const category = findOrThrow(
```
mirhamasala, line 171: I think we should rename:

```suggestion
                      const subcategory = findOrThrow(
```

I know we at some point discussed that this should be `category` but since we'll effectively still have the parent category, I think it's better we stick to `subcategory` (also in the CMS) - Does this make sense?

mirhamasala (approved):
This is amazing, @CharlyMartin! 🤩 Can't wait to see it all together. 🙌🏼 (Left a tiny comment.)

src/app/ecosystem-explorer/page.tsx
CharlyMartin: As discussed, we will keep this as `category` for now.

PR #938, opened 2024-12-18
[UXIT-1709] Refactor Touch Target

barbaraperic:
## 📝 Description

**Type:** Refactor

Improve touch target implementation across components by:
- Standardising touch target values: `touchTargetPadding`, `touchTargetOffset`, `visibleElementSize`
- Creating a new `TouchTarget` type definition


mirhamasala:
@barbaraperic (CC: @CharlyMartin)

Not sure this is exactly what we had in mind. The issue with this implementation is that the util function simply returns the config for a given key. It's a bit rigid and the naming tightly couple this util to specific UI components.

I think we are looking for something that help us enforce consistent sizing based on the WCAG guidelines (min. 44x44px touch targets) while remaining flexible for edge cases and make it easier to maintain and update touch target size globally.

For example, (not verified Claude suggestion), more something like:

```
import { twMerge } from 'tailwind-merge'

type TouchTargetSize = 'small' | 'medium' | 'large'
type TouchTargetShape = 'square' | 'circle' | 'pill'

interface TouchTargetOptions {
  size?: TouchTargetSize
  shape?: TouchTargetShape
  disabled?: boolean
  customPadding?: string
  customOffset?: string
  className?: string
}

const TARGET_SIZES = {
  small: {
    padding: 'p-2', // 16px touch target
    visibleSize: 'size-4', // 16px visible element
  },
  medium: {
    padding: 'p-3', // 24px touch target
    visibleSize: 'size-5', // 20px visible element
  },
  large: {
    padding: 'p-4', // 32px touch target
    visibleSize: 'size-6', // 24px visible element
  },
} as const

const TARGET_SHAPES = {
  square: '',
  circle: 'rounded-full',
  pill: 'rounded-full',
} as const

/**
 * Creates standardized touch target styles following WCAG guidelines
 * @param options Configuration options for the touch target
 * @returns Object containing className strings for the wrapper and visible element
 */
export function createTouchTarget(options: TouchTargetOptions = {}) {
  const {
    size = 'medium',
    shape = 'square',
    disabled = false,
    customPadding,
    customOffset,
    className = '',
  } = options

  const baseSize = TARGET_SIZES[size]
  const shapeClass = TARGET_SHAPES[shape]

  // Wrapper contains the full touch target area
  const wrapperClasses = twMerge(
    'relative',
    customPadding || baseSize.padding,
    customOffset && `-m-${customOffset}`,
    disabled && 'pointer-events-none opacity-50',
    className
  )

  // Inner element contains the visible content
  const innerClasses = twMerge(
    'relative',
    baseSize.visibleSize,
    shapeClass,
    'flex items-center justify-center'
  )

  return {
    wrapperClassName: wrapperClasses,
    innerClassName: innerClasses,
  }
}

/**
 * Helper to validate if a touch target meets minimum size requirements
 * @param element HTMLElement to check
 * @returns boolean indicating if element meets size requirements
 */
export function isValidTouchTarget(element: HTMLElement): boolean {
  const rect = element.getBoundingClientRect()
  const minSize = 44 // WCAG recommendation in pixels
  
  return rect.width >= minSize && rect.height >= minSize
}

/**
 * Example usage:
 * 
 * function Button({ children, ...props }) {
 *   const { wrapperClassName, innerClassName } = createTouchTarget({
 *     size: 'medium',
 *     shape: 'circle'
 *   })
 * 
 *   return (
 *     <div className={wrapperClassName}>
 *       <button className={innerClassName} {...props}>
 *         {children}
 *       </button>
 *     </div>
 *   )
 * }
 */
 ```

We can discuss it further today. 🙏🏼

**Review by mirhamasala — CHANGES_REQUESTED** [review]
Let's discuss later today! 🙏🏼

File: src/app/_components/Navigation/DesktopNavigation/DesktopNavigation.tsx [hunk 1]
```diff
@@ -44,10 +38,7 @@ export function DesktopNavigation() {
   return (
     <ul
       aria-label="Navigation items"
-      className={clsx(
-        'relative z-10 hidden lg:flex lg:items-center lg:gap-0.5',
-        TOUCH_TARGET_MAIN_NAV_ITEM.offsetWidth,
-      )}
+      className="relative z-10 hidden lg:flex lg:items-center lg:gap-0.5"
```
barbaraperic, line 41: I removed the offset as it had no effect on the element

src/app/_components/Table/TooltipHeader.tsx
```diff
@@ -31,20 +35,20 @@ export function TooltipHeader({
...
-                TOUCH_TARGET.iconWrapperPadding,
+                TOUCH_TARGET_ICON.touchAreaPadding,
                 'flex items-center justify-center rounded-full transition ease-in-out',
                 open
                   ? 'bg-brand-600/70 text-brand-200'
                   : 'text-brand-400 group-hover:bg-brand-700 group-hover:text-brand-300',
               )}
             >
-              <Icon component={Question} size={TOUCH_TARGET.iconSize} />
+              <Icon component={Question} size={20} />
```
barbaraperic, line 51: This could be moved to TOUCH_TARGET_ICON as `visibleElementSize` but we usually directly pass the value to the component, so I kept it this way. Let me know if you want it as `visibleElementSize` in TOUCH_TARGET_ICON 🙂

mirhamasala: Yes! I think we can use `visibleElementSize` in this case. Thank you.

src/app/_components/Table/TooltipHeader.tsx
```diff
@@ -10,12 +12,15 @@ type TableHeaderWithTooltipProps = {
...
-} as const
+const TOUCH_TARGET: TouchTarget = {
+  touchAreaPadding: 'p-2',
+  touchAreaOffset: '-m-2',
+}
+
+const TOUCH_TARGET_ICON: TouchTarget = {
+  visibleElementSize: 20,
+  touchAreaPadding: 'p-1.5',
+}
```
mirhamasala, line 23: What about something like:

```suggestion
const TOUCH_TARGET_ICON: TouchTarget & { visibleElementSize: number } = {
  visibleElementSize: 20,
  touchAreaPadding: 'p-1.5',
};
```

mirhamasala (approved):
Almost there – Just a nit or two! 🙌🏼 Thank you for reworking this! ✨

src/app/_components/Navigation/DesktopNavigation/DesktopNavigation.tsx
barbaraperic: ah right! you think we could include it directly in the class? or prefer to be in the TOUCH_TARGET const?

src/app/_components/Table/TooltipHeader.tsx
barbaraperic: love it!

src/app/_types/touchTargetType.ts
```diff
@@ -0,0 +1,5 @@
+export type TouchTarget = {
+  visibleElementSize?: string | number
```
mirhamasala, line 2: Is it ever a number?

barbaraperic: Icon requires size to be a `number`

mirhamasala (approved):
LGTM! ❤️

PR #951, opened 2024-12-23
[UXIT-1891] Refactor Featured Entries: move logic to parent component [skip percy]

barbaraperic:
## Description

**Type**: Refactor

This PR refactors the featured entries components by moving the logic to the parent component. 

The empty state checks were removed from `FeaturedEcosystemProjects` and `FeaturedGrantGraduates`  where Zod schemas guarantee non-empty arrays.


src/app/(homepage)/page.tsx
```diff
@@ -138,7 +147,11 @@ export default function Home() {
           title="News & Blog"
           description="The latest updates and announcements from the Filecoin ecosystem and Filecoin Foundation."
         >
-          <FeaturedBlogPosts />
+          {hasFeaturedBlogPosts ? (
```
mirhamasala, line 150: We probably shouldn't show the entire section if there are no featuredBlogPosts..

src/app/grants/page.tsx
```diff
@@ -101,11 +101,9 @@ export default function Grants() {
         </CardGrid>
       </PageSection>
 
-      {grantGraduates.length > 0 && (
-        <PageSection kicker="Past Recipients" title="Grant Graduates">
-          <FeaturedGrantGraduates grantGraduates={grantGraduates} />
-        </PageSection>
-      )}
+      <PageSection kicker="Past Recipients" title="Grant Graduates">
```
mirhamasala, line 104: We can't guarantee these exist - Since filter can return an empty arran. 

```ts
const grantGraduatesSlugs = featuredGrantGraduatePaths.map(
  extractSlugFromFilename,
)

const grantGraduates = ecosystemProjects.filter((item) =>
  grantGraduatesSlugs?.includes(item.slug),
)
```

Try changing the type here to be a none-empty arre and it will be clear what I mean.

```ts
type FeaturedGrantsGraduatesProps = {
  grantGraduates: Array<EcosystemProject>
}
```

src/app/(homepage)/page.tsx
mirhamasala: Same comment [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/951/files#r1896524926)

mirhamasala (changes requested):
Left a few comments 🙏🏼

src/app/grants/page.tsx
barbaraperic: Ah I see.

Tho Zod will scream at us if the `grantsGraduates` are empty, it could happen somewhere in the process that we pass an empty array.

Screenshot 2024-12-24 at 09 22 51


Thank you, I added the condition for the whole section and non empty type for `FeaturedGrantGraduates` 👍

src/app/ecosystem-explorer/utils/createFeaturedProjectsZodField.ts
```diff
@@ -0,0 +1,20 @@
...
+  return z
+    .array(MarkdownPathSchema)
+    .transform((paths) => {
+      const ecosystemProjects = getEcosystemProjectsData()
+      const slugs = paths.map(extractSlugFromFilename)
+      return ecosystemProjects.filter((item) => slugs.includes(item.slug))
+    })
+    .refine((filtered) => filtered.length > 0, {
+      message: errorMessage,
+    })
```
barbaraperic, line 19: Think I might be missing something - what is the case when we would see the custom error message?

If the path is incorrect or file is missing, it would be caught by `MarkdownPathSchema`?

src/app/ecosystem-explorer/utils/createFeaturedProjectsZodField.ts
```diff
@@ -0,0 +1,20 @@
+import { z } from 'zod'
+
+import { extractSlugFromFilename } from '@/utils/fileUtils'
+
+import { MarkdownPathSchema } from '@/schemas/FrontmatterSchema'
+
+import { getEcosystemProjectsData } from './getEcosystemProjectData'
+
+export function createFeaturedProjectsZodField(errorMessage: string) {
```
barbaraperic, line 9: 👏

src/app/ecosystem-explorer/utils/createFeaturedProjectsZodField.ts
mirhamasala: > Think I might be missing something - what is the case when we would see the custom error message?

Empty array..

barbaraperic: Ah I commented out and it left an ","  in the array so it didn't show the right message. Sorry!

PR #955, opened 2024-12-26
Upgrade to Next.js 15 [skip percy]

mirhamasala:
## 📝 Description

This PR upgrades the application to Next.js 15 and React 19 and refactors the `Blog`, `Event`, and `Ecosystem Explorer` pages to comply with the new async request API. The main motivation for this change is the introduction of asynchronous handling of `searchParams`, which now requires server-side resolution (`await props.searchParams`). 

To accommodate this, we split server-side logic (e.g., data fetching and metadata generation) from client-side interactions (e.g., filtering and pagination). This ensures compatibility with Next.js 15 while maintaining a modular and maintainable architecture.

Additionally, we have applied the `next-async-request-api` codemod for API updates, and included some other minor 

- **Move file tracing config out of experimental.**
- **Uninstall Sharp.**
- **Change global CSS extension.**

## 🔖 Resources

- [Next.js 15 - Changelog](https://nextjs.org/blog/next-15)


src/app/blog/[slug]/page.tsx
```diff
@@ -33,7 +34,8 @@ export function generateMetadata({ params }: BlogPostProps) {
   })
 }
 
-export default function BlogPost({ params }: BlogPostProps) {
+export default async function BlogPost(props: BlogPostProps) {
+  const params = await props.params
```
Comment (CharlyMartin) [inline src/app/blog/[slug]/page.tsx:38]: Nit:

```suggestion
  const { slug } = await props.params
```

src/app/blog/page.tsx
```diff
@@ -1,70 +1,36 @@
...
 import { getMetaData } from './utils/getMetaData'
 
-const NoSSRPagination = dynamic(
-  () => import('@/components/Pagination').then((module) => module.Pagination),
-  { ssr: false },
-)
-
 type Props = {
-  searchParams: NextServerSearchParams
+  searchParams: Promise<AsyncNextServerSearchParams>
```
CharlyMartin, line 26: `AsyncNextServerSearchParams` already includes a Promise no?


```suggestion
  searchParams: AsyncNextServerSearchParams
```

src/app/blog/components/BlogContent.tsx
```diff
@@ -0,0 +1,173 @@
+'use client'
+
```
CharlyMartin, line 2: I think this can stay a server component no?

```suggestion
```

src/app/ecosystem-explorer/page.tsx
```diff
@@ -73,40 +41,16 @@ export const metadata = createMetadata({
 })
 
 const categories = getEcosystemCMSCategories()
+const ecosystemProjects = getEcosystemProjectsData()
```
CharlyMartin, line 44: Why not calling this function in `EcosystemExplorerContent`? Since we only call it here to pass is as a prop.

If so, we could do the same for events and projects.

CharlyMartin (commented):
I did a first pass after reading the changelog, and added some comments. I will have another look Thursday.

Looking forward to merging this one to get some of the caching goodness :)

CharlyMartin:
Also, should we move `BreakpointDebugger` to the other side? The bottom left spot is getting crowded 😅

CleanShot 2024-12-31 at 17 45 09

src/app/ecosystem-explorer/page.tsx
mirhamasala: Yes, in this case, it's possible.

For the blog and events, not, because we use it to get the featured post and event.

mirhamasala:
Thank you, @CharlyMartin. 🙏🏼 Apologies for the review nits—there were a few rounds of changes involved. 🙈 I appreciate you catching them. All CR suggestions have been implemented.

src/app/ecosystem-explorer/page.tsx
CharlyMartin: On the events page, we only use the result of `getEventsData` in `EventsContent`, so in that case, we could call it directly in `EventsContent`.

On the blog, not.

When we implement [data caching](https://nextjs.org/docs/app/building-your-application/data-fetching/fetching#caching-data-with-an-orm-or-database), we will be able to call a function everywhere we need it, rather than calling it once and passing props. Next.js will hit the cache instead of calling the function twice.

CharlyMartin (approved):

PR #965, opened 2024-12-27
[UXIT-1932] Add Dependabot configuration [skip percy]

mirhamasala:
Adds automated dependency management with the following features:
- Weekly updates for both npm packages and GitHub Actions
- Groups all npm dependencies into a single PR to reduce noise
- Groups all GitHub Actions updates into a single PR 
- Uses conservative "auto" versioning strategy for npm updates
- Runs weekly to maintain a balance between staying updated and PR volume

Configuration includes:
- Automated updates for npm dependencies in root directory
- Automated updates for GitHub Actions workflows
- Weekly schedule for both ecosystems
- Grouped updates to minimize PR noise


.github/dependabot.yml
```diff
@@ -0,0 +1,29 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
```
CharlyMartin, line 7: I'm not too familiair with dependabot setup but if I understand correctly, this only checks the `package.json` in the root directory right? The one Turbo uses.

Should we add a section for the Next app too?

```yml
  - package-ecosystem: "npm"
    directory: "/apps/site"
    schedule:
      interval: "weekly"
    groups:
      site-dependencies:
        patterns:
          - "*"
```

mirhamasala: Totally! ❤️

PR #969, opened 2024-12-30
[UXIT-1933] Group API routes using Next.js route groups [skip percy]

mirhamasala:
Move API routes under (api) folder to leverage Next.js route groups for better code organization while maintaining public accessibility.


mirhamasala:
Closing:  Implementation still breaks the public routes - Will revisit when we upgrade to Next.js 15

PR #974, opened 2025-01-06
[UXIT-1854] Adapt checkbox styles

barbaraperic:
## 📝 Description

This PR includes new checkbox styles - [default / hover / focus / checked](https://www.figma.com/design/0pBfQs5tKI6bx5ypPIfVRA/FIL.org-Prototypes?node-id=75-4945&t=8W6MSd4F2hwV25AW-0)

**Type:** Bug fix

## 📸 Screenshots

https://github.com/user-attachments/assets/bbd905b9-b090-4c1f-acd2-c69bb35011a8

### DIFF
--- a/apps/site/src/app/_components/Form/FormCheckbox.tsx
+++ b/apps/site/src/app/_components/Form/FormCheckbox.tsx
@@ -37,7 +37,7 @@ export function FormCheckbox({
         <div
           className={clsx(
             TOUCH_TARGET.visibleElementSize,
-            'rounded bg-white p-0.5 text-white hover:bg-brand-200 focus:bg-brand-200 group-focus:outline group-focus:outline-1 group-focus:outline-offset-2 group-focus:outline-white group-data-[disabled]:cursor-not-allowed group-data-[checked]:bg-brand-400 hover:group-data-[checked]:bg-brand-500',
+            'rounded bg-brand-100 p-0.5 text-brand-100 hover:bg-brand-200 group-focus:bg-brand-200 group-focus:outline group-focus:outline-1 group-focus:outline-offset-2 group-focus:outline-brand-100 group-data-[disabled]:cursor-not-allowed group-data-[checked]:bg-brand-400 hover:group-data-[checked]:bg-brand-500',
           )}
         >
           <span className="hidden group-data-[checked]:block">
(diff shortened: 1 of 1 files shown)


apps/site/src/app/_components/Form/FormCheckbox.tsx
```diff
@@ -37,7 +37,7 @@ export function FormCheckbox({
         <div
           className={clsx(
             TOUCH_TARGET.visibleElementSize,
-            'rounded bg-brand-100 p-0.5 text-brand-100 group-focus:outline group-focus:outline-1 group-focus:outline-offset-2 group-focus:outline-brand-100 group-data-[disabled]:cursor-not-allowed group-data-[checked]:bg-brand-400',
+            'rounded bg-white p-0.5 text-white hover:bg-brand-200 focus:bg-brand-200 group-focus:outline group-focus:outline-1 group-focus:outline-offset-2 group-focus:outline-white group-data-[disabled]:cursor-not-allowed group-data-[checked]:bg-brand-400 hover:group-data-[checked]:bg-brand-500',
```
mirhamasala, line 40: Can we check with @filipagr if we really want to use `white` here? This might be unintentional since we tend to use `brand-100`. 🙏🏼🥹

CharlyMartin: Figma says this should have a white background, so it is technically correct. But Mirha has a point, we always use brand-100 instead of pure white.

@filipagr which one is it?? 😁

CharlyMartin (changes requested):
@barbaraperic I think it's just missing this [state](https://www.figma.com/design/0pBfQs5tKI6bx5ypPIfVRA/FIL.org-Prototypes?node-id=75-4950&t=D3ZTKSJace7lOkHq-4): focus state when unchecked.

Figma says the background should be light blue, and currently it's white.

CleanShot 2025-01-06 at 15 49 48@2x

apps/site/src/app/_components/Form/FormCheckbox.tsx
filipagr: If we always use `brand-100`, then it's better to use it 🙏🏻 Thanks for catching it!

CharlyMartin (approved):
☀️

PR #977, opened 2025-01-06
[UXIT-1963] Events Page · Fix CMS Region Field [skip percy]

mirhamasala:
## 📝 Description

Region should be optional so that we can also handle Virtual events per hint: `"Select the region where the in-person event takes place. Leave empty for virtual events."`

- **Type:** Bug fix 

## 📸 Screenshots

BUG

CleanShot 2025-01-06 at 12 38 43@2x

### DIFF
--- a/apps/site/src/app/_utils/filterUtils.ts
+++ b/apps/site/src/app/_utils/filterUtils.ts
@@ -1,14 +1,12 @@
 import { VIRTUAL_EVENT_FILTER_OPTION } from '@/events/constants/constants'
+import type { Location } from '@/events/schemas/LocationSchema'
 
 type WithCategory = {
   category: string
 }
 
 type WithLocation = {
-  location: {
-    primary: string
-    region?: string | null
-  }
+  location: Location
 }
 
 export function entryMatchesCategoryQuery<Entry extends WithCategory>(
--- a/apps/site/src/app/_utils/filterUtils.ts
+++ b/apps/site/src/app/_utils/filterUtils.ts
@@ -7,7 +7,7 @@ type WithCategory = {
 type WithLocation = {
   location: {
     primary: string
-    region?: string
+    region?: string | null
   }
 }
 
--- a/apps/site/src/app/events/schemas/LocationSchema.ts
+++ b/apps/site/src/app/events/schemas/LocationSchema.ts
@@ -35,7 +35,7 @@ function isVirtual(location: Location) {
 }
 
 function hasRegion(location: Location) {
-  return location.region !== undefined && location.region !== null
+  return Boolean(location.region)
 }
 
 function virtualEventHasNoRegion(location: Location) {
--- a/apps/site/src/app/events/schemas/LocationSchema.ts
+++ b/apps/site/src/app/events/schemas/LocationSchema.ts
@@ -3,7 +3,7 @@ import { z } from 'zod'
 import { getCMSFieldOptionsAndValidIds } from '@/utils/getCMSFieldOptionsAndValidIds'
 import { createEnumSchema } from '@/utils/zodUtils'
 
-type Location = {
+export type Location = {
   primary: string
   region?: string | null
 }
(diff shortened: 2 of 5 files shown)


barbaraperic:
I believe we also need to handle the case:
```ts
location:
  primary: Virtual
  region: null
```

This will happen when we modify from non-virtual to virtual event. 


```ts

type Location = {
  primary: string
  region?: string | null
}

export const LocationSchema = z
  .object({
    primary: z.string(),
    region: RegionSchema.nullable().optional(),
  })

function hasRegion(location: Location): boolean {
  return location.region !== undefined && location.region !== null
}
```

mirhamasala:
> I believe we also need to handle the case:
> 
> ```ts
> location:
>   primary: Virtual
>   region: null
> ```
> 
> This will happen when we modify from non-virtual to virtual event.
> 
> ```ts
> type Location = {
>   primary: string
>   region?: string | null
> }
> 
> export const LocationSchema = z
>   .object({
>     primary: z.string(),
>     region: RegionSchema.nullable().optional(),
>   })
> 
> function hasRegion(location: Location): boolean {
>   return location.region !== undefined && location.region !== null
> }
> ```

Oh, good catch! ❤️ That's probably the case for all optional fields within an object then. We might need to add a ticket to check! 🙏🏼

barbaraperic (approved):

apps/site/src/app/_utils/filterUtils.ts
```diff
@@ -7,7 +7,7 @@ type WithCategory = {
 type WithLocation = {
```
mirhamasala, line 7: @CharlyMartin - Is there a reason why we are not deriving this type from the `LocationSchema` (`apps/site/src/app/events/schemas/LocationSchema.ts`)?

CharlyMartin: These filter functions, `entryMatchesCategoryQuery` & `entryMatchesLocationQuery`, should be quite generic since they can filter any object that matches the type signature.

It works well for categories since it's just a string, and multiple resources have categories.

But I suppose for `location` we can reuse the type since it's only used for events and is quite specific.

```ts
// apps/site/src/app/events/schemas/LocationSchema.ts
export type Location = {
  primary: string
  region?: string | null
}

// apps/site/src/app/_utils/filterUtils.ts
type WithLocation = {
  location: Location
}

export function entryMatchesLocationQuery<Entry extends WithLocation>(
  entry: Entry,
  query?: string,
) {
  if (query === VIRTUAL_EVENT_FILTER_OPTION.id) {
    return entry.location.primary === VIRTUAL_EVENT_FILTER_OPTION.name
  }

  return entry.location.region === query
}
```

apps/site/src/app/events/schemas/LocationSchema.ts
```diff
@@ -18,7 +18,7 @@ const RegionSchema = createEnumSchema(validRegionIds)
 export const LocationSchema = z
   .object({
     primary: z.string(),
-    region: RegionSchema.optional(),
+    region: RegionSchema.nullable().optional(),
```
CharlyMartin, line 21: Not sure if that reads better, but this is equivalent to [nullish](https://zod.dev/?id=nullish)

```suggestion
    region: RegionSchema.nullish(),
```

CharlyMartin:
Reviewed ✨

Also, when I start the CMS locally, I see no entries 🥲 Anyone had the issue / fixed it?

CleanShot 2025-01-07 at 10 08 15

mirhamasala:
@CharlyMartin 

>Also, when I start the CMS locally, I see no entries 🥲 Anyone had the issue / fixed it?

I don't have this issue..

barbaraperic:
> Reviewed ✨
> 
> Also, when I start the CMS locally, I see no entries 🥲 Anyone had the issue / fixed it?
> 
> CleanShot 2025-01-07 at 10 08 15

I had it! It could be that you ran the server from filecoin-foundation/

> Reviewed ✨
> 
> Also, when I start the CMS locally, I see no entries 🥲 Anyone had the issue / fixed it?
> 
> CleanShot 2025-01-07 at 10 08 15

I had it! My issue was that I was running the server on `filecoin-foundation/apps/site` and it should have been on `filecoin-foundation`

PR #978, opened 2025-01-06
[UXIT-1909] Include linter for ordering TS props [skip percy]

barbaraperic:
## 📝 Description

**Type:** Bug fix 

This PR implements TypeScript interface property sorting through ESLint configuration - `typescript-sort-keys`

`"asc" `- enforce properties to be in ascending order.
`"caseSensitive": false` - enforce properties to be in non case-sensitive order. 
`"requiredFirst": true` - enforce optional properties to come after required ones.

[Documentation
](https://github.com/infctr/eslint-plugin-typescript-sort-keys/blob/master/docs/rules/interface.md)


mirhamasala (changes requested):
Can we look into switching from `eslint-plugin-typescript-sort-keys` to `@typescript-eslint/member-ordering` for our prop sorting. I think there are a few advantages:

It's maintained directly by the TS-ESLint team, so we get better integration with the rest of the TypeScript ESLint ecosystem and more regular updates.

The member-ordering rule is way more flexible - we can create custom orderings that make more sense for React props. For example, we can:

- Put required props first
- Group React-specific props like 'children' and 'className'
- Put all event handlers/callbacks at the end
- Mix alphabetical ordering within groups

And since it's part of the core @typescript-eslint plugin that we're already using, it reduces our dependency overhead.

Something like:

```js
// .eslintrc.js
module.exports = {
  plugins: ['@typescript-eslint'],
  rules: {
    '@typescript-eslint/member-ordering': [
      'error',
      {
        default: {
          memberTypes: [
            // React specific props first
            'signature',
            'public-static-field', // static fields

            // Required props first
            'required-field',
            
            // Common React props
            ['field', { name: 'children' }],
            ['field', { name: 'className' }],
            ['field', { name: 'style' }],
            
            // Props by type
            ['field', { type: 'string' }],
            ['field', { type: 'number' }],
            ['field', { type: 'boolean' }],
            
            // Objects and arrays
            ['field', { type: 'object' }],
            ['field', { type: 'array' }],

            // Event handlers and callbacks last
            ['field', { filter: 'on[A-Z].*' }], // for onClick, onChange etc
            ['field', { suffix: 'Callback' }],
            'method'
          ],
          order: 'alphabetically'
        }
      }
    ]
  }
};
```

What do you think? Also - thinking about this - It might be helpful to research/Claude a best practice order in React projects?

barbaraperic:
> Can we look into switching from `eslint-plugin-typescript-sort-keys` to `@typescript-eslint/member-ordering` (https://typescript-eslint.io/rules/member-ordering/) for our prop sorting. I think there are a few advantages:
> 
> It's maintained directly by the TS-ESLint team, so we get better integration with the rest of the TypeScript ESLint ecosystem and more regular updates.
> 
> The member-ordering rule is way more flexible - we can create custom orderings that make more sense for React props. For example, we can:
> 
> * Put required props first
> * Group React-specific props like 'children' and 'className'
> * Put all event handlers/callbacks at the end
> * Mix alphabetical ordering within groups
> 
> And since it's part of the core @typescript-eslint plugin that we're already using, it reduces our dependency overhead.
> 
> Something like:
> 
> ```js
> // .eslintrc.js
> module.exports = {
>   plugins: ['@typescript-eslint'],
>   rules: {
>     '@typescript-eslint/member-ordering': [
>       'error',
>       {
>         default: {
>           memberTypes: [
>             // React specific props first
>             'signature',
>             'public-static-field', // static fields
> 
>             // Required props first
>             'required-field',
>             
>             // Common React props
>             ['field', { name: 'children' }],
>             ['field', { name: 'className' }],
>             ['field', { name: 'style' }],
>             
>             // Props by type
>             ['field', { type: 'string' }],
>             ['field', { type: 'number' }],
>             ['field', { type: 'boolean' }],
>             
>             // Objects and arrays
>             ['field', { type: 'object' }],
>             ['field', { type: 'array' }],
> 
>             // Event handlers and callbacks last
>             ['field', { filter: 'on[A-Z].*' }], // for onClick, onChange etc
>             ['field', { suffix: 'Callback' }],
>             'method'
>           ],
>           order: 'alphabetically'
>         }
>       }
>     ]
>   }
> };
> ```
> 
> What do you think? Also - thinking about this - It might be helpful to research/Claude best practices for ordering in React projects?

@typescript-eslint/member-ordering sounds good! 😊 I've tried to make it work with your example, but no luck. [I couldn't find examples in the documentation](https://typescript-eslint.io/rules/member-ordering/)? But the alphabetical ordering and required first is working well. Claude has some opinions on best practice - similar what you posted, but I couldn't find anything specific in the [Typescript documentation](https://www.typescriptlang.org/). The closest example it [here](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html), they use required first and alphabetical. Let me know what you think! We can build something custom, it's an option.

PR #989, opened 2025-01-08
[UXIT-1974] - Update public folder path in config.yml

CharlyMartin:
## 📝 Description

This PR fixes the `public_folder` path in the CMS config. It needs to be relative to the base of the built site, not absolute.

CleanShot 2025-01-08 at 17 01 43@2x

https://decapcms.org/docs/configuration-options/


CharlyMartin:
Live tested with @mirhamasala

PR #1017, opened 2025-01-20
Refactor dependabot configuration for monorepo

CharlyMartin:
## 📝 Description

This PR groups the Dependabot updates in a single configuration, hoping it fixes the current problem:

To update a package with our new monorepo setup with workspaces, the command to run is `npm install <package-name> --workspace <workspace-name>` from the root. This updates the `package.json` in the relevant workspace, as well as the `package-lock.json` at the root.

Dependabot currently updates only the workspace `package.json` when bumping to newer versions. As a result, after merging its PRs, the next person to run `npm install` will see changes in the root `package-lock.json` since Dependabot doesn’t update this file.

This PR tries to solve this issue by pointing to both directories containing a manifest file. It is unclear whether it will solve the issue, most content on the topic doesn't provide a clear answer:
- https://github.com/dependabot/dependabot-core/issues/5226
- https://github.com/dependabot/dependabot-core/issues/6270

## 🔖 Resources

- https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#directories-or-directory--


barbaraperic (approved):
I guess we can test it when it gets merged right? The only question I have is why we changed it from weekly to daily? Not sure if you've seen this, but we can try it if for some reason this version fails - https://github.com/dependabot/dependabot-core/issues/5226#issuecomment-1179434437

.github/dependabot.yml
```diff
@@ -4,24 +4,18 @@
...
-      dependencies:
-        patterns:
-          - "*"
-    versioning-strategy: auto
-
-  - package-ecosystem: "npm"
-    directory: "/apps/site"
-    schedule:
-      interval: "weekly"
+      interval: "daily"
```
CharlyMartin, line 11:
```suggestion
      interval: "weekly"
```

.github/dependabot.yml
```diff
@@ -4,24 +4,18 @@
...
+      - "/apps/site"
     schedule:
       interval: "weekly"
     groups:
       dependencies:
         patterns:
           - "*"
     versioning-strategy: auto
+    commit-message:
+      prefix: "config-v2" # increment this number when we change the config
```
CharlyMartin, line 18:
```suggestion
```

PR #1026, opened 2025-01-22
Make useSort isomorphic

CharlyMartin:
> [!NOTE]
> This is PR 3/3 to make our hooks isomorphic, meaning they don't contain any server or client-specific code and can run in both environments. More information on the motivation is in the [Notion ticket](https://www.notion.so/filecoin/FF-Make-hooks-isomorphic-1827631f2825809083eee8c67df2660f?pvs=4). 

## 📝 Description

This PR is about making `useSort` isomorphic, as well as removing the explicit `defaultsTo` prop in favour of using the first item in the array of sort configuration

## 🛠️ Key Changes

- Remove `NextServerSearchParams` from `useSort` props and use `sortQuery` instead
- Implement changes on all pages
- Use the first sort configuration item as the default sort, instead of explicitly using a sort key. The first item is usually the default, so this duplication is unnecessary.

## 🧪 How to Test

Make sure the sort works the same as on fil.org

### DIFF
--- a/apps/site/src/app/_hooks/useSort.ts
+++ b/apps/site/src/app/_hooks/useSort.ts
@@ -1,18 +1,13 @@
 import { useMemo } from 'react'
 
-import type { NextServerSearchParams } from '@/types/searchParams'
 import type { SortConfig } from '@/types/sortTypes'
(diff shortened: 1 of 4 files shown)


barbaraperic (dismissed):
⭐️

PR #1038, opened 2025-01-27
Statically Cache Blog Page

ianconsolata:
## 📝 Description

This is a small refactor to ensure that the blog page (and the associated blog page data) is effectively cached. This forces the server component to always be statically cached, and return the statically cached page for all requests regardless of the search params.

In order to ensure search still works despite this caching, the blog section (the only bit that changes with search params) is moved to a client component, and rendered client side after the statically rendered page is returned. This seems to be working, though I cannot verify it's being cached by Vercel correctly without deployment.

It does still seem to be making server requests every time the search params are updated, which should be fixe at some point, but all those requests should be hitting the cached version of the page still, so should not be causing us to have serverless function invocations.

- **Type:** Bug fix

## 🛠️ Key Changes

- app/blog/page is now forced to be static
- blog search logic factored out into separate client component

## 🧪 How to Test

I am not sure how to effectively test. For now, I just verified that the functionality still works as expected on that page, which seems good enough to merge to production. Once merged, we should check the Vercel logs. If I am right about what this does, we should see a drastic reduction in serverless function calls on this route (perhaps even to 0), and we should see 100% caching of this page.

This will not affect the laggy search rendering, and more work will be needed to improve the performance there and determine


CharlyMartin:
Merged as part of https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1040

PR #1039, opened 2025-01-28
Revert "Add unstable_cache to getBlogPostsData [skip percy]"

mirhamasala:
Reverts FilecoinFoundationWeb/filecoin-foundation#1032


CharlyMartin:
This PR was reverted because `unstable_cache` didn’t reduce the number of function invocations, as we thought it would.

The only noticeable change was in the log labels:

Before implementing `unstable_cache` the label read `use cache endpoint ...`
CleanShot 2025-01-28 at 10 39 39

After implementing `unstable_cache`, it read `got fetch cache entry for ...`
CleanShot 2025-01-28 at 10 39 21

PR #1042, opened 2025-01-28
Remove force-static from Blog page [skip percy]

CharlyMartin:
## 📝 Description

This PR removes the `export const dynamic` and `export const revalidate` directives.

From Next's documentation
> The new model in the app directory favors granular caching control at the fetch request level over the binary all-or-nothing model of getServerSideProps and getStaticProps at the page-level in the pages directory. The dynamic option is a way to opt back in to the previous model as a convenience and provides a simpler migration path.

## 🔖 Resources

- https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#dynamic
- https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#revalidate

### DIFF
--- a/apps/site/src/app/blog/page.tsx
+++ b/apps/site/src/app/blog/page.tsx
@@ -1,3 +1,5 @@
+import { Suspense } from 'react'
+
 import { PATHS } from '@/constants/paths'
 
 import { graphicsData } from '@/data/graphicsData'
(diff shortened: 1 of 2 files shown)


CharlyMartin:
Live reviewed with @mirhamasala

PR #1043, opened 2025-01-28
[UXIT-1924] Include Markdown lint GitHub Action

barbaraperic:
## 📝 Description

Added markdown lint to site lint github action. 

- **Type:** New feature

`"default": true `- Enables all default markdownlint rules as a baseline
`"MD001": true` - Enforces heading levels to increment by only one level at a time (e.g., H1 → H2, not H1 → H3)
`"MD009": false` - Disables the rule preventing trailing spaces
`"MD013": false `- Disables the line length limit rule.
`"MD024": { "siblings_only": true }` - Allows duplicate headings, but only if they're not siblings (not at the same level in the same section)
`"MD025": true` - Prevents multiple top-level (h1) headings in the same document
`"MD029": { "style": "ordered" }` - Enforces ordered numbering style for ordered lists (1., 2., 3., etc.) instead of allowing all ones
`"MD033": false` - Disables the rule that prevents inline HTML. This allows you to use HTML tags in your markdown
`"MD034": false` - Disables the rule requiring bare URLs to be enclosed in angle brackets
`"MD036": false` - Disables the rule preventing emphasis (bold/italic) on headings
`"MD041": false `- Disables the rule requiring the first line to be a top-level heading
`"MD045": false` - Disables the rule requiring images to have alt text
`"MD046": false `- Disables the rule enforcing a consistent code block style

## 🔖 Resources

- [markdownlint-cli2-action](https://github.com/marketplace/actions/markdownlint-cli2-action)


apps/site/.markdownlint.json
```diff
@@ -0,0 +1,19 @@
+{
+  "default": true,
+  "MD001": true,
+  "MD009": false,
```
mirhamasala, line 4: Why do we want to disable this one? Is there a use case where this is necessary?

`"MD009": false - Disables the rule preventing trailing spaces`

apps/site/.markdownlint.json
```diff
@@ -0,0 +1,19 @@
...
+  "MD029": {
+    "style": "ordered"
+  },
+  "MD033": false,
+  "MD034": false,
+  "MD036": false,
+  "MD041": false,
+  "MD045": false,
+  "MD046": false
+}
```
mirhamasala, line 19: Is there another format than `json` that we could use for the config? So that we could add explanatory comments? 

```json
{
  "default": true,                    // Enable all rules by default
  "MD001": true,                      // Enforce header increment rules (e.g., H1 -> H2 -> H3)
  "MD009": false,                     // Allow trailing spaces
  "MD013": false,                     // Disable line length restrictions
  "MD024": {
    "siblings_only": true             // Allow duplicate headers if not siblings (same level)
  },
  "MD025": true,                      // Allow only one top-level header
  "MD029": {
    "style": "ordered"                // Enforce ordered list item prefix (1. 2. 3. etc)
  },
  "MD033": false,                     // Allow inline HTML
  "MD034": false,                     // Allow bare URLs
  "MD036": false,                     // Allow emphasis on words without spaces
  "MD041": false,                     // Allow content before first header
  "MD045": false,                     // Allow images without alt text
  "MD046": false                      // Allow different code block styles
}
```

mirhamasala (approved):
Marvellous. 🤩

apps/site/.markdownlint.json
barbaraperic: Yes, exactly. There are around 100 pages that have some space at the end of the line and it returns an error.

apps/site/.markdownlint.json
barbaraperic: Great idea!

mirhamasala (approved):
LGTM! ☀️☀️

PR #1046, opened 2025-01-29
[UXIT-2045] About Page · Update Reports section [skip percy]

mirhamasala:
Update Reports section to reflect latest annual reports.

BEFORE
CleanShot 2025-01-29 at 10 13 27@2x

AFTER
CleanShot 2025-01-29 at 10 25 20@2x

### DIFF
--- a/apps/site/src/app/about/page.tsx
+++ b/apps/site/src/app/about/page.tsx
@@ -50,7 +50,7 @@ export default function About() {
         description={header.description}
         image={graphicsData.about}
         cta={{
-          href: FILECOIN_FOUNDATION_URLS.annualReports['2023'],
+          href: FILECOIN_FOUNDATION_URLS.annualReports['2024'],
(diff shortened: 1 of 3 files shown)


apps/site/src/app/about/page.tsx
barbaraperic, line 1: We could use the `alt` that we have defined in reportsData?

```ts
<Card
    as="div"
    title={title}
    description={description}
    image={
    image && {
          ...(image || graphicsData.imageFallback.data),
           alt: image?.alt || 'Report Image',
          ...
 />
```

barbaraperic (approved):
Looks good! Left one comment

apps/site/src/app/about/page.tsx
mirhamasala: These images don't really convey anything more than the title already does? So, I'd say decorative?

PR #1054, opened 2025-01-29
[UXIT-2050] - Use async getAllMarkdownData function

CharlyMartin:
## 📝 Description

This PR implements async `getAllMarkdownData` in blog, digest, ecosystem projects and events. These pages are now async as well.

## 🛠️ Key Changes

- Removes `getAllMarkdownData` from `utils/getMarkdownData.ts`
- Renames `getAllMarkdownDataAsync` to `getAllMarkdownData`
- Introduces `getFeaturedEcosystemProjects` to replace `createFeaturedProjectsZodField`, as we cannot use async function in our Zod parser.
- Re-introduces `MarkdownEntryPathSchema` instead of `z.string()`

## 🧪 How to Test

Everything should behave the same as on fil.org


mirhamasala (approved):
Wowie! Let's go! ❤️

PR #1057, opened 2025-02-03
Add generateStaticParams to slug pages

CharlyMartin:
## 📝 Description

This PR implements `generateStaticParams` to statically generate pages at build time instead of on-demand at request time.

## 🛠️ Key Changes

- Adds `generateStaticParams` to `[slug]` pages
- Creates `SlugParams` type
- Add `Suspense` boundary around `Tabs`

## 🔖 Resources

https://nextjs.org/docs/app/api-reference/functions/generate-static-params

### DIFF
--- a/apps/site/src/app/blog/[slug]/page.tsx
+++ b/apps/site/src/app/blog/[slug]/page.tsx
@@ -20,24 +20,6 @@ type BlogPostProps = {
   params: Promise<SlugParams>
 }
 
-export async function generateStaticParams() {
-  const entries = await getBlogPostsData()
-  return entries.map(({ slug }) => ({ slug }))
-}
-
-export async function generateMetadata(props: BlogPostProps) {
-  const { slug } = await props.params
-  const data = getBlogPostData(slug)
-
-  return createMetadata({
-    seo: {
-      ...data.seo,
-      image: data.image?.src || graphicsData.blog.data.src,
-    },
-    path: `${PATHS.BLOG.path}/${data.slug}` as DynamicPathValues,
-  })
-}
-
 export default async function BlogPost(props: BlogPostProps) {
   const { slug } = await props.params
   const data = getBlogPostData(slug)
@@ -68,4 +50,22 @@ export default async function BlogPost(props: BlogPostProps) {
       </div>
     </PageLayout>
   )
+}
+
+export async function generateStaticParams() {
+  const entries = await getBlogPostsData()
+  return entries.map(({ slug }) => ({ slug }))
+}
+
+export async function generateMetadata(props: BlogPostProps) {
+  const { slug } = await props.params
+  const data = getBlogPostData(slug)
+
+  return createMetadata({
+    seo: {
+      ...data.seo,
+      image: data.image?.src || graphicsData.blog.data.src,
+    },
+    path: `${PATHS.BLOG.path}/${data.slug}` as DynamicPathValues,
+  })
 }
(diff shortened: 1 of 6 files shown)


apps/site/src/app/blog/[slug]/page.tsx
```diff
@@ -9,15 +11,21 @@ import { PageLayout } from '@/components/PageLayout'
...
 
 type BlogPostProps = {
-  params: Promise<{
-    slug: string
-  }>
+  params: Promise<SlugParams>
+}
+
+export async function generateStaticParams() {
+  const entries = await getBlogPostsData()
```
Comment (mirhamasala) [inline apps/site/src/app/blog/[slug]/page.tsx:24]: So now that we are using `getBlogPostsData()`, I think we should add a follow up ticket to remove `getBlogPostData()` to:

- Eliminate duplicate file reads
- Maintain single source of truth 
- Simplify caching
- Consolidate transforms

What do you think?

apps/site/src/app/blog/[slug]/page.tsx
```diff
@@ -9,15 +11,21 @@ import { PageLayout } from '@/components/PageLayout'
...
+
+  return entries.map((entry) => ({
+    slug: entry.slug,
+  }))
```
Comment (mirhamasala) [inline apps/site/src/app/blog/[slug]/page.tsx:28]:
```suggestion
  return entries.map(({ slug }) => ({ slug }))
```

mirhamasala (changes requested):
Looking good @CharlyMartin - Just an organization nit and something I'd like your input on. Thank you! 🙏🏼

apps/site/src/app/blog/[slug]/page.tsx
```diff
@@ -9,15 +11,21 @@ import { PageLayout } from '@/components/PageLayout'
...
 import { generateStructuredData } from './utils/generateStructuredData'
 
 type BlogPostProps = {
-  params: Promise<{
-    slug: string
-  }>
+  params: Promise<SlugParams>
+}
+
+export async function generateStaticParams() {
```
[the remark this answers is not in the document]
Comment (CharlyMartin) [reply apps/site/src/app/blog/[slug]/page.tsx:23]: Sure, happy with that!

apps/site/src/app/blog/[slug]/page.tsx
Comment (CharlyMartin) [reply apps/site/src/app/blog/[slug]/page.tsx:24]: If we can cache the result of `getBlogPostsData`, it could be a nice option. I don't fully understand the difference between React cache and the `unstable_cache`. Reach cache might only be for the duration of a session, to be confirmed.

Intuitively I would do it the other way around, where we don't read all entries just to get one. Though if caching works then it doesn't matter.

```ts
export async function getEventData(slug: string) {
  const event = await getMarkdownData({
    slug,
    directoryPath: EVENTS_DIRECTORY_PATH,
    zodParser: EventFrontmatterSchema.parse,
  })

  return {
    ...event,
    seo: {
      ...event.seo,
      title: event.seo.title || `${event.title} ${METADATA_TITLE_SUFFIX}`,
    },
  }
}

export async function getEventsData() {
  const filenames = await glob('**/*.md', {
    cwd: path.join(process.cwd(), EVENTS_DIRECTORY_PATH),
  })

  return Promise.all(
    filenames.map((filename) => {
      const slug = extractSlugFromFilename(filename)
      return getEventData(slug)
    }),
  )
}
```

Comment (mirhamasala) [reply apps/site/src/app/blog/[slug]/page.tsx]: I don't understand this:

> Intuitively I would do it the other way around, where we don't read all entries just to get one.

Aren't we already needing to get all the entries `const entries = await getBlogPostsData()` now that we are using `generateStaticParams`?

mirhamasala (approved):
Wonderful. ❤️

apps/site/src/app/blog/[slug]/page.tsx
Comment (CharlyMartin) [reply apps/site/src/app/blog/[slug]/page.tsx]: I see what you mean: We fetch all entries in `generateStaticParams` so we could re-use the result of the function call to get a single entry.

I think that works fine. It would be better if we could cache this get-all function. Though, since everything happens at build time, it might not make a big difference.

Happy to discuss it further!

PR #1105, opened 2025-02-14
[FF][FFDW][UXIT-2119] Set up shared TypeScript config and integrate monorepo workspaces [skip percy]

mirhamasala:
## 📝 Description

This PR introduces a shared TypeScript configuration package and integrates it across the monorepo. The goal is to maintain consistency, improve maintainability, and centralize TypeScript settings across all workspaces. 

### **Motivation & Problem Solved:**
- Standardizing TypeScript settings across projects.
- Reducing duplicated TypeScript configurations.
- Ensuring all workspaces follow the same linting and compilation rules.

**Type:** Refactor

## 🛠️ Key Changes

- Added `ts-config` package for shared TypeScript settings.
- Configured monorepo to include the `packages/*` workspace.
- Added `@filecoin-foundation/typescript-config` to dependencies.
- Updated `tsconfig.json` to extend from `@filecoin-foundation/typescript-config/base.json`.
- Configured TypeScript paths to align with the monorepo structure.

## 📌 To-Do Before Merging

- [x] Verify all workspaces build correctly with the new configuration.
- [x] Ensure no breaking changes in existing projects.

## 🧪 How to Test

- **Setup:**  
  1. Run `npm install` to ensure all dependencies are installed.
  2. Verify that `@filecoin-foundation/typescript-config` is correctly installed.

- **Steps to Test:**
  1. Run `npm run build` to check if TypeScript compiles without errors.
  2. Run `npm run lint` to verify that linting works with the new configuration.
  3. Open a project within the monorepo and confirm that TypeScript path mappings work as expected.

- **Expected Results:**  
  - The monorepo should build successfully.
  - No TypeScript configuration errors should occur.
  - Linting should work without issues.

## 📸 Screenshots

N/A – no UI changes.

## 🔖 Resources

- [Turbo Monorepo Documentation](https://turbo.build/repo/docs)

## Notes

Ideally, we would have another template (`app.json`), which would extend `base.json`, and then we would extend `app.json` in both apps. However, Next.js does not seem to properly recognize the configuration when fully extended, causing issues with the expected behavior. Due to this limitation, we are keeping the current setup. If a workaround is found in the future, we can revisit this approach.

```ts
// typescript-config/app.json

{
  "extends": "@filecoin-foundation/typescript-config/base.json",
  "compilerOptions": {
    "noEmit": true,
    "allowJs": true,
    "jsx": "preserve",
    "paths": {
      "@/*": ["./src/app/*"],
      "@/actions/*": ["./src/app/_actions/*"],
      "@/assets/*": ["./public/assets/*"],
      "@/components/*": ["./src/app/_components/*"],
      "@/constants/*": ["./src/app/_constants/*"],
      "@/content/*": ["./src/content/*"],
      "@/data/*": ["./src/app/_data/*"],
      "@/hooks/*": ["./src/app/_hooks/*"],
      "@/schemas/*": ["./src/app/_schemas/*"],
      "@/services/*": ["./src/app/_services/*"],
      "@/styles/*": ["./src/app/_styles/*"],
      "@/support/*": ["./cypress/support/*"],
      "@/types/*": ["./src/app/_types/*"],
      "@/utils/*": ["./src/app/_utils/*"],
      "@filecoin-foundation/ui": ["../packages/ui/src"]
    },
    "plugins": [
      {
        "name": "next"
      }
    ]
  },
  "include": [
    "next-env.d.ts",
    ".next/types/**/*.ts",
    "src/app/_types/**/*.d.ts",
    "**/*.ts",
    "**/*.tsx",
    "cypress/**/*.ts"
  ]
}
```


packages/typescript-config/base.json
```diff
@@ -0,0 +1,20 @@
+{
```
CharlyMartin, line 1: Matt Pocock recommends adding moduleDetection `"moduleDetection": "force"` to _avoid 'cannot redeclare block-scoped variable' errors._

https://www.totaltypescript.com/tsconfig-cheat-sheet#base-options
https://www.totaltypescript.com/cannot-redeclare-block-scoped-variable

packages/typescript-config/base.json
```diff
@@ -0,0 +1,20 @@
...
+  "$schema": "https://json.schemastore.org/tsconfig",
+  "compilerOptions": {
+    "verbatimModuleSyntax": true,
+    "target": "ES2022",
+    "lib": ["dom", "dom.iterable", "esnext"],
```
CharlyMartin, line 6: Matt Pocock recommends using `es2022` instead of `esnext` because it _is the best option for stability_

https://www.totaltypescript.com/tsconfig-cheat-sheet#running-in-the-dom

apps/site/tsconfig.json
```diff
@@ -1,49 +1,37 @@
...
-    "allowJs": true,
-    "skipLibCheck": true,
-    "strict": true,
     "noEmit": true,
-    "esModuleInterop": true,
-    "module": "esnext",
-    "moduleResolution": "node",
-    "resolveJsonModule": true,
-    "isolatedModules": true,
+    "allowJs": true,
```
CharlyMartin, line 5: This one is already included in `base.json`

apps/site/tsconfig.json
```diff
@@ -1,49 +1,37 @@
...
+        "name": "next"
+      }
+    ]
   },
   "include": [
     "next-env.d.ts",
-    "**/*.ts",
-    "**/*.tsx",
     ".next/types/**/*.ts",
     "src/app/_types/**/*.d.ts",
```
CharlyMartin, line 32: Would that not be covered by `"**/*.ts"`

apps/ffdweb-site/tsconfig.json
```diff
@@ -1,23 +1,9 @@
...
-    "allowJs": true,
-    "skipLibCheck": true,
-    "strict": true,
     "noEmit": true,
-    "esModuleInterop": true,
-    "module": "esnext",
-    "moduleResolution": "bundler",
-    "resolveJsonModule": true,
-    "isolatedModules": true,
+    "allowJs": true,
```
CharlyMartin, line 5: Same comment, already included in the base config

CharlyMartin (commented):
This is super cool Mirha! Our monorepo is getting real fancy 😄

I’ve added comments for reference / discussion. Let me know if they make sense 🙏

packages/typescript-config/base.json
mirhamasala: We are using `es2022`?

apps/site/tsconfig.json
mirhamasala: Yes, it should be.

packages/typescript-config/package.json
```diff
@@ -0,0 +1,9 @@
+{
```
CharlyMartin, line 1: I forgot to mention that Prettier needs to be run on this page

packages/typescript-config/base.json
mirhamasala: Enabling this is currently causing issues with import theme from 'tailwindcss/defaultTheme', so I’ll keep things as they are until we sort out the Tailwind V4 migration.

packages/typescript-config/base.json
mirhamasala: @CharlyMartin – We can’t use `es2022` because we are currently using `toSorted`, which was introduced in `ES2023`. If we downgrade to `es2022`, `toSorted` won’t be recognized. Since our code relies on this feature, we need to stick with `esnext` to ensure compatibility.

CharlyMartin: Yes, good catch!

PR #1106, opened 2025-02-14
[UXIT-1701] Add tooltips to Share Post icons

barbaraperic:
## 📝 Description

This PR includes new `Tooltip` component used for Share Icons, improved accessibility for `CopyToClipboard` and refactor for `Popover` component (previously named `Tooltip`).

- **Type:** Bug fix / New feature / Refactor

## 🛠️ Key Changes

- Added tooltip functionality
- Implemented share icons
- Refactored popover component
- Enhanced accessibility for copy to clipboard functionality

### Hover and focus state

https://github.com/user-attachments/assets/fa367924-fb66-47f8-9772-564d63c32fce

### Voice Over

https://github.com/user-attachments/assets/fc26c6bb-bdaa-40c7-8fa0-2e598d08037b

## 🔖 Resources

- [Radix Tooltip](https://www.radix-ui.com/primitives/docs/components/tooltip)
- [ARIA tooltip pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/)
- [Radix Popover](https://www.radix-ui.com/primitives/docs/components/popover)
- [ARIA dialog pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/)
- [Tooltips vs Popups](https://www.nngroup.com/articles/tooltip-guidelines/)
- [Tooltips best practices](https://css-tricks.com/tooltip-best-practices/)


apps/site/src/app/_components/Tooltip/Tooltip.tsx
```diff
@@ -27,29 +27,31 @@ export function Tooltip({ children, description, side = 'top' }: TooltipProps) {
...
+            hideWhenDetached
+            id={tooltipId}
+            sideOffset={GAP_BETWEEN_TOOLTIP_AND_TRIGGER}
+            side={side}
+            role="tooltip"
+            className={clsx(
+              styles['tooltip-content'],
+              styles['tooltip-animation'],
+            )}
+            onEscapeKeyDown={() => setOpen(false)}
```
CharlyMartin, line 48: This might not be needed?

https://www.radix-ui.com/primitives/docs/components/tooltip#keyboard-interactions

apps/site/src/app/_components/CopyToClipboard.tsx
```diff
@@ -51,15 +52,22 @@ export function CopyToClipboard({
...
-        className={clsx(
-          'focus:brand-outline hover:text-brand-400',
-          TOUCH_TARGET.touchAreaPadding,
-        )}
-        onClick={() => handleCopy(text)}
-      >
-        <Icon component={Link} size={32} weight="light" />
-      </Button>
+      <Tooltip description="Copy link to clipboard" side="bottom">
+        <Button
```
CharlyMartin, line 56: I believe `Button` should have an `aria-describedby` attribute referencing the ID from the Tooltip, as done in `PopoverHeader`

https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/

apps/site/src/app/_components/NotificationDialog.tsx
```diff
@@ -31,6 +31,8 @@ export function NotificationDialog({
     >
       <DialogPanel
         transition
+        role="alertdialog"
+        aria-modal="true"
```
CharlyMartin, line 35: Are these not automatically provided by HeadlessUI? Just curious, I know they’re usually thorough with attributes, but maybe not in this case.

CharlyMartin (commented):
Looking really good Barbara ☀️

apps/site/src/app/_components/CopyToClipboard.tsx
barbaraperic: Actually looking at their [source code](https://github.com/radix-ui/primitives/blob/6e75e117977c9e6ffa939e6951a707f16ba0f95e/packages/react/tooltip/src/tooltip.tsx#L287C1-L288C1) it seems like they handle it automatically?

barbaraperic: [Same for the id](https://github.com/radix-ui/primitives/blob/6e75e117977c9e6ffa939e6951a707f16ba0f95e/packages/react/tooltip/src/tooltip.tsx#L226) and [type="tooltip"](https://github.com/radix-ui/primitives/blob/6e75e117977c9e6ffa939e6951a707f16ba0f95e/packages/react/tooltip/src/tooltip.tsx#L564)?  Think we can remove those?

apps/site/src/app/_components/NotificationDialog.tsx
barbaraperic: The [documentation](https://headlessui.com/react/dialog#dialog) shows that it accepts either 'dialog' or 'alertdialog' as valid roles, with 'dialog' being the default. Since we are using the dialog to alert anout successfully copied link I used the `alertdialog`. Also, the type `dialog` didn't work when I tested the voiceover. 😬

barbaraperic: You're right, they do handle the aria modal internally, but the voiceover wasn't working without it, so I added it explicitly.

apps/site/src/app/_components/CopyToClipboard.tsx
CharlyMartin: Interesting! If they are still present in the HTML after removing them then yes

apps/site/src/app/_components/Table/PopoverHeader.tsx
```diff
@@ -22,18 +22,17 @@ const TOUCH_TARGET_ICON: TouchTarget & { visibleElementSize: number } = {
...
   return (
     <div className="flex items-center gap-1">
       <span>{title}</span>
-      <Tooltip description={description}>
-        {({ open, tooltipId }) => (
+      <Popover description={description}>
+        {({ open }) => (
           <Button
             aria-label={`More info on ${title}`}
-            aria-describedby={tooltipId}
```
CharlyMartin, line 36: @barbaraperic I inspected the HTML locally and didn't see `aria-describedby` on the button

Do you see it on your side?

```html
<button aria-label="More info on Type" class="group focus-visible:brand-outline p-2 -m-2" type="button" aria-haspopup="dialog" aria-expanded="true" aria-controls="radix-:Ridtuunat7:" data-state="open" data-headlessui-state="">
</button>
```

I looked at `PopoverHeader` the on http://localhost:3000/filecoin-plus/allocators

barbaraperic: True! I reverted back the id for the popover :) Thanks!

CharlyMartin: Great! Should we do the same on the `Tooltip` component then? What do you think?

barbaraperic: Don't think we need to? for tooltip `id` and `aria-describedby` are assigned automatically

apps/site/src/app/_components/CopyToClipboard.tsx
```diff
@@ -51,15 +52,22 @@ export function CopyToClipboard({
...
+          className={clsx(
+            'focus:brand-outline hover:text-brand-400',
+            TOUCH_TARGET.touchAreaPadding,
+          )}
+          onClick={() => handleCopy(text)}
+          onKeyDown={(e) => {
+            if (e.key === 'Enter' || e.key === ' ') {
```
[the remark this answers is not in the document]
barbaraperic: Ah we actually don't need it. When I was testing it, I thought there was an issue with the copy functionality, but it turns out it's working properly. Thanks, I'll go ahead and remove it!

barbaraperic:
Closing this PR in [favor of new branch](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1137) - due to extensive merge conflicts, too big resolution effort for a few files.

PR #1113, opened 2025-02-18
Improve CMS Event Times

CharlyMartin:
## 📝 Description

This PR updates the event's schedule's `start` & `end` fields by removing the date portion. This will improve the CMS UX for the content team, standardize the data, and clarify things for us.

> [!NOTE]
> Times are now parsed as strings, and only converted to Date instance when needed, for comparison or formatting for example.

## 🛠️ Key Changes

- Updates `config.yml` with new `time_iso_format` format
- Renames `start` to `start-time` and `end` to `end-time`
- Adds `IsoTimeSchema`
- Updates Markdown entries accordingly

## 🧪 How to Test

Run the CMS locally, create an event and ensure the `start-time` and `end-time` fields are only times, not dates:
`Event > Schedule > Day > Event`

## 📸 Screenshots

CleanShot 2025-02-19 at 11 46 16@2x

## 🔖 Resources

https://decapcms.org/docs/widgets/datetime/

### DIFF
--- a/apps/site/src/app/events/[slug]/utils/dateUtils.ts
+++ b/apps/site/src/app/events/[slug]/utils/dateUtils.ts
@@ -3,6 +3,8 @@ import { format } from 'date-fns'
 
 import { formatDate } from '@/utils/dateUtils'
 
+import { ISO_TIME_REGEX } from '@/schemas/DateTimeSchema'
+
 const PLACEHOLDER_DATE = '1970-01-01'
 
 export function formatShortDate(date: Date) {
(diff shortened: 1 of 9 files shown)


apps/site/src/app/events/[slug]/utils/dateUtils.ts
```diff
@@ -3,12 +3,26 @@ import { format } from 'date-fns'
 
 import { formatDate } from '@/utils/dateUtils'
 
+import { ISO_TIME_REGEX } from '@/schemas/DateTimeSchema'
+
+const PLACEHOLDER_DATE = '1970-01-01'
```
Comment (mirhamasala) [inline apps/site/src/app/events/[slug]/utils/dateUtils.ts:8]: What is this?

mirhamasala (approved):
Just a little question .. but LGTM! ☀️

apps/site/src/app/events/[slug]/utils/dateUtils.ts
Comment (CharlyMartin) [reply apps/site/src/app/events/[slug]/utils/dateUtils.ts:8]: This is a placeholder date so can create a `Date` instance from the time string. 

```ts
new Date("15:00:00.000Z")
// Invalid Date

new Date("1970-01-01T15:00:00.000Z")
// Date Thu Jan 01 1970 17:00:00 GMT+0200 (South Africa Standard Time)
```

PR #1126, opened 2025-02-20
WIP

CharlyMartin:
## 📝 Description

Please include a summary of the changes. Provide context and motivation for the change, and describe what problem it solves.

- **Type:** Bug fix / New feature / Documentation / Refactor

## 🛠️ Key Changes

- ...

## 📌 To-Do Before Merging

- [ ] [Task 1 - Brief description]
- [ ] [Task 2 - Brief description]
- ...

## 🧪 How to Test

- **Setup:** [Setup details]
- **Steps to Test:**
  1. [Step 1 - Brief description]
  2. [Step 2 - Brief description]
  3. ...
- **Expected Results:** [Outcome]
- **Additional Notes:** [Additional info]

## 📸 Screenshots

## 🔖 Resources

Feel free to share any references to documentation, libraries, blog posts, or other resources that you consulted or used during the implementation of the changes.

- [Resource 1]
- [Resource 2]
- ...

### DIFF
--- a/apps/site/src/app/events/schemas/EventBaseFrontmatterSchema.ts
+++ b/apps/site/src/app/events/schemas/EventBaseFrontmatterSchema.ts
@@ -1,3 +1,4 @@
+import { isAfter } from 'date-fns'
 import { z } from 'zod'
 
 import { IsoDateSchema } from '@/schemas/DateTimeSchema'
@@ -11,3 +12,19 @@ export const EventBaseFrontmatterSchema = z
     'external-link': z.string().url().optional(),
   })
   .strict()
+  .refine(
+    (data) => {
+      const { 'end-date': endDate, 'start-date': startDate } = data
+
+      if (!endDate) {
+        return true
+      }
+
+      return isAfter(endDate, startDate)
+    },
+    {
+      message:
+        'end-date must be greater than start-date. If end-date is the same as start-date, leave it empty',
+      path: ['end-date'],
+    },
+  )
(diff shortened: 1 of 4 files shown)


apps/ff-site/src/app/events/utils/getEventData.ts
```diff
@@ -1,3 +1,7 @@
+import assert from 'node:assert'
```
barbaraperic, line 1: Curious why you chose `assert`? Could we have gone with `throw new Error`?  
``` ts
  const isEndDateBeforeStart = endDate && !isAfter(endDate, startDate)

  if (isEndDateBeforeStart) {
    throw new Error(
      `${event.title}: end-date ${endDate} must be greater than start-date ${startDate}`,
    )
  }

Comment (barbaraperic) [reply apps/ff-site/src/app/events/utils/getEventData.ts:1]: No, I guess there's not much difference between the two, as in, they both catch the bug. But I was thinking more about the semantics of it. As I understood from the [Node documentation](https://nodejs.org/api/assert.html), assert is used for events that are unlikely to happen, for "verifying invariants", and throw Error is more for errors that are expected to happen, as in, they come from a user that can make a mistake (user in our case are users of CMS tool).

Comment (CharlyMartin) [reply apps/ff-site/src/app/events/utils/getEventData.ts:1]: Fair enough, I'll update :)

File: apps/ff-site/src/app/events/utils/getEventData.ts [hunk 2]
```diff
@@ -41,3 +45,37 @@ function transformEventData(
     },
   }
 }
+
+function assertEndIsAfterStart(
```
barbaraperic, line 49: Ah should we also then rename the function? Since we're not using `assert` or `isAfter`? Maybe something more generic like `validateEventTime`, or something like that, you're better in naming than me lol

PR #1139, opened 2025-02-25
Test

CharlyMartin:
## 📝 Description

Please include a summary of the changes. Provide context and motivation for the change, and describe what problem it solves.

- **Type:** Bug fix / New feature / Documentation / Refactor

## 🛠️ Key Changes

- ...

## 📌 To-Do Before Merging

- [ ] [Task 1 - Brief description]
- [ ] [Task 2 - Brief description]
- ...

## 🧪 How to Test

- **Setup:** [Setup details]
- **Steps to Test:**
  1. [Step 1 - Brief description]
  2. [Step 2 - Brief description]
  3. ...
- **Expected Results:** [Outcome]
- **Additional Notes:** [Additional info]

## 📸 Screenshots

## 🔖 Resources

Feel free to share any references to documentation, libraries, blog posts, or other resources that you consulted or used during the implementation of the changes.

- [Resource 1]
- [Resource 2]
- ...


.github/workflows/site-lint.yml
```diff
@@ -3,19 +3,12 @@ name: Site Lint
 on:
   push:
     branches: [main, cms/*, ecosystem-submission/*]
```
mirhamasala, line 5: Isn't this very specific to ff-site?

mirhamasala:
@CharlyMartin 

> Should we set up Percy on ffdweb now? I think it's too early as there will be many intentional changes at the beginning.

I think we can set it up - but we don't need to use the command yet.

> Should we move apps/ff-site/.markdownlint.jsonc to the root? Or to packages/eslint-config? Probably not necessary to make a package out of it.

How are they related? Can it be nicely integrated into eslint-config package? I think best practices dictate that they should be separate packages. Do we also need a shared prettier config? Or is that now all handled by eslint-config?

> The linter works, but it never fails the build; it only produces warnings. Should we change that?

In which case do you think it would be important for it to fail the buid?

.github/workflows/site-lint.yml
CharlyMartin: Yes, indeed.

Maybe we create a second action to lint only Markdown files in `cms/*` & `ecosystem-submission/*` PRs?

apps/ffdweb-site/package.json
```diff
@@ -6,7 +6,9 @@
     "dev": "next dev",
     "build": "next build",
     "start": "next start",
-    "lint": "next lint"
+    "lint": "next lint --fix",
+    "test:cypress": "echo 'test cypress command to be implemented'",
+    "test:cypress:percy": "echo 'test percy command to be implemented'"
```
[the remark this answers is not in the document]
CharlyMartin: The action failed because the command was missing, so I added a dummy one.

https://github.com/FilecoinFoundationWeb/filecoin-foundation/actions/runs/13517470550/job/37769089897#step:8:16

CharlyMartin:
@mirhamasala 

> Is it set?
> 
> > Should we set up Percy on ffdweb now? I think it's too early as there will be many intentional changes at the beginning.
> 
> I think we can set it up - but we don't need to use the command yet.
> 
> > Should we move apps/ff-site/.markdownlint.jsonc to the root? Or to packages/eslint-config? Probably not necessary to make a package out of it.
> 
> How are they related? Can it be nicely integrated into eslint-config package? I think best practices dictate that they should be separate packages. Do we also need a shared prettier config? Or is that now all handled by eslint-config?
> 
> > The linter works, but it never fails the build; it only produces warnings. Should we change that?
> 
> In which case do you think it would be important for it to fail the buid?

CharlyMartin:
@mirhamasala 

> Is it set?

I don't know. `env.TURBO_CACHE_DIR` is used in `repo-ci.yml` but I can't find where it's declared. Do you know? Is it automatically set by Vercel?

> I think we can set it up - but we don't need to use the command yet.

Ok, I'll set up Percy on ffdweb.

> How are they related? Can it be nicely integrated into eslint-config package? I think best practices dictate that they should be separate packages. Do we also need a shared prettier config? Or is that now all handled by eslint-config?

`apps/ff-site/.markdownlint.jsonc` is used by https://github.com/DavidAnson/markdownlint-cli2-action. It's not related to ESLINT, but it's related to linting. It's just a file, so I don't think we need a package for that. I would just put it at the root of the monorepo. Do you agree?

About Prettier, yes we should have a shared config too 🙂

> In which case do you think it would be important for it to fail the build?

I'm not sure. ESLINT used to fail the build when import orders were out of whack for instance. Now, it doesn't. If it never fails the build, what is the point of having it in the action pipeline?

.github/workflows/site-lint.yml
CharlyMartin: Remove it and test with a CMS PR

CharlyMartin:
@mirhamasala From chatGPT:

> When you run `npm run lint --fix` in a GitHub action, it doesn’t automatically commit those fixes to the codebase. The linting will apply fixes in the temporary filesystem of the GitHub action runner. However, changes are not automatically pushed back to the repository unless you set up additional steps to commit and push them. It’s generally not recommended to commit and push those changes. Linting should act as a check. Autocorrecting and pushing from CI/CD might cause unexpected issues or conflicts.

The answer makes sense to me. So I'd say we either remove the lining action or modify it so that it can fail a build.

So I suggest we remove it from the action pipeline. Instead, we can add a git hook (with Husky) to run linting locally on each commit and let the author commit the fixes.

CharlyMartin:
@mirhamasala About `${{ env.TURBO_CACHE_DIR }}`

It is indeed an allowed value but it's not set anywhere. I tried logging it [here](https://github.com/FilecoinFoundationWeb/filecoin-foundation/actions/runs/13522937961/job/37786260053) with a `Debug Cache Directory` job and it was empty.

So I set it to `.turbo/cache`, the [default value](https://turbo.build/repo/docs/reference/run#--cache-dir-path)

.github/workflows/site-lint.yml
CharlyMartin: So I created a separate linter action `.github/workflows/apps-markdown-linter.yml` only for the Markdown files.

This one is useful because it [fails the PR checks](https://github.com/FilecoinFoundationWeb/filecoin-foundation/actions/runs/13525637859/job/37795360459?pr=1139) if the markdown content is inconsistent. For instance, an h4 following an h1.

I'm gonna delete the `ESLINT` action because, as we said, it never fails [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1139#issuecomment-2682037170)

.github/workflows/apps-cypress-percy-test.yml
mirhamasala, line 1: This workflow is called `site-cypress-percy.yml` - but it's shared?

mirhamasala (approved):
@CharlyMartin - Good to go! ☀️ Just a naming nit question. Please feel free to merge after you have a look. Thank you for cleaning this up and making it work. 🙏🏼

.github/workflows/apps-cypress-percy-test.yml
CharlyMartin: No it's called `apps-cypress-percy-test.yml` now

mirhamasala: Got it! LGTM!

mirhamasala: (Was looking at the wrong line!)

PR #1148, opened 2025-02-26
Move Breadcrumbs to shared components [skip percy]

CharlyMartin:
## 📝 Description

This PR moves the existing `BreadCrumbs` component and dependent `utils` in the shared UI package.

## 🛠️ Key Changes

1. Creates `packages/ui/src/BreadCrumbs.tsx` & `packages/ui/src/BreadCrumbsLayout.tsx`
3. Installs `@phosphor-icons/react`, `clsx` and `next` as peer dependencies in shared UI package
4. Creates `packages/utils/src/stringUtils.ts` with `capitalize` and `truncate` functions
5. Creates `breadcrumbs-active` class in `apps/ff-site/src/app/_styles/globals.css`


apps/ff-site/src/app/_components/BreadCrumbsLayout.tsx
```diff
@@ -1,13 +1,30 @@
...
 export function BreadCrumbsLayout({ children }: LayoutProps) {
+  const pathname = usePathname()
+
   return (
     <div className="flex grow flex-col gap-8">
-      <BreadCrumbs />
+      <BreadCrumbs
+        currentPath={pathname}
+        homePath={PATHS.HOME.path}
+        homeLabel={PATHS.HOME.label}
```
mirhamasala, line 21: Can we just use `/` and `Home`?

```suggestion
        homePath={PATHS.HOME.path}
        homeLabel={PATHS.HOME.label}
```

packages/ui/src/BreadCrumbs.tsx
```diff
@@ -0,0 +1,74 @@
...
+            <li key={href} className="inline-flex items-center gap-2.5">
+              {!isRoot && (
+                <Icon
+                  component={CaretRight}
+                  size={20}
+                  color="subtle"
+                  weight="bold"
+                />
+              )}
+              {renderLink({
```
mirhamasala, line 56: Let's us `TextLink`

apps/ff-site/src/app/_components/BreadCrumbsLayout.tsx
```diff
@@ -1,13 +1,30 @@
...
+import { usePathname } from 'next/navigation'
+
+import { BreadCrumbs } from '@filecoin-foundation/ui/BreadCrumbs'
+import type { Route } from 'next'
+
+import { PATHS } from '@/constants/paths'
 
 import type { LayoutProps } from '@/layout'
 
 export function BreadCrumbsLayout({ children }: LayoutProps) {
```
mirhamasala, line 13: Let's make it a shared component

packages/ui/package.json
```diff
@@ -15,21 +15,26 @@
...
     "react": "^19.0.0",
     "react-dom": "^19.0.0",
-    "tailwindcss": "^4.0.5",
-    "postcss": "^8.5.1",
-    "@tailwindcss/postcss": "^4.0.5"
+    "tailwindcss": "^4.0.5"
+  },
+  "dependencies": {
+    "@phosphor-icons/react": "^2.1.7",
+    "clsx": "^2.1.1"
```
mirhamasala, line 38: We should add these to FFDW `package.json` instead - That should fix the Cypress errors. We can also include them under peer dependencies.

mirhamasala (changes requested):
@CharlyMartin - Let's rework as discussed. 🙏🏼

packages/ui/package.json
CharlyMartin: I did it in the other PR https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1139/commits/84dbd69db3f70171773d4af6ad486ba8965d97e5

packages/ui/src/BreadCrumbs.tsx
```diff
@@ -1,48 +1,36 @@
...
+type BreadCrumbsProps = {
+  homePath?: string
+  homeLabel?: string
+}
 
-export function BreadCrumbs() {
+export function BreadCrumbs({
+  homePath = '/',
+  homeLabel = 'Home',
+}: BreadCrumbsProps) {
```
mirhamasala, line 19: Why do we still want to pass props?

mirhamasala (changes requested):
@CharlyMartin - Left a [nit comment](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1148#discussion_r1971687430). Also, can we actually implement it for FFDW? Or will that happen in a separate PR? ~❤️

packages/ui/src/BreadCrumbs.tsx
CharlyMartin: It's just to set the default values

CharlyMartin:
> @CharlyMartin - Left a [nit comment](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1148#discussion_r1971687430). Also, can we actually implement it for FFDW? Or will that happen in a separate PR? ~❤️

Ah yes of course I'll implement in FFDW 🙂

packages/ui/src/BreadCrumbs.tsx
mirhamasala: Shall we just hardcode them then? ☺️

CharlyMartin:
@mirhamasala—I added the breadcrumbs to FFDW, too. Right now, they are purple. Figma says white and purple, but it's not feasible with our current implementation.

packages/ui/src/BreadCrumbs.tsx
CharlyMartin: Hardcoded :)

mirhamasala (approved):
@CharlyMartin ~ I pushed a CR. We can align styles with Filipa at a later stage. I'm going to merge this right no to prevent conflicts as we go since this PR touches all the files. 🙏🏼

PR #1149, opened 2025-02-26
[UXIT-2248] Move buildImageSizeProp to shared Utils [skip percy]

barbaraperic:
## 📝 Description

In this PR we move buildImageSizeProp to shared `packages/utils` folder

- **Type:** Refactor

## 🛠️ Key Changes

- moved `buildImageSizeProp` to `packages/utils/src/` folder
- updated the imports
- temporarily moved the `breakpoints` constant and its type (`BreakpointValue`) to `buildImageSizeProp.ts`, until we decide on the best location for these shared constants and types (possibly in a dedicated constants/types file or a more appropriate package folder). Note that this means there are duplication with existing constants/types files.


packages/utils/src/buildImageSizeProp.ts
```diff
@@ -1,6 +1,12 @@
...
-import { breakpoints } from '@/constants/tailwindConstants'
+const breakpoints = {
+  sm: '640px',
+  md: '768px',
+  lg: '1024px',
+  xl: '1280px',
+  '2xl': '1536px',
+} as const
+
+type BreakpointValue = keyof typeof breakpoints
```
mirhamasala, line 9: Don't we want to move `tailwindConstants`?

mirhamasala: Sorry - Just read the PR comment. Maybe we can move it under `utils/constants` - just as we do for `utils/types` for now - Until we have a better understanding. At least, this way, we don't have duplicates.

mirhamasala (approved):
@barbaraperic LGTM! ❤️

PR #1168, opened 2025-03-03
[UXIT-2222] FFDW Structured Data for Static Pages [skip percy]

barbaraperic:
## 📝 Description

This PR includes structured data for static pages.

- **Type:** New feature

### DIFF
--- a/apps/ffdweb-site/src/app/_constants/structuredDataConstants.ts
+++ b/apps/ffdweb-site/src/app/_constants/structuredDataConstants.ts
@@ -1,0 +1,43 @@
+import type { Organization, WithContext } from 'schema-dts'
+
+import {
+  BASE_URL,
+  FFDW_URLS,
+  ORGANIZATION_NAME,
+  ORGANIZATION_NAME_SHORT,
+} from '@/constants/siteMetadata'
+
+const { social, email } = FFDW_URLS
+
+export const SCHEMA_CONTEXT_URL = 'https://schema.org'
+export const SCHEMA_EVENT_ATTENDANCE_MODE_ONLINE_URL =
+  'https://schema.org/OnlineEventAttendanceMode'
+export const SCHEMA_EVENT_ATTENDANCE_MODE_OFFLINE_URL =
+  'https://schema.org/OfflineEventAttendanceMode'
+
+const DESCRIPTION =
+  'FFDW is a nonprofit organization committed to preserving humanity’s most important information by funding the development of open-source tools.'
+
+export const ORGANIZATION_SCHEMA_BASE: WithContext<Organization> = {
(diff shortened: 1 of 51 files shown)


apps/ffdweb-site/src/app/_utils/generateWebPageStructuredData.ts
mirhamasala, line 1: This seems to be exactly the same as for ff-site. Is there a reason not to extract it into the shared util library?

apps/ffdweb-site/src/app/_constants/structuredDataConstants.ts
```diff
@@ -0,0 +1,43 @@
...
+const { social, email } = FFDW_URLS
+
+export const SCHEMA_CONTEXT_URL = 'https://schema.org'
+export const SCHEMA_EVENT_ATTENDANCE_MODE_ONLINE_URL =
+  'https://schema.org/OnlineEventAttendanceMode'
+export const SCHEMA_EVENT_ATTENDANCE_MODE_OFFLINE_URL =
+  'https://schema.org/OfflineEventAttendanceMode'
+
+const DESCRIPTION =
+  'FFDW is a nonprofit organization committed to preserving humanity’s most important information by funding the development of open-source tools.'
```
mirhamasala, line 19: I think we should import this from `apps/ffdweb-site/src/app/(homepage)/page.tsx`

apps/ffdweb-site/src/app/_constants/structuredDataConstants.ts
```diff
@@ -0,0 +1,43 @@
...
+  ORGANIZATION_NAME_SHORT,
+} from '@/constants/siteMetadata'
+
+const { social, email } = FFDW_URLS
+
+export const SCHEMA_CONTEXT_URL = 'https://schema.org'
+export const SCHEMA_EVENT_ATTENDANCE_MODE_ONLINE_URL =
+  'https://schema.org/OnlineEventAttendanceMode'
+export const SCHEMA_EVENT_ATTENDANCE_MODE_OFFLINE_URL =
+  'https://schema.org/OfflineEventAttendanceMode'
```
mirhamasala, line 16: Do we need these two for FFDW?

apps/ffdweb-site/src/app/_utils/generateWebPageStructuredData.ts
barbaraperic: they have different paths and base url, but I'll look into standardising the function 👍

mirhamasala: Ah, I see. Let's keep it duplicated for now. Thank you for investigating.

PR #1176, opened 2025-03-04
[UXIT-2289] FFDW FAQs page

barbaraperic:
## 📝 Description

This PR includes FFDW FAQs page.

- **Type:**  New feature

## 📸 Screenshots

screencapture-localhost-3000-faqs-2025-03-04-14_34_47 (2)

### DIFF
--- a/apps/ffdweb-site/src/app/faqs/components/PageHeader.tsx
+++ b/apps/ffdweb-site/src/app/faqs/components/PageHeader.tsx
@@ -1,0 +1,41 @@
+import { Heading } from '@filecoin-foundation/ui/Heading'
+
+type TitleProps = {
+  children: string
+}
+
+type PageHeaderProps = {
+  heading: TitleProps['children']
+  subheading: TitleProps['children']
+  kicker?: string
+}
+
+export function PageHeader({ kicker, heading, subheading }: PageHeaderProps) {
+  return (
+    <header className="flex max-w-3xl flex-col gap-8">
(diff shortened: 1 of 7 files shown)


packages/ui/src/PageLayout.tsx
```diff
@@ -7,7 +7,7 @@ type PageLayoutProps = {
 
 const gapMap = {
   16: 'gap-16',
-  32: 'gap-32',
+  32: 'sm:gap-32 gap-16',
```
mirhamasala, line 10: I can't find this back in the design .. ?

apps/ffdweb-site/src/app/faqs/components/PageHeader.tsx
mirhamasala, line 1: We don't need to export this so we can simplify it quite a bit! https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1176 🙏🏼

packages/ui/src/PageLayout.tsx
barbaraperic: Might be a design hiccup - Home, About and Projects have `gap-32` on all screens, but the rest of the pages have [`gap-16` on `sm` screen. ](https://www.figma.com/design/P67wEFE8S7ay0b7FErfqPE/FFDweb.org-Prototypes?node-id=5677-56322&t=5zpUvTGCKR8EJM6S-4)

PR #1178, opened 2025-03-04
[FF][FFDW] Make Card shareable [skip percy]

CharlyMartin:
## 📝 Description

This PR moves `Card` & `Meta` in our shared UI package. The implementation on FFDW will be done in a second PR.

## 🛠️ Key Changes

- Moves `Card` & `Meta` in `packages/ui`
- Creates `Card` wrapper in `apps/ff-site/src/app/_components/` to set `BASE_DOMAIN` locally

## 📸 Screenshots

No UI changes.


apps/ff-site/src/app/_styles/globals.css
```diff
@@ -219,4 +219,33 @@
...
+    }
+  }
+
+  .card-image {
+    @apply rounded-lg px-1 pt-1;
+  }
+
+  .card-link {
+    @apply text-brand-300;
+  }
```
mirhamasala, line 250: Why are these not grouped under `.card`?

packages/ui/src/Meta.tsx
```diff
@@ -14,7 +14,7 @@ export function Meta({ metaData }: MetaProps) {
   }
 
   return (
-    <div className="flex flex-wrap gap-x-2 gap-y-1 text-brand-300">
+    <div className="text-brand-300 flex flex-wrap gap-x-2 gap-y-1">
```
mirhamasala, line 17: Should this style be extracted for both apps?

mirhamasala (approved):
Wonderful! 🙌🏼 Two nits. And adding a note to refactor the other components using `Shared` prefix and wrapper for `BASE_DOMAIN`. Love this solution. ❤️

packages/ui/src/Meta.tsx
CharlyMartin: Ah yes! Good catch :)

apps/ff-site/src/app/_styles/globals.css
CharlyMartin: `Card.Link` is used as a standalone component somewhere else.

If we nest `.card-link` inside `.card`, then it won't work for standalone components? Right?

mirhamasala: Ah, I didn't realize. Thank you!

PR #1181, opened 2025-03-05
[FF][FFDW] [UXIT-2304] Shared Tooltip component [skip percy]

barbaraperic:
## 📝 Description

This PR includes:
- `Tooltip` and `Popover` component being moved to a shared library
- updated imports


packages/utils/postcss.config.mjs
mirhamasala, line 1: Why?

mirhamasala (changes requested):
I don't think we're handling the CSS imports the right way. Do you want to have another look at the docs?

barbaraperic:
> I don't think we're handling the CSS imports the right way. Do you want to have another look at the docs?

Yes, sorry, I'll take another look!

CharlyMartin (approved):
🤘

apps/ff-site/src/app/_styles/globals.css
```diff
@@ -25,6 +25,10 @@
   --color-icon-error: theme('colors.brand.error');
 
   --spacing-readable: 60ch;
+
+  --tooltip-background-color: var(--color-brand-200);
+  --tooltip-color: var(--color-brand-800);
+  --tooltip-arrow-color: var(--color-brand-200);
```
mirhamasala, line 31: Just checking - did we not decide to use classnames?

barbaraperic: Correct 👍 I pushed a fix

packages/ui/src/Tooltip/Tooltip.tsx
```diff
@@ -33,7 +33,7 @@ export function Tooltip({ children, description, side = 'top' }: TooltipProps) {
             sideOffset={GAP_BETWEEN_TOOLTIP_AND_TRIGGER}
             side={side}
             role="tooltip"
-            className="tooltip-content tooltip-animation"
+            className="tooltip-base tooltip-content tooltip-animation"
```
mirhamasala, line 36: This is a nit - but I feel like it would be nice to extract this into a constant and then import it in `Popover` - it would ensure that they're always in sync.

packages/ui/src/Tooltip/Tooltip.tsx
```diff
@@ -33,7 +33,7 @@ export function Tooltip({ children, description, side = 'top' }: TooltipProps) {
...
           >
             {description}
             <RadixTooltip.Arrow className="tooltip-arrow" />
```
mirhamasala, line 39: I would probably do it for this className, too.

apps/ff-site/src/app/_styles/globals.css
```diff
@@ -279,4 +279,31 @@
...
+    @apply hover:text-brand-400;
+  }
+
+  /* NOTIFICATION DIALOG */
+  .notification {
+    @apply rounded-lg border border-brand-100/20 bg-brand-800;
+
+    &.notification-title {
+      @apply text-brand-100;
+    }
```
mirhamasala, line 294: These are defaults, coming from `site-layout` - so going to remove them! 🙏🏼

packages/ui/src/Tooltip/Tooltip.tsx
mirhamasala: Here's a [a1f558f](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1181/commits/a1f558f87147c283b2b18466d55e719e86f09e92) - This actually helps us not have classes in two different files (except for the animation, which we could still move to a shared globals file if we wanted as some point.)

PR #1182, opened 2025-03-05
Improve BreadCrumbs styling [skip percy]

CharlyMartin:
## 📝 Description

This PR refactors `BreadCrumbs` and fixes the inactive style of FFDW

### DIFF
--- a/apps/ff-site/src/app/_styles/globals.css
+++ b/apps/ff-site/src/app/_styles/globals.css
@@ -116,12 +116,16 @@
   }
 
   /* BREADCRUMBS */
-  .breadcrumbs-inactive {
-    @apply text-brand-300;
-  }
+  .breadcrumb-link {
+    @apply hover:underline focus:brand-outline;
 
-  .breadcrumbs-active {
-    @apply text-brand-400;
+    &:not([aria-current='page']) {
+      @apply text-brand-300;
+    }
+
+    &[aria-current='page'] {
+      @apply text-brand-400;
+    }
   }
 
   /* ICON BUTTONS */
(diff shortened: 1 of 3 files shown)


apps/ff-site/src/app/_styles/globals.css
```diff
@@ -116,12 +116,16 @@
   }
 
   /* BREADCRUMBS */
-  .breadcrumbs-inactive {
-    @apply text-brand-300;
-  }
+  .breadcrumb-link {
+    @apply hover:underline focus:brand-outline;
```
mirhamasala, line 120: This is the same for both ..

packages/ui/src/BreadCrumbs.tsx
```diff
@@ -3,21 +3,32 @@
...
 const HOME_LABEL = 'Home'
 
-export function BreadCrumbs() {
+const gapMap = {
+  1: 'gap-1',
+  1.5: 'gap-1.5',
+  2: 'gap-2',
+  2.5: 'gap-2.5',
+  3: 'gap-3',
+}
```
mirhamasala, line 19: Do we need these many options?

CharlyMartin: Sure yeah that works!

packages/ui/src/BreadCrumbs.tsx
```diff
@@ -3,21 +3,31 @@
...
 import type { Route } from 'next'
 
 const HOME_PATH = '/'
 const HOME_LABEL = 'Home'
 
-export function BreadCrumbs() {
+const gapMap = {
+  small: 'gap-1.5',
+  default: 'gap-2',
+  large: 'gap-2.5',
```
mirhamasala, line 17: @CharlyMartin - I don't actually understand this. 🙈 We don't seem to be using it?

mirhamasala (approved):
🪤

PR #1183, opened 2025-03-05
Shared CopyToClipboard [skip percy]

barbaraperic:
## 📝 Description

[PR 1](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1181)

This PR moves `CopyToClipboard` and `NotificationDialog` to shared packaged.

- **Type:** Refactor

PR #1187, opened 2025-03-05
[FFDW][UXIT-2304] Share Article component [skip percy]

barbaraperic:
## 📝 Description

[PR 1](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1181)
[PR 2](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1183)

This PR moves `ShareArticle` to shared package.

- **Type:**  Refactor

Screenshot 2025-03-07 at 13 21 02
Screenshot 2025-03-07 at 13 23 39


apps/ff-site/src/app/_styles/globals.css
```diff
@@ -266,4 +266,15 @@
...
+  /* SHARE ARTICLE */
+  .share-article {
+    &.title {
+      @apply text-brand-300;
+    }
+
+    &.tooltip {
+      @apply hover:text-brand-400;
+    }
+  }
```
mirhamasala, line 279: Does this actually work as expected? I don't think this targets `share-article-title` and `share-article-tooltip` but `share-article title` and `share-article tooltip`? Right? 😵‍💫🙈

mirhamasala (changes requested):
@barbaraperic  ~ Amazing work! We're getting there. 🥰 Left a question! 🙏🏼

apps/ff-site/src/app/_styles/globals.css
mirhamasala: Sorry, forget what I said! I just realized it's `&.`

The only change I would make is:

```
  .share-article {
    &.share-article-title {
      @apply text-brand-300;
    }

    &.share-article-tooltip {
      @apply hover:text-brand-400;
    }
  }
```

I think it's what we are trying to do in the last couple of PRs - need to refactor the rest later! 🙏🏼

mirhamasala: Though if we don't actually have a `share-article` class - it can be removed.

mirhamasala (approved):
Wonderful! ❤️

PR #1196, opened 2025-03-06
REFACTOR: Social component style variants

barbaraperic:
## 📝 Description

In FFDW we have 2 variants of social media icons (see screenshot below). This PR includes modification to `Social` shared component.

- **Type:**  Refactor

Screenshot 2025-03-06 at 14 22 32


apps/ffdweb-site/src/app/_styles/globals.css
```diff
@@ -269,4 +269,9 @@
   .meta {
     @apply text-neutral-50;
   }
+
+  /* SOCIAL */
+  .social-media-secondary {
```
mirhamasala, line 274: Shouldn't this be primary?

mirhamasala: Do we actually have a `secondary` or is it just `inherit`?

mirhamasala (changes requested):
Wonderful! ❤️ Left a question.

apps/ffdweb-site/src/app/_styles/globals.css
mirhamasala: Pushed a CR for this [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1196/commits/105f05087ba3100552396c341d9d25e02e3205be) (Ignore the data changes - They are fixed in the Scout CR)

PR #1212, opened 2025-03-10
Fix file utils import [skip percy]

CharlyMartin:
## 📝 Description

This PR removes `isValidMarkdownPath` from the shared file utilities and moves it close to where it's used.


apps/ff-site/src/app/_schemas/MarkdownEntryPathSchema.ts
mirhamasala, line 1: Don't we need `MarkdownEntryPathSchema` and `isValidMarkdownPath` ` for `ffdw-site`?

CharlyMartin: My understanding is that the featured entries will be hard-coded for FFDW, and `MarkdownEntryPathSchema` is only used for that purpose.

But maybe my understanding is not accurate?

mirhamasala: Ah, yes! All good then! Thank you. Let Get This Merged! ☀️

PR #1217, opened 2025-03-11
[UXIT-2324] FFDW Blog Header [skip percy]

barbaraperic:
## 📝 Description

This PR includes a Blog header component 

- **Type:** New feature

## 📸 Screenshots

Screenshot 2025-03-11 at 12 51 41
Screenshot 2025-03-11 at 12 51 51
Screenshot 2025-03-11 at 12 52 01

### DIFF
--- a/apps/ff-site/src/app/_components/PageHeader.tsx
+++ b/apps/ff-site/src/app/_components/PageHeader.tsx
@@ -6,13 +6,13 @@ import {
 import { SectionDivider } from '@/components/SectionDivider'
 
 type PageHeaderProps = SharedPageHeaderProps & {
-  sectionDividerText?: string
+  sectionDividerTitle?: string
 }
 
-export function PageHeader({ sectionDividerText, ...props }: PageHeaderProps) {
+export function PageHeader({ sectionDividerTitle, ...props }: PageHeaderProps) {
   return (
     <div className="flex flex-col gap-4">
-      {sectionDividerText && <SectionDivider title={sectionDividerText} />}
+      {sectionDividerTitle && <SectionDivider title={sectionDividerTitle} />}
       <SharedPageHeader {...props} />
     </div>
   )
--- a/apps/ff-site/src/app/blog/page.tsx
+++ b/apps/ff-site/src/app/blog/page.tsx
@@ -46,7 +46,7 @@ export default async function Blog() {
       />
 
       <PageHeader
-        sectionDividerText="Featured"
+        sectionDividerTitle="Featured"
         title={featuredPost.title}
         description={featuredPost.description}
         metaData={[formatDate(featuredPost.publishedOn)]}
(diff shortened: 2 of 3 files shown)


apps/ffdweb-site/src/app/_styles/globals.css
```diff
@@ -281,14 +281,14 @@
 
   /* TAGS */
   .tag-label {
-    @apply border-2;
+    @apply border-1;
```
mirhamasala, line 284: Is `border-1` the same as `border`?

apps/ffdweb-site/src/app/blog/components/PageHeader.tsx
mirhamasala, line 1: I didn't look much into this - but for `ff-site`, I think we use the main `PageHeader` - Is it not possible to do the same for FFDW? Do we need a separate component?

barbaraperic: Yes, you're right! I'll amend :)

mirhamasala (commented):
@barbaraperic @CharlyMartin - Early review: This is looking really good. Loving the shared component pattern. And lots of nice clean ups. 🤩 I do have a question about `SectionDivider` - I don't think it's used nor will be used in `ffdw-site` - meaning we shouldn't move it out to the shared UI library, right?

CharlyMartin:
> @barbaraperic @CharlyMartin - Early review: This is looking really good. Loving the shared component pattern. And lots of nice clean ups. 🤩 I do have a question about `SectionDivider` - I don't think it's used nor will be used in `ffdw-site` - meaning we shouldn't move it out to the shared UI library, right?

@mirhamasala Good point. What do you think of this?

```tsx
// apps/ff-site/src/app/_components/PageHeader.tsx

import {
  PageHeader as SharedPageHeader,
  type PageHeaderProps as SharedPageHeaderProps,
} from '@filecoin-foundation/ui/PageHeader'

import { SectionDivider } from '@/components/SectionDivider'

type PageHeaderProps = SharedPageHeaderProps & {
  sectionDividerText?: string
}

export function PageHeader({ sectionDividerText, ...props }: PageHeaderProps) {
  return (
    <div className="flex flex-col gap-4">
      {sectionDividerText && <SectionDivider title={sectionDividerText} />}
      <SharedPageHeader {...props} />
    </div>
  )
}
```

CharlyMartin (approved):
Nice team rework 🤘

apps/ff-site/src/app/_components/PageHeader.tsx
```diff
@@ -1,114 +1,22 @@
...
-      />
+    <div className="flex flex-col gap-4">
+      {sectionDividerText && <SectionDivider title={sectionDividerText} />}
+      <SharedPageHeader {...props} />
     </div>
   )
 }
+
+PageHeader.Title = SharedPageHeader.Title
+PageHeader.Image = SharedPageHeader.Image
```
mirhamasala, line 22: @CharlyMartin Wouldn't it be cleaner to import `SharedPageHeader.Title` and `SharedPageHeader.Image` directly in the components that need it?

mirhamasala:
@CharlyMartin I love it! 😻

> What do you think of this?

PR #1224, opened 2025-03-12
[UXIT-2335] Extend sitemap [skip percy]

barbaraperic:
## 📝 Description

Moves `generateSitemap` to shared `utils`, and extends FFDW sitemap to `blog`, `digest`, and `projects`.

- **Type:**  New feature
Screenshot 2025-03-12 at 14 42 07


apps/ffdweb-site/src/app/sitemap.ts
```diff
@@ -1,12 +1,29 @@
...
+      },
+      digestArticles: {
+        getData: getDigestArticlesData,
+        basePath: PATHS.DIGEST.path,
+      },
+      projects: {
+        getData: getProjectsData,
+        basePath: PATHS.PROJECTS.path,
+      },
+    },
```
CharlyMartin, line 27: I'm wondering if we could pass an array here instead of an object? We don't use the object keys (`blog`, `digestArticles`, etc.) in the `generateSitemap` function.

Not blocking, just an observation :)

```ts
dynamicRoutes: [
  {
    getData: getBlogPostsData,
    basePath: PATHS.BLOG.path,
  },
  {
    getData: getDigestArticlesData,
    basePath: PATHS.DIGEST.path,
  },
  {
    getData: getEcosystemProjectsData,
    basePath: PATHS.ECOSYSTEM_EXPLORER.path,
  },
  {
    getData: getEventsData,
    basePath: PATHS.EVENTS.path,
  },
]
```

packages/utils/src/generateSitemap.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export default async function generateSitemap({
+  paths,
+  baseUrl,
+  dynamicRoutes,
+}: SitemapConfig) {
+  const staticRoutes = Object.values(paths).map((pathConfig) => ({
+    url: `${baseUrl}${pathConfig.path}`,
+    lastModified: new Date(),
```
CharlyMartin, line 32: I was that we dropped `priority: pathConfig.path === '/' ? 1.0 : 0.7,` in the FFDW sitemap.

Just flagging in case it's not intentional.

CharlyMartin (approved):
Super cool refactor Barbara! 👌👌👌

PR #1225, opened 2025-03-12
Implement Filter Container on FFDW

CharlyMartin:
## 📝 Description

This PR adds `FilterContainer` to the Blog Posts, Projects and Learning Resources on FFDW.

The current implementation works well for blog posts and learning resources, but would need to be adapted for projects, once we implement the Current / Past Partnerships filter.

## 📸 Screenshots
CleanShot 2025-03-13 at 10 11 44@2x
---
CleanShot 2025-03-12 at 14 40 37@2x

### DIFF
--- a/apps/ffdweb-site/src/app/_components/FilterContainer.tsx
+++ b/apps/ffdweb-site/src/app/_components/FilterContainer.tsx
@@ -48,7 +48,9 @@ export function FilterContainer({
 
         <ResultsContainer hasResults={hasResults}>{children}</ResultsContainer>
 
-        <BottomFilterContainer>{bottom}</BottomFilterContainer>
+        <BottomFilterContainer hasResults={hasResults}>
+          {bottom}
+        </BottomFilterContainer>
       </MainContainer>
     </SectionContainer>
   )
@@ -85,16 +87,6 @@ function SecondaryFilterContainer({ children, className }: ContainerProps) {
   return (
     children && (
       <div className={clsx(className, 'w-full md:max-w-56')}>{children}</div>
-    )
-  )
-}
-
-function BottomFilterContainer({ children }: ContainerProps) {
-  return (
-    children && (
-      <div className="mt-16 flex justify-center">
-        <div className="max-w-readable w-full">{children}</div>
-      </div>
     )
   )
 }
@@ -145,5 +137,18 @@ function ResultsContainer({ children, hasResults }: ContainerWithResultsProps) {
     children
   ) : (
     <NoSearchResultsMessage baseDomain={BASE_DOMAIN} />
+  )
+}
+
+function BottomFilterContainer({
+  children,
+  hasResults,
+}: ContainerWithResultsProps) {
+  return (
+    hasResults && (
+      <div className="mt-16 flex justify-center">
+        <div className="max-w-readable w-full">{children}</div>
+      </div>
+    )
   )
 }
--- a/apps/ffdweb-site/src/app/_components/FilterContainer.tsx
+++ b/apps/ffdweb-site/src/app/_components/FilterContainer.tsx
@@ -70,11 +70,12 @@ function MainFilterContainer({ children }: ContainerProps) {
   )
 }
 
-function TopFiltersContainer({ children }: ContainerProps) {
+function TopFiltersContainer({ children, className }: ContainerProps) {
   return (
     <div
       className={clsx(
         topFiltersGap,
+        className,
         'flex flex-col items-center justify-between md:flex-row',
       )}
     >
--- a/apps/ffdweb-site/src/app/_components/FilterContainer.tsx
+++ b/apps/ffdweb-site/src/app/_components/FilterContainer.tsx
@@ -1,39 +1,149 @@
 import type { ReactElement } from 'react'
 
 import { NoSearchResultsMessage } from '@filecoin-foundation/ui/NoSearchResultsMessage'
+import { clsx } from 'clsx'
 
 import { BASE_DOMAIN } from '@/constants/siteMetadata'
 
 type FilterContainerProps = {
   hasResults: boolean
-  search: ReactElement
-  pagination: ReactElement
+  bottom: ReactElement
+  top: {
+    main: ReactElement
+    secondary?: ReactElement
+  }
+  side?: {
+    desktop: ReactElement
+    mobile: ReactElement
+  }
   children: ReactElement
 }
 
+const topFiltersGap = 'gap-4'
+
 export function FilterContainer({
   hasResults,
-  search,
-  pagination,
+  top,
+  bottom,
+  side,
   children,
 }: FilterContainerProps) {
+  const hasSecondaryFilters = Boolean(side?.mobile) || Boolean(top.secondary)
+
   return (
-    <section>
-      <div className="mb-8">
-        <div className="max-w-readable w-full">{search}</div>
+    <SectionContainer hasSidebar={Boolean(side)}>
+      <SideFilterContainer>{side?.desktop}</SideFilterContainer>
+
+      <MainContainer hasSidebar={Boolean(side)}>
+        <TopFiltersContainer>
+          <SecondaryFiltersContainer hasChildren={hasSecondaryFilters}>
+            <SecondaryFilterContainer className="block lg:hidden">
+              {side?.mobile}
+            </SecondaryFilterContainer>
+            <SecondaryFilterContainer>{top.secondary}</SecondaryFilterContainer>
+          </SecondaryFiltersContainer>
+
+          <MainFilterContainer>{top.main}</MainFilterContainer>
+        </TopFiltersContainer>
+
+        <ResultsContainer hasResults={hasResults}>{children}</ResultsContainer>
+
+        <BottomFilterContainer>{bottom}</BottomFilterContainer>
+      </MainContainer>
+    </SectionContainer>
+  )
+}
+
+type ContainerProps = React.ComponentProps<'div'>
+
+function SideFilterContainer({ children }: ContainerProps) {
+  return children && <div className="hidden lg:block">{children}</div>
+}
+
+function MainFilterContainer({ children }: ContainerProps) {
+  return (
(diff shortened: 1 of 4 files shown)


mirhamasala:
Early review - I don't believe pagination should show when there are no results?

CleanShot 2025-03-13 at 10 23 05@2x

apps/ffdweb-site/src/app/_components/FilterContainer.tsx
```diff
@@ -0,0 +1,149 @@
...
+        <BottomFilterContainer>{bottom}</BottomFilterContainer>
+      </MainContainer>
+    </SectionContainer>
+  )
+}
+
+type ContainerProps = React.ComponentProps<'div'>
+
+function SideFilterContainer({ children }: ContainerProps) {
+  return children && <div className="hidden lg:block">{children}</div>
```
mirhamasala, line 60: Shouldn't this check be happening outside of this component?

apps/ffdweb-site/src/app/_components/FilterContainer.tsx
```diff
@@ -0,0 +1,149 @@
...
+      )}
+    >
+      {children}
+    </div>
+  )
+}
+
+function SecondaryFilterContainer({ children, className }: ContainerProps) {
+  return (
+    children && (
```
mirhamasala, line 86: Same [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1225/files#r1993214567)

apps/ffdweb-site/src/app/_components/FilterContainer.tsx
```diff
@@ -0,0 +1,149 @@
...
+      <div className={clsx(className, 'w-full md:max-w-56')}>{children}</div>
+    )
+  )
+}
+
+function BottomFilterContainer({ children }: ContainerProps) {
+  return (
+    children && (
```
mirhamasala, line 94: Same [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1225/files#r1993214567)

apps/ffdweb-site/src/app/_components/FilterContainer.tsx
```diff
@@ -0,0 +1,149 @@
...
+type ContainerWithChildrenProps = ContainerProps & {
+  hasChildren: boolean
+}
+
+function SecondaryFiltersContainer({
+  children,
+  hasChildren,
+}: ContainerWithChildrenProps) {
+  return (
+    hasChildren && (
```
mirhamasala, line 131: Same [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1225/files#r1993214567)

apps/ffdweb-site/src/app/_components/FilterContainer.tsx
```diff
@@ -0,0 +1,149 @@
...
+    </div>
+  )
+}
+
+function TopFiltersContainer({ children }: ContainerProps) {
+  return (
+    <div
+      className={clsx(
+        topFiltersGap,
+        'mb-6 flex flex-col items-center justify-between md:flex-row',
```
mirhamasala, line 76: Can `mb-6` be part of the parent?

apps/ffdweb-site/src/app/_components/FilterContainer.tsx
```diff
@@ -0,0 +1,149 @@
...
+    children && (
+      <div className={clsx(className, 'w-full md:max-w-56')}>{children}</div>
+    )
+  )
+}
+
+function BottomFilterContainer({ children }: ContainerProps) {
+  return (
+    children && (
+      <div className="mt-16 flex justify-center">
```
mirhamasala, line 95: Can `mt-16` be part of the parent?

apps/ffdweb-site/src/app/_components/FilterContainer.tsx
mirhamasala, line 1: I think once some of these nits have been addressed, these components could be standalone and live inside their own files, right?

CharlyMartin:
> Early review - I don't believe pagination should show when there are no results?

@mirhamasala Yes, totally. That was my intention, but I somehow undid it during the refactor. Fixed it!

apps/ffdweb-site/src/app/_components/FilterContainer.tsx
CharlyMartin: Yes, I agree usually, but for layout components, Claude suggested the following, which made sense to me.

> For layout components specifically, I recommend handling the conditional rendering inside the component. This means your component should internally check if children exists and return null if it doesn't.

What do you think?

mirhamasala: Why does it make sense?

apps/ffdweb-site/src/app/_components/FilterContainer.tsx
```diff
@@ -0,0 +1,155 @@
...
+const topFiltersGap = 'gap-4'
+
+export function FilterContainer({
+  hasResults,
+  top,
+  bottom,
+  side,
+  children,
+}: FilterContainerProps) {
+  const hasSecondaryFilters = Boolean(side?.mobile) || Boolean(top.secondary)
```
barbaraperic, line 31: If we pass `mobile`, but **not** `secondary`  - 

we will still render (an empty) `<SecondaryFilterContainer>{top.secondary}</SecondaryFilterContainer>` right? That would prevent the `main` filter to take full width if I'm not wrong...?

apps/ffdweb-site/src/app/_components/FilterContainer.tsx
CharlyMartin: # Checking at Call Site vs. Component Level

## Comparing Both Approaches

### Call Site Checking (Outside Component)

**Advantages:**
- Components remain simpler and more focused on rendering
- More explicit control at the usage point
- Potentially fewer component re-renders
- Follows the principle that components should be predictable in what they render

**Disadvantages:**
- Requires discipline from every developer using the component
- Can lead to repetitive conditional logic across multiple call sites
- May result in inconsistent usage patterns

### Component-Level Checking (Inside Component)

**Advantages:**
- Encapsulates the rendering logic within the component
- Ensures consistent behavior across all usages
- Simplifies the API for consumers
- Follows the principle of self-contained components

**Disadvantages:**
- Components have additional responsibility beyond pure rendering
- May be less obvious to consumers what conditions trigger rendering
- Slightly more complex component implementation

## React Community Practices

In the React community, both approaches are used, but there are some patterns:

1. **UI Libraries** (like Material-UI, Chakra UI) often handle null checks internally to provide a more forgiving API

2. **Application Code** often leans toward call site checking for more explicit control

3. **Compound Components** (like your FilterContainer pattern) often handle their own rendering logic internally

## Recommendation

Both approaches are valid, but I'd recommend:

1. **For layout components that are purely structural** (like your FilterContainer components): Handle null checks internally to make them more robust and easier to use

2. **For content-driven components**: Consider call site checking for more explicit control

3. **Be consistent** within your codebase - pick one approach for similar component types

If you prefer call site checking, you could refactor your components to be pure rendering functions and move the conditional logic to where they're used. This would make the components simpler but require more discipline from developers using them.

The best approach often depends on your team's preferences and your specific use cases. Both approaches can lead to clean, maintainable code when used consistently.

apps/ffdweb-site/src/app/_components/FilterContainer.tsx
CharlyMartin: We don't render it because `SecondaryFilterContainer` checks internally if `children` exists before rendering JSX.

barbaraperic: Screenshot 2025-03-13 at 15 03 05

apps/ffdweb-site/src/app/_components/FilterContainer.tsx
CharlyMartin: How about we do something similar to `Card` instead?

apps/ffdweb-site/src/app/_components/FilterContainer.tsx
CharlyMartin: I think I'm gonna have to rework it into two components, one with and one without sidebar.

apps/ffdweb-site/src/app/_components/FilterContainer.tsx
mirhamasala: That's super interesting. Thank you for sharing. It all makes sense to me. I'd lean for now towards Call Site - just to keep it consistent with the rest of our code. Still, happy to reconsider the approach as we build out this monorepo.

apps/ffdweb-site/src/app/_components/FilterContainer.tsx
CharlyMartin: They are small enough that they can, I think

apps/ffdweb-site/src/app/_components/FilterContainer.tsx
CharlyMartin: @barbaraperic I fixed it :)

apps/ffdweb-site/src/app/_components/FilterContainer.tsx
CharlyMartin: For the Current / Past Partnerships filter on /projects, we will need to create a new layout using pieces from this one (I think)

apps/ffdweb-site/src/app/_components/FilterContainer.tsx
```diff
@@ -0,0 +1,135 @@
...
+            )}
+          </SecondaryFiltersContainer>
+
+          <MainFilterContainer>{top.main}</MainFilterContainer>
+        </TopFiltersContainer>
+
+        {hasResults ? (
+          children
+        ) : (
+          <NoSearchResultsMessage baseDomain={BASE_DOMAIN} />
```
CharlyMartin, line 66: Add 64px top margin on `NoSearchResultsMessage`, as per @filipagr's instructions

PR #1226, opened 2025-03-12
[UXIT-2356] FFDW Blog Post Structured Data [skip percy]

barbaraperic:
## 📝 Description

This PR includes a shared `generateBlogPostStructuredData` function, refactor of Blog Post `generateStructuredData` and `StructuredDataScript` in FFDW Blog Post.

- **Type:** New feature / Refactor


packages/utils/src/generateBlogPostStructuredData.ts
```diff
@@ -0,0 +1,49 @@
+import { SCHEMA_CONTEXT_URL } from './constants/structuredDataConstants'
+
+import type { BlogPosting, Organization, WithContext } from 'schema-dts'
+
+type GenerateBlogPostStructuredDataProps = {
```
mirhamasala, line 5: @CharlyMartin - Would you refactor this? Would you use something like `type GenerateBlogPostStructuredDataProps = Partial<BlogPosting> & {`?

mirhamasala (commented):
@barbaraperic - Looks good to me! But I'd still like Charly's TS input on this here [>>](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1226/files#r1991891343). 🙏🏼

packages/utils/src/generateBlogPostStructuredData.ts
barbaraperic: Thank you both!

PR #1229, opened 2025-03-13
[UXIT-2358] FFDW Homepage Learning Resources Section [skip percy]

barbaraperic:
## 📝 Description

This PR includes a shared `FocusAreaCard` and new section on FFDW Homepage - Learning Resources

- **Type:** New feature  / Refactor

## 📸 Screenshots

Screenshot 2025-03-13 at 10 33 52

Screenshot 2025-03-13 at 10 33 41


packages/ui/src/FocusAreaCard.tsx
```diff
@@ -0,0 +1,58 @@
...
+  )
+}
+
+FocusAreaCard.Icon = function FocusAreaCardIcon({
+  icon,
+}: Pick<FocusAreaCardProps, 'icon'>) {
+  if (!icon) return null
+
+  return (
+    <div className="accent-icon-regular">
```
barbaraperic, line 54: maybe there's a better way to reference this class? ideally it should be `focus-area-card-icon`

apps/ffdweb-site/src/app/(homepage)/constants/learningResources.ts
```diff
@@ -0,0 +1,31 @@
+import { PATHS } from '@/_constants/paths'
```
mirhamasala, line 1:
```suggestion
import { PATHS } from '@/constants/paths'
```

packages/ui/src/FocusAreaCard.tsx
mirhamasala: What about

```
  .accent-icon-regular {
    @apply theme('colors.icon.accent');
  }
```

and

```
.focus-area-card-icon {
   @apply theme('colors.icon.accent');
}
```

Since we do have

```
  --color-icon-primary: theme('colors.brand.primary.300');
  --color-icon-accent: theme('colors.brand.secondary.100');
  --color-icon-subtle: theme('colors.neutral.300'); /* CHECK */
  ```
  
  Happy to rename to `--color-icon-brand-primary` if that makes it clearer?
  
  What do you think?

File: apps/ffdweb-site/src/app/(homepage)/page.tsx [hunk 3]
```diff
@@ -98,7 +100,13 @@ export default async function Home() {
           'Learn how decentralized technologies are creating a better future for a better web. Explore educational content, research papers, tutorials, interactive content, and more resources from FFDW and beyond.',
         ]}
       >
-        <div className="bg-brand-primary-800 grid h-96 w-full grid-cols-3 gap-4" />
+        {/* <CardGrid cols="mdThree">
```
barbaraperic, line 103: Commented out as we [don't want to show cards for now. ](https://filecoinfoundation.slack.com/archives/C06PL3W4FPE/p1741865993744089?thread_ts=1741864069.939209&cid=C06PL3W4FPE)

packages/ui/src/FocusAreaCard.tsx
```diff
@@ -0,0 +1,58 @@
...
+        </Heading>
+        {children}
+      </div>
+    </li>
+  )
+}
+
+FocusAreaCard.Image = function FocusAreaCardImage({
+  image,
+}: Pick<FocusAreaCardProps, 'image'>) {
```
mirhamasala, line 40: At this point, this is not the correct type anymore. Image should be present.

```suggestion
FocusAreaCard.Image = function FocusAreaCardImage({
  image,
}: {
  image: NonNullable<FocusAreaCardProps['image']>
}) {
```

packages/ui/src/FocusAreaCard.tsx
```diff
@@ -0,0 +1,58 @@
...
+export function FocusAreaCard({
+  title,
+  children,
+  image,
+  icon,
+}: FocusAreaCardProps) {
+  return (
+    <li className="focus-area-card">
+      {image && <FocusAreaCard.Image image={image} />}
+      {icon && <FocusAreaCard.Icon icon={icon} />}
```
mirhamasala, line 27: What if both an image and an icon are passed?

mirhamasala (changes requested):
@barbaraperic Some logic issues still, I think. And, there's one more thing, I do think for FFDW the entire card should be clickable so not sure this is the right card to use?

Please have a look, and ping @CharlyMartin for a review when done. 🙏🏼

barbaraperic:
> @barbaraperic Some logic issues still, I think. And, there's one more thing, I do think for FFDW the entire card should be clickable so not sure this is the right card to use?
> 
> Please have a look, and ping @CharlyMartin for a review when done. 🙏🏼

Thank you for the comments! 🙏 I'll look into it, I might have over done it with trying to generalise the component...

CharlyMartin (approved):
@barbaraperic I added a commit to (try to) simplify `IconCardWithCTA` and `CTALink`.

Feel free to revert if it doesn't make sense.

The rest looks good :)

barbaraperic:
> @barbaraperic I added a commit to (try to) simplify `IconCardWithCTA` and `CTALink`.
> 
> Feel free to revert if it doesn't make sense.
> 
> The rest looks good :)

Looks good! thank you :)

PR #1233, opened 2025-03-13
Add Canonicals

CharlyMartin:
## 📝 Description

This PR adds canonical links to FFDW

## 📸 Screenshots

CleanShot 2025-03-13 at 17 11 22@2x

### DIFF
--- a/apps/ffdweb-site/src/app/_utils/createMetadata.ts
+++ b/apps/ffdweb-site/src/app/_utils/createMetadata.ts
@@ -1,12 +1,14 @@
 import type { Metadata } from 'next'
 
 export type MetadataParams = {
+  path: `/${string}`
   metaTitle: string
   metaDescription: string
   overrideTitle?: boolean
 }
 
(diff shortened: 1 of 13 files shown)


apps/ffdweb-site/src/app/_utils/createMetadata.ts
```diff
@@ -1,18 +1,23 @@
 import type { Metadata } from 'next'
 
 export type MetadataParams = {
+  path: `/${string}`
```
CharlyMartin, line 4: Here, I used `/${string}` instead of `PathValues | DynamicPathValues` because in some places we have to do ``${PATHS.BLOG.path}/${data.slug}` as DynamicPathValues,`, which defeats the purpose of checking types.

After the launch, we can rework `createPathConfig` to try to improve types

mirhamasala (approved):
Excellent! 🙌🏼 SEO Slayers.

PR #1239, opened 2025-03-14
[UXIT-2372] Delete descriptions fields from Blog Post data

barbaraperic:
## 📝 Description
 
This PR removes descriptions fields from Blog Post data and truncates description in `PageHeader` and Blog Post cards.

P.S. We could probably do some refactoring since we do a similar thing in the `Card` component. But maybe for another PR...

- **Type:** Refactor

Screenshot 2025-03-17 at 09 45 15
Screenshot 2025-03-17 at 12 08 22

### DIFF
--- a/packages/ui/src/DescriptionText.tsx
+++ b/packages/ui/src/DescriptionText.tsx
@@ -1,31 +1,16 @@
 import clsx from 'clsx'
 
-type DescriptionProps = {
+export type DescriptionProps = {
   isClamped?: boolean
-  children: DescriptionTextType
+  children: React.ReactNode
 }
 
-export type DescriptionTextType = React.ReactNode
-
 export function DescriptionText({ children, isClamped }: DescriptionProps) {
-  if (Array.isArray(children)) {
-    return (
-      <>
-        {children.map((item, index) => (
-          <p
-            className={clsx(isClamped && 'line-clamp-3 text-ellipsis')}
-            key={index}
-          >
-            {item}
-          </p>
-        ))}
-      </>
-    )
-  }
+  const childrenArray = Array.isArray(children) ? children : [children]
 
-  return (
-    <p className={clsx(isClamped && 'line-clamp-3 text-ellipsis')}>
-      {children}
+  return childrenArray.map((item, index) => (
+    <p key={index} className={clsx(isClamped && 'line-clamp-3 text-ellipsis')}>
+      {item}
     </p>
-  )
+  ))
 }
--- a/packages/ui/src/DescriptionText.tsx
+++ b/packages/ui/src/DescriptionText.tsx
@@ -1,16 +1,21 @@
+import clsx from 'clsx'
+
 type DescriptionProps = {
-  className?: string
+  isClamped?: boolean
   children: DescriptionTextType
 }
 
 export type DescriptionTextType = React.ReactNode
 
-export function DescriptionText({ className, children }: DescriptionProps) {
+export function DescriptionText({ children, isClamped }: DescriptionProps) {
   if (Array.isArray(children)) {
     return (
       <>
         {children.map((item, index) => (
-          <p className={className} key={index}>
+          <p
+            className={clsx(isClamped && 'line-clamp-3 text-ellipsis')}
+            key={index}
+          >
             {item}
           </p>
         ))}
@@ -18,5 +23,9 @@ export function DescriptionText({ className, children }: DescriptionProps) {
     )
   }
 
-  return <p className={className}>{children}</p>
+  return (
+    <p className={clsx(isClamped && 'line-clamp-3 text-ellipsis')}>
+      {children}
+    </p>
+  )
 }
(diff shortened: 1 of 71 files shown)


packages/ui/src/PageHeader.tsx
```diff
@@ -46,7 +51,7 @@ export function PageHeader({
           {isFeatured && <TagLabel variant="secondary">Featured</TagLabel>}
           <PageHeader.Title>{title}</PageHeader.Title>
           {metaData && <Meta metaData={metaData} />}
-          {description && <DescriptionText>{description}</DescriptionText>}
+          <Description description={description} />
```
mirhamasala, line 54: Why remove the check?

packages/ui/src/PageHeader.tsx
```diff
@@ -18,6 +18,11 @@ type TitleProps = {
   children: string
 }
 
+type PageHeaderDescriptionProps = {
+  text: string
+  isClamped?: boolean
```
mirhamasala, line 23: What if we moved `isClamped` to `DescriptionText`?

packages/ui/src/PageHeader.tsx
```diff
@@ -97,3 +102,23 @@ PageHeader.Image = function PageHeaderImage({
...
+      <p
+        className={clsx(description.isClamped && 'line-clamp-3 text-ellipsis')}
+      >
+        {description.text}
+      </p>
+    )
+  }
+
+  return <DescriptionText>{description}</DescriptionText>
+}
```
mirhamasala, line 124: I think it would be cleaner if we add `isClamped` to `DescriptionText`. What do you think?

barbaraperic: Yes, I like that!

packages/ui/src/PageHeader.tsx
```diff
@@ -97,3 +101,25 @@ PageHeader.Image = function PageHeaderImage({
...
+
+function Description({
+  description,
+}: {
+  description: NonNullable<PageHeaderDescriptionProp>
+}) {
+  if (
+    typeof description === 'object' &&
+    !Array.isArray(description) &&
+    'text' in description
```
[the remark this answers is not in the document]
barbaraperic: Potentially we could set description type as 

```
export type PageHeaderDescriptionProp =
  | { text: DescriptionTextType }
  | { clampedText: DescriptionTextType }
```

that way we can do something like:

```
function Description({
  description,
}: {
  description: NonNullable<PageHeaderDescriptionProp>
}) {
  if ('clampedText' in description) {
    return (
      <DescriptionText isClamped>{description.clampedText}</DescriptionText>
    )
  }

  return <DescriptionText>{description.text}</DescriptionText>
}
```

And we would need to refactor all PageHeader description props to:

`description={{ text: featuredPost.description }}
`

Just an idea...

packages/ui/src/PageHeader.tsx
```diff
@@ -18,9 +15,7 @@ type TitleProps = {
   children: string
 }
 
-export type PageHeaderDescriptionProp =
-  | DescriptionTextType
-  | { text: string; isClamped?: boolean }
+type PageHeaderDescriptionProp = string | { text: string; isClamped?: boolean }
```
barbaraperic, line 18: @CharlyMartin I believe we also pass an array of strings as well (Digest in FF for example)

CharlyMartin: @barbaraperic Thanks! I pushed a new version as a suggestion. Let me know what you think of it.

It's still a bit messy, but I don't know if we have another option.

CharlyMartin:
The build is still failing, my refactor is not working, will revert!

packages/ui/src/DescriptionText.tsx
```diff
@@ -1,22 +1,16 @@
...
-    )
-  }
+export function DescriptionText({ children, isClamped }: DescriptionProps) {
+  const childrenArray = Array.isArray(children) ? children : [children]
 
-  return <p className={className}>{children}</p>
+  return childrenArray.map((item, index) => (
+    <p key={index} className={clsx(isClamped && 'line-clamp-3 text-ellipsis')}>
+      {item}
+    </p>
```
mirhamasala, line 14: Technically, it could happen that two paragraphs are clamped, right?

packages/ui/src/PageHeader.tsx
```diff
@@ -97,3 +101,19 @@ PageHeader.Image = function PageHeaderImage({
...
+  if (typeof description === 'object' && 'text' in description) {
+    return (
+      <DescriptionText isClamped={description.isClamped}>
+        {description.text}
+      </DescriptionText>
+    )
+  }
+
+  return <DescriptionText>{description}</DescriptionText>
+}
```
mirhamasala, line 119: @CharlyMartin @barbaraperic - Do we still need this now that `isClamped` has been moved?

mirhamasala: As discussed, let's simplify to only accept the object prop.

mirhamasala (approved):
Wonderful teamwork, @barbaraperic @CharlyMartin - Looking much cleaner! Merging! ❤️

PR #1245, opened 2025-03-17
Add generateStaticParams to /blog/slug on FFDW

CharlyMartin:
## 📝 Description

This PR adds `generateStaticParams` to FFDWEB `blog/[slug]/page.tsx` and solves the 404 bug on `/blog/:slug`.

This bug happens because `fileExists` returns false when calling `getMarkdownData` on FFDW. I don't know why that is. I noticed that functions run on different Node versions, which could be the source of the problem?

```ts
export async function generateStaticParams() {
  const entries = await getBlogPostsData()
  return entries.map(({ slug }) => ({ slug }))
}
```


mirhamasala (approved):
Thank you for fixing this! ✨

PR #1253, opened 2025-03-17
[UXIT-2343] FFDW set plausible [skip percy]

barbaraperic:
## 📝 Description

This PR adds Plausible to FFDW and moves `NotFoundAnalytics` to shared UI.

- **Type:** New feature

## 📸 Screenshots

Screenshot 2025-03-17 at 14 44 08


apps/ffdweb-site/src/app/not-found.tsx
```diff
@@ -1,10 +1,12 @@
 import { ErrorMessage } from '@filecoin-foundation/ui/ErrorMessage'
+import { NotFoundAnalytics } from '@filecoin-foundation/ui/NotFoundAnalytics'
 
 import { BASE_DOMAIN } from '@/constants/siteMetadata'
 
 export default function NotFound() {
```
mirhamasala, line 6: Could `NotFound` and `NotFoundAnalytics` be abstracted into the shared UI library as well?

mirhamasala (changes requested):
Looking good! Couple of nits.

Also `next-plausible` need to be added as a peer dependency in `packages/ui/package.json` and as scout CR - `@radix-ui/react-tooltip` is also a peer dependency not a dependency. 🙏🏼

CharlyMartin (approved):

mirhamasala (approved):
Thank you, @barbaraperic & @CharlyMartin! LGTM! ☀️

PR #1277, opened 2025-03-18
[UXIT-2349] FFDW Add Author Bio

barbaraperic:
## 📝 Description

This PR includes a new `AuthorsBio` component and refactor of `formatAuthors` which is now a shared function.

- **Type:** New feature / Refactor

**NOTE**: an edge case we might wanna consider - if only one bio is present, it will still show the names of the both authors in the title (see screenshot below). 

Not sure if we want it that way? Or you think it's fine? Currently we are show `AuthorsBio` if **at least one** bio is present. Quick fix could be if we change the title to something like **About the author/s.**... but also happy to refactor to show only the names of the authors to have bios. Let me know thoughts 😊

Screenshot 2025-03-18 at 17 11 25

## 📸 Screenshots
Screenshot 2025-03-18 at 16 11 27
Screenshot 2025-03-18 at 16 11 35
Screenshot 2025-03-18 at 16 11 55


packages/ui/src/Avatar.tsx
```diff
@@ -1,15 +1,12 @@
...
 
 const IMAGE_SIZE = 32
 const sharedAvatarStyle = 'rounded-full ring avatar-ring'
 
-export type AvatarProps = {
-  firstName: string
-  lastName: string
-  image?: { src: string }
-}
+export type AvatarProps = Author
```
barbaraperic, line 9: or we directly include `Author` as a type in `Avatar` component?

packages/utils/src/types/authorTypes.ts
```diff
@@ -0,0 +1,10 @@
+export type Author = {
+  firstName: string
+  lastName: string
+  image?: { src: string }
+}
+
+export type AuthorExtended = Author & {
```
barbaraperic, line 7: other name suggestions welcome!

packages/utils/src/schemas/DigestArticleFrontmatterSchema.ts
```diff
@@ -12,6 +12,7 @@ export const DigestArticleFrontmatterSchema = DynamicBaseDataSchema.extend({
       'last-name': z.string(),
       image: ImagePropsSchema.optional(),
       company: z.string(),
+      bio: z.string().optional(),
```
mirhamasala, line 15: Isn't optional for FFDW - We might have to extend this Schema.

packages/utils/src/formatAuthors.ts
mirhamasala, line 1: Wonder if we should keep this close to the component.

apps/ffdweb-site/src/app/digest/[slug]/page.tsx
```diff
@@ -46,6 +49,7 @@ export default async function DigestArticle(props: DigestArticleProps) {
           }}
         />
         {content && <MarkdownContent>{content}</MarkdownContent>}
+        {hasAuthorBio && <AuthorsBio authors={authors} />}
```
Comment (mirhamasala) [inline apps/ffdweb-site/src/app/digest/[slug]/page.tsx:52]: FFDW Digest articles should always have an author bio.

apps/ffdweb-site/src/app/digest/[slug]/components/AuthorsBio.tsx
```diff
@@ -0,0 +1,15 @@
...
+import type { AuthorExtended } from '@filecoin-foundation/utils/types/authorTypes'
+
+type AuthorBioProps = {
+  authors: Array<AuthorExtended>
+}
+
+export function AuthorsBio({ authors }: AuthorBioProps) {
+  return (
+    <div className="flex flex-col gap-2 sm:w-2/3">
+      <h2 className="text-base font-bold">{formatAuthors({ authors })}</h2>
```
Comment (mirhamasala) [inline apps/ffdweb-site/src/app/digest/[slug]/components/AuthorsBio.tsx:11]: Not sure an h2 is appropriate for this? Might not need to be a heading?

apps/ffdweb-site/src/app/digest/[slug]/components/AuthorsBio.tsx
```diff
@@ -0,0 +1,15 @@
+import { formatAuthors } from '@filecoin-foundation/utils/formatAuthors'
+import type { AuthorExtended } from '@filecoin-foundation/utils/types/authorTypes'
+
+type AuthorBioProps = {
+  authors: Array<AuthorExtended>
+}
+
+export function AuthorsBio({ authors }: AuthorBioProps) {
+  return (
+    <div className="flex flex-col gap-2 sm:w-2/3">
```
Comment (mirhamasala) [inline apps/ffdweb-site/src/app/digest/[slug]/components/AuthorsBio.tsx:10]: section? footer?

mirhamasala:
@barbaraperic 

> Not sure if we want it that way? Or you think it's fine? Currently we are show AuthorsBio if at least one bio is present. Quick fix could be if we change the title to something like About the author/s.... but also happy to refactor to show only the names of the authors to have bios. Let me know thoughts 😊

We should make the bio field required for FFDW and adapt the article that uses a combined bio to have two bios - please ask Jordan for copy if needed.

packages/utils/src/schemas/DigestArticleFrontmatterSchema.ts
barbaraperic: Ah right, so I guess we have to move it from the shared library, like with the Blog Posts

packages/utils/src/formatAuthors.ts
barbaraperic: Move it back in the AvatarGroup?

apps/ffdweb-site/src/app/digest/[slug]/components/AuthorsBio.tsx
Comment (barbaraperic) [reply apps/ffdweb-site/src/app/digest/[slug]/components/AuthorsBio.tsx:11]: For me it made semantic sense, but happy to go with paragraph

barbaraperic:
> @barbaraperic
> 
> > Not sure if we want it that way? Or you think it's fine? Currently we are show AuthorsBio if at least one bio is present. Quick fix could be if we change the title to something like About the author/s.... but also happy to refactor to show only the names of the authors to have bios. Let me know thoughts 😊
> 
> We need to make the bio field required for FFDW. Also, for the article using a combined bio, we should update it to have two separate bios. Please check with Jordan for the copy if needed. For the layout, we can simply duplicate the design of a single bio section for rendering. 🙏🏼

@mirhamasala like this? 

Screenshot 2025-03-19 at 11 43 23

packages/utils/src/formatAuthors.ts
barbaraperic: I guess if we going with separate bios, no need to using `formatAuthors` :)  So I can revert back the changes

packages/utils/src/schemas/DigestArticleFrontmatterSchema.ts
mirhamasala: No, I don't think we need to move it. FF doesn't use `bio` so we could just extend this Schema with `bio` for FFDW?

mirhamasala: Actually, I'm really sorry. Come to think of it, I think we should make `bio` optional .. 😔

mirhamasala (approved):
Super nit CR. But this is excellent - Thank you for your patience. ❤️

PR #1298, opened 2025-03-24
Fix Cypress Metadata Test on FFDW [skip percy]

CharlyMartin:
## 📝 Description

This PR simplifies how we handle metadata on FFDW (and FF) by removing open graph configuration from the CMS and defaulting to the page’s meta title and description for social metadata.

I’m considering opening a second PR to standardize the `createMetaData` function and share it between both apps, so they rely on the same API.

## 🛠️ Key Changes

- Removes `open_graph_config` from both `config.yml` and SEO Zod schema
- Uses implicit and Twitter and Open Graph titles (instead of explicitly passing it)
- Fixes the Cypress bag on FFDW - https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1298/commits/f5e45b62bb99734b55f773e83b36ac4bce6d4177

## 🧪 How to Test

- Run cypress tests on both apps

| FF | FFDW |
|--------|--------|
| CleanShot 2025-03-25 at 11 36 30@2x | CleanShot 2025-03-25 at 11 37 31@2x |


barbaraperic (commented):
Just an idea... if we group `cy.get` functions, and use scoped selectors I believe we can reduce the execution time of the test.

Something like:

``` ts
  cy.get('head').within(() => {
    // Basic metadata group
    cy.get('title,meta[name="description"],link[rel="canonical"]').should(
      ($elements) => {
        expect($elements.eq(0)).to.have.text(metaTitle)
        expect($elements.eq(1)).to.have.attr('content', description)
        expect($elements.eq(2)).to.have.attr('href', fullUrl)
      },
    )

    // Open Graph metadata group
    cy.get('meta[property^="og:"]').should(($elements) => {
      expect($elements.filter('[property="og:title"]')).to.have.attr(
        'content',
        metaTitle,
      )
      expect($elements.filter('[property="og:description"]')).to.have.attr(
        'content',
        description,
      )
      expect($elements.filter('[property="og:image"]')).to.have.attr('content')
    })

    // Twitter metadata group
    cy.get('meta[name^="twitter:"]').should(($elements) => {
      expect($elements.filter('[name="twitter:title"]')).to.have.attr(
        'content',
        metaTitle,
      )
      expect($elements.filter('[name="twitter:description"]')).to.have.attr(
        'content',
        description,
      )
      expect($elements.filter('[name="twitter:image"]')).to.have.attr('content')
    })
  })
```

BEFORE
Screenshot 2025-03-25 at 12 41 24

AFTER
Screenshot 2025-03-25 at 12 41 39

apps/ffdweb-site/src/app/digest/[slug]/page.tsx
```diff
@@ -83,10 +83,7 @@ export async function generateMetadata(props: DigestArticleProps) {
     title: { absolute: `${seo.title} | ${ORGANIZATION_NAME_SHORT}` },
     description: seo.description,
     image: seo.image || image?.src || graphicsData.digest.data.src,
-    openGraph: {
-      type: 'article',
-      ...seo.openGraph,
-    },
+    openGraph: { type: 'article' },
```
Comment (barbaraperic) [inline apps/ffdweb-site/src/app/digest/[slug]/page.tsx:86]: Should we add `<meta property="og:type" content="article">` to FF pages (blog, ecosystem, digest) as well?

CharlyMartin:
> Just an idea... if we group `cy.get` functions, and use scoped selectors I believe we can reduce the execution time of the test.

@barbaraperic this is cool, thank you! 🙏

I'm gonna add a ticket to implement and test this setup in another PR. I want to merge this one soon because it's currently blocking content PRs.

apps/ffdweb-site/src/app/digest/[slug]/page.tsx
Comment (CharlyMartin) [reply apps/ffdweb-site/src/app/digest/[slug]/page.tsx:86]: https://www.notion.so/filecoin/SHARED-Share-createdMetaData-function-between-both-apps-1c17631f282580dc8783c3824976734c?pvs=4

PR #1301, opened 2025-03-25
Move Listbox to shared UI package

CharlyMartin:
## 📝 Description

This PR moves the `Listbox` components from FF to our shared UI package in preparation for adding filters to the FFDW index pages.


packages/ui/src/Listbox/ListboxButton.tsx
```diff
@@ -34,9 +34,9 @@ export function ListboxButton({
 }: ListboxButtonProps) {
   return (
     <HeadlessUIListboxButton
+      aria-invalid={hasError}
```
barbaraperic, line 37: Nice!

apps/ff-site/src/app/_styles/globals.css
```diff
@@ -334,4 +334,25 @@
   .page-header-image {
     @apply rounded-lg border border-brand-100;
   }
+
+  /* LISTBOX */
+  .listbox-button {
+    @apply rounded-lg border border-brand-300 bg-brand-800 text-brand-300 hover:border-current hover:text-brand-400;
```
barbaraperic, line 340: I believe `border-brand-300` is not need as we have `text-brand-300`? it should inherit

barbaraperic (approved):
🚀

apps/ff-site/src/app/_styles/globals.css
CharlyMartin: Indeed, thank you!

PR #1302, opened 2025-03-26
Share useFilter [skip percy]

CharlyMartin:
## 📝 Description

This PR moves `useFilter` and its related constants to our shared hooks package, in preparation for adding filters to the FFDW index pages.

It also completes the migration of utility types to the shared types package.

### DIFF
--- a/packages/hooks/src/useFilter/index.ts
+++ b/packages/hooks/src/useFilter/index.ts
@@ -1,13 +1,6 @@
-import { useFilter } from './useFilter'
-import {
-  DEFAULT_CATEGORY_FILTER_OPTION,
-  DEFAULT_LOCATION_FILTER_OPTION,
-  DEFAULT_FILTER_ID,
-} from './filterConstants'
-
 export {
-  useFilter,
   DEFAULT_CATEGORY_FILTER_OPTION,
-  DEFAULT_LOCATION_FILTER_OPTION,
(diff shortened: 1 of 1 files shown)


packages/hooks/src/useFilter/index.ts
```diff
@@ -0,0 +1,13 @@
+import { useFilter } from './useFilter'
+import {
+  DEFAULT_CATEGORY_FILTER_OPTION,
+  DEFAULT_LOCATION_FILTER_OPTION,
+  DEFAULT_FILTER_ID,
+} from './filterConstants'
```
barbaraperic, line 6: We can also do this? 

``` ts
export {
  DEFAULT_CATEGORY_FILTER_OPTION,
  DEFAULT_FILTER_ID,
  DEFAULT_LOCATION_FILTER_OPTION,
} from './filterConstants'
export { useFilter } from './useFilter'
```

apps/ff-site/src/app/_types/utils.ts
barbaraperic, line 1: empty file?

barbaraperic (dismissed):
🙌

packages/hooks/src/useFilter/index.ts
CharlyMartin: Ah very cool, I didn't know that was possible, thank you!

PR #1304, opened 2025-03-26
Share FilterListbox

CharlyMartin:
## 📝 Description

This PR moves `FilterListbox` in `packages/ui` to be able to use it in FFDW.

It also moves `entryMatchesCategoryQuery` in `packages/hooks/`, alongside `useFilter`, and re-organises the `useFilter` folder.


barbaraperic (approved):
🤘

PR #1305, opened 2025-03-26
Implement Category Filter on FFDW

CharlyMartin:
## 📝 Description

This PR implements the category filter on FFDW

## 🛠️ Key Changes

- Add FilterListbox style `apps/ffdweb-site/src/app/_styles/globals.css`
- Move `useListboxQueryState` to shared hooks

## 📸 Screenshots

CleanShot 2025-03-26 at 16 53 44@2x
---
CleanShot 2025-03-26 at 16 54 22@2x

### DIFF
--- a/apps/ffdweb-site/src/app/_styles/globals.css
+++ b/apps/ffdweb-site/src/app/_styles/globals.css
@@ -343,6 +343,6 @@
   }
 
   .listbox-option {
-    @apply ui-active:bg-brand-primary-500 text-brand-primary-100;
+    @apply ui-active:bg-brand-primary-500;
   }
 }
--- a/apps/ffdweb-site/src/app/_styles/globals.css
+++ b/apps/ffdweb-site/src/app/_styles/globals.css
@@ -335,7 +335,7 @@
 
   /* LISTBOX */
   .listbox-button {
-    @apply text-brand-primary-300 hover:text-brand-primary-400 focus:text-brand-primary-100 border border-current;
+    @apply text-brand-primary-300 hover:text-brand-primary-400 focus:text-brand-primary-100 border;
   }
 
   .listbox-options {
--- a/apps/ffdweb-site/src/app/_styles/globals.css
+++ b/apps/ffdweb-site/src/app/_styles/globals.css
@@ -332,4 +332,17 @@
   .accent-icon-fill {
     @apply bg-brand-secondary-100 text-neutral-950;
   }
+
+  /* LISTBOX */
+  .listbox-button {
+    @apply text-brand-primary-300 hover:text-brand-primary-400 focus:text-brand-primary-100 border border-current;
+  }
+
+  .listbox-options {
+    @apply border-brand-primary-300 text-brand-primary-100 border bg-neutral-950;
+  }
+
+  .listbox-option {
+    @apply ui-active:bg-brand-primary-500 text-brand-primary-100;
+  }
 }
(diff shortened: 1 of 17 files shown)


barbaraperic (commented):
There's a tiny issue with the pagination layout if there are more than 9 pages

Screenshot 2025-03-27 at 09 31 50

apps/ffdweb-site/src/app/_styles/globals.css
```diff
@@ -332,4 +332,17 @@
...
     @apply bg-brand-secondary-100 text-neutral-950;
   }
+
+  /* LISTBOX */
+  .listbox-button {
+    @apply text-brand-primary-300 hover:text-brand-primary-400 focus:text-brand-primary-100 border border-current;
+  }
+
+  .listbox-options {
+    @apply border-brand-primary-300 text-brand-primary-100 border bg-neutral-950;
```
barbaraperic, line 342: do we need `border`?  🤔

apps/ffdweb-site/src/app/_styles/globals.css
```diff
@@ -332,4 +332,17 @@
...
+  }
+
+  .listbox-option {
+    @apply ui-active:bg-brand-primary-500 text-brand-primary-100;
+  }
```
barbaraperic, line 347:
```suggestion
  .listbox-option {
    @apply ui-active:bg-brand-primary-500;
  }
```

I think we can remove text color, it inheriths from` listbox-options`

apps/ffdweb-site/src/app/_styles/globals.css
```diff
@@ -332,4 +332,17 @@
   .accent-icon-fill {
     @apply bg-brand-secondary-100 text-neutral-950;
   }
+
+  /* LISTBOX */
+  .listbox-button {
+    @apply text-brand-primary-300 hover:text-brand-primary-400 focus:text-brand-primary-100 border border-current;
+  }
```
barbaraperic, line 339: don't think we need `border-current?` 🤔

barbaraperic (approved):
Left a few comments, working really well 👌 One thing I noticed tho - not blocking in any way, just a nit. 

If you're on page 5, you have the url "blog?page=5", if you then filter by category "interviews", I get url "blog?page=5&category=interviews". It correctly shows all posts will interviews label, but we only have 1 page. Maybe it would be good to reset the page parameter when a filter is applied? But again, don't know how much work that is, and is it worth it. Will leave it up to you! :)

Screenshot 2025-03-27 at 09 01 21

apps/ffdweb-site/src/app/_styles/globals.css
CharlyMartin: I think so, it sets the border width https://tailwindcss.com/docs/border-width

CharlyMartin:
> Left a few comments, working really well 👌 One thing I noticed tho - not blocking in any way, just a nit.
> 
> If you're on page 5, you have the url "blog?page=5", if you then filter by category "interviews", I get url "blog?page=5&category=interviews". It correctly shows all posts will interviews label, but we only have 1 page. Maybe it would be good to reset the page parameter when a filter is applied? But again, don't know how much work that is, and is it worth it. Will leave it up to you! :)

Thanks for flagging @barbaraperic! This is indeed problematic. I thought we already had something in place to prevent that, but apparently not. Will look into it!

apps/ffdweb-site/src/app/_styles/globals.css
CharlyMartin: It looks like we don't indeed

PR #1310, opened 2025-03-27
Refactor CategoryFilter to use FilterSidebar component

CharlyMartin:
## 📝 Description

This PR implements filtering by resource type on the learning resources page on FFDW.

## 🛠️ Key Changes

1. Moves `CategorySidebar` to our shared UI package and rename it `FilterSidebar`
2. Adds `filter-sidebar-button` to both `globals.css` files
3. Creates `useListboxOptions` in `packages/hooks`

## 📸 Screenshots

CleanShot 2025-03-27 at 15 12 02
---
CleanShot 2025-03-27 at 15 12 21


packages/utils/src/constants/urlParamsConstants.ts
```diff
@@ -3,3 +3,4 @@ export const LOCATION_KEY = 'location'
 export const PAGE_KEY = 'page'
 export const SEARCH_KEY = 'search'
 export const SORT_KEY = 'sort'
+export const RESOURCE_TYPE_KEY = 'resource'
```
barbaraperic, line 6: maybe `RESOURCE_KEY` so it follows the above pattern?

packages/ui/src/FilterSidebar.tsx
```diff
@@ -0,0 +1,84 @@
...
+}
+
+FilterSidebar.Container = function List({
+  children,
+  as: Component = 'ul',
+}: FilterContainerProps) {
+  return (
+    <Component className={clsx('space-y-3', TOUCH_TARGET.touchAreaOffset)}>
+      {children}
+    </Component>
```
barbaraperic, line 58: Should we keep `ul` as default (and remove `as`) since we specifically use `li` as children?

packages/ui/src/FilterSidebar.tsx
```diff
@@ -0,0 +1,84 @@
...
+    <button
+      data-selected={isSelected}
+      className={clsx(
+        'filter-sidebar-button focus:brand-outline w-full cursor-pointer text-pretty px-4 py-3 text-left',
+        typeof count === 'number' && 'inline-flex items-baseline gap-2',
+        TOUCH_TARGET.touchAreaPadding,
+      )}
+      onClick={handleClick}
+    >
+      <span className="text-nowrap font-bold">{name}</span>
```
barbaraperic, line 78: regarding `text-nowrap` - should we handle a case where there's a long resource name? 

Screenshot 2025-03-27 at 17 44 24

barbaraperic (commented):
Should the filters be other way around? 🤔

Screenshot 2025-03-27 at 17 24 46

Screenshot 2025-03-27 at 17 52 08

packages/ui/src/FilterSidebar.tsx
CharlyMartin: Yes, probably!

barbaraperic (approved):

packages/ui/src/FilterSidebar.tsx
CharlyMartin: Hm, a bit tricky. Since we don't have that problem I'd say no :)

PR #1320, opened 2025-04-01
Move useSort to shared hooks package

CharlyMartin:
## 📝 Description

This PR moves `useSort` to our shared hooks package, in preparation for the Current/Past Partnerships filter on FFDW. `useSort` will be used for the toggle mechanism, as our `useFilter` is not suited for this implementation.

The filter will be implemented in a second PR.

## 🛠️ Key Changes

- Updates `SortConfig` to use `id` and `name`, instead of `key` and `label`, to be consistent with `OptionType`

## 📌 To-Do Before Merging

- [ ] Rename `useSort` to something more generic as this function does more than sorting. How about `useTransform`?


mirhamasala (approved):
@CharlyMartin  ~ Looks good to me. 🙏🏼 Only not so sure about `useTransform` - What does it transform? Does it do too much? Do we need to break it up?

CharlyMartin:
> @CharlyMartin ~ Looks good to me. 🙏🏼 Only not so sure about `useTransform` - What does it transform? Does it do too much? Do we need to break it up?

How about one of these?
1. `useListHandler`
2. `useListTransform`
3. `useListRefine`

The hook accepts a list of entries and re-arranges it depending on which config applies.

PR #1322, opened 2025-04-02
Add blog graphic and update metadata [skip percy]

CharlyMartin:
## 📝 Description

This PR adds a graphic image for blog. It is used a meta image for `/blog` and a fallback meta image for `/blog/slug`


mirhamasala (approved):
Fallback all the things. ☀️

PR #1323, opened 2025-04-02
Implement Partnership Toggle Filter on Projects

CharlyMartin:
## 📝 Description

This PR adds the Current/Past Partnerships filter to FFDW projects

## 🛠️ Key Changes

- Creates `PartnershipToggleFilter`
- Creates `ProjectsFilterContainer`, reusing some parts of `FilterContainer`

## 📸 Screenshots

CleanShot 2025-04-02 at 16 22 18


packages/utils/src/constants/urlParamsConstants.ts
```diff
@@ -4,3 +4,4 @@ export const PAGE_KEY = 'page'
 export const SEARCH_KEY = 'search'
 export const SORT_KEY = 'sort'
 export const RESOURCE = 'resource'
```
mirhamasala, line 6: Is this not a key? 

```suggestion
export const RESOURCE_KEY = 'resource'
```

mirhamasala: Does it make sense for FFDW specific constants to live in `packages`?

mirhamasala (changes requested):
@CharlyMartin ~ Code is looking great. ☀️ Left a nit. One thing that needs looking into though is the layout shift that happens when toggling between the options. I assume it has to do with the font-weight changing.

packages/utils/src/constants/urlParamsConstants.ts
CharlyMartin: I don't mind it, they are just generic prefixes, nothing too "unique".

CharlyMartin:
> @CharlyMartin ~ Code is looking great. ☀️ Left a nit. One thing that needs looking into though is the layout shift that happens when toggling between the options. I assume it has to do with the font-weight changing.

Indeed. We could add a CSS transition for a smoother shift. Or make both texts the same fonts and play with colour contrast to show selected/unselected. What do you think?

CharlyMartin:
Todo
- [ ] Add a partnership status pill on the `project/slug` page

mirhamasala (approved):
Happy to merge ☀️ ~ Just a note to maybe discuss with Filipa: This doesn't look clickable to me. 🙈

PR #1324, opened 2025-04-04
Add category select to FFDW projects [skip percy]

CharlyMartin:
## 📝 Description

This PR adds a category filter to the projects page on FFDW

## 🛠️ Key Changes

- Adds optional `optionsPosition` prop to `FilterListbox`

## 📸 Screenshots

| lg | xs |
|--------|--------|
| CleanShot 2025-04-04 at 09 51 08 | CleanShot 2025-04-04 at 09 54 24 |

### DIFF
--- a/apps/ffdweb-site/src/app/projects/components/ProjectsContent.tsx
+++ b/apps/ffdweb-site/src/app/projects/components/ProjectsContent.tsx
@@ -63,6 +63,7 @@ export function ProjectsContent({ projects }: ProjectsContentProps) {
       top={{
         main: <PartnershipToggleFilter />,
         secondary: <Search query={searchQuery} />,
+        tertiary: <Search query={searchQuery} />,
       }}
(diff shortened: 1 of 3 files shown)


mirhamasala (changes requested):
@CharlyMartin ~ Code is looking clean! ✨ There's a layout shift though between the search and categories element when changing categories. Also, `Science and Environment` doesn't seem to have any projects – Might need to check with @jordanlynnf whether we should remove it?

CharlyMartin:
> @CharlyMartin ~ Code is looking clean! ✨ There's a layout shift though between the search and categories element when changing categories. Also, `Science and Environment` doesn't seem to have any projects – Might need to check with @jordanlynnf whether we should remove it?

Indeed, I didn't notice the shift, thanks for catching it :) I'll message @jordanlynnf about it!

PR #1336, opened 2025-04-07
[UXIT-2297] FFDW lighthouse analysis on SEO, Best Practices and Accessibility [skip percy]

barbaraperic:
### Accessibility score

1. **Heading Hierarchy Issue** ⚠️ 
We get a slightly lower score on accessibility on pages where we use `Card` elements without a `PageSection` title because "Heading elements are not in a sequentially-descending order". Meaning, on pages without a section title, we're using an H1 for the page title, then jumping directly to H3 for card titles. This violates the sequential heading order requirement (H1 → H3, skipping H2).
   
I didn't make Card modifications, as I assume we might add PageSections at some point. But there are some potential solutions we can consider, such as dynamically adjust Card heading levels based on page context.

2. **Missing Accessible Names** ✅ (Fixed)
   This PR adds an `aria-label` to `CopyToClipboard` as we were getting an accessibility error "Buttons do not have accessible name". Error is gone after the implementation.

Screenshot 2025-04-07 at 15 47 07

### SEO score

**Generic Link Text** ⚠️ 

For the SEO, we get a slightly lower score because we use "Read More" CTA on Card, and they prefer descriptive text.
Screenshot 2025-04-07 at 15 44 27

I left it as is, as this involves design decisions.

#### Homepage
Screenshot 2025-04-07 at 15 13 22

#### About
Screenshot 2025-04-07 at 15 13 48

#### Projects
Screenshot 2025-04-07 at 15 39 00

#### Projects/[slug]
Screenshot 2025-04-07 at 15 16 59

#### Learning Resources
Screenshot 2025-04-07 at 15 16 27

#### Digest
Screenshot 2025-04-07 at 15 22 21

#### Digest/[slug]
Screenshot 2025-04-07 at 15 27 14

#### FAQs
Screenshot 2025-04-07 at 15 28 47


packages/ui/src/CopyToClipboard.tsx
```diff
@@ -17,6 +17,7 @@ import { NOTIFICATION_DIALOG_DURATION_MS } from '@filecoin-foundation/utils/cons
 type CopyToClipboardProps = {
   text: string
   notificationTitle: string
+  'aria-label'?: string
```
mirhamasala, line 20: Why not `ariaLabel`?

packages/ui/src/ShareArticle.tsx
```diff
@@ -47,13 +47,14 @@ export function ShareArticle({
           <CopyToClipboard
             text={articleUrl}
             notificationTitle="Link copied to clipboard!"
+            aria-label={`Copy link to article: ${articleTitle}`}
```
mirhamasala, line 50: Reads a bit funny. What about `Copy link to the article titled "${articleTitle}"` or `Copy article link: ${articleTitle}`

mirhamasala (approved):
@barbaraperic -For the hierarchy heading issue, I pushed a couple of commits. I’ll leave it up to you whether you’d prefer to pull them out into a separate PR or merge them as part of this one.

For the link text issue, do we have a ticket for that? If not, could we add it under follow-up work so it doesn’t get lost? Thanks!

packages/ui/src/CopyToClipboard.tsx
barbaraperic: ✅

packages/ui/src/ShareArticle.tsx
barbaraperic: ✅

PR #1373, opened 2025-04-21
Move metadata Cypress tests to shared package

CharlyMartin:
## 📝 Description

This PR moves the Cypress metadata testing function into a shared package so it can be reused later on FF (in a second PR).

## 📸 Screenshots

CleanShot 2025-04-21 at 15 22 26


mirhamasala (approved):
@CharlyMartin ~ Exciting move! 🙌🏼✨

PR #1377, opened 2025-04-22
Use shared testPageMetaData testing function in FF (1/2)

CharlyMartin:
## 📝 Description

This PR implements the shared `testPageMetaData` testing function in FF.

The slug pages have not implemented the new function yet, as the fetching logic needs to be reworked to match the new setup. See https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1378.

## 🛠️ Key Changes

- Adds Cypress tasks `getPageFrontmatterSeo` to read SEO attributes from Markdown pages
- Moves `ROOT_METADATA` out of the main layout and into `@/constants/siteMetadata`
- Deletes `_govhub_page_spec.cy.ts` as the page doesn't exist anymore
- Adds metadata tests for the ecosystem-explorer project form

## ⚙️ Still to do

- [x] Rework the `cypress/e2e/single_*_page_spec.cy.ts`
- [x] Share `cy.percySnapshot()` between both apps

## ❓Open Questions

- [x] Should we remove `verifyLinks` since we have the link checker running every week?
- [x] Should we remove the govhub config from the CMS and `PATHS`?

## 📸 Screenshots

CleanShot 2025-04-22 at 14 47 05@2x


apps/ff-site/cypress/tasks/getPageFrontmatterSeo.ts
```diff
@@ -0,0 +1,17 @@
+import fs from 'fs/promises'
+
+import matter from 'gray-matter'
+
+import { MARKDOWN_EXTENSION } from '../../src/app/_constants/paths'
```
mirhamasala, line 5: Do we use aliases in tests?

apps/ff-site/cypress/tasks/getPageFrontmatterSeo.ts
```diff
@@ -0,0 +1,17 @@
...
+export type PageFrontmatterSeo = {
+  title: string
+  description: string
+}
+
+export async function getPageFrontmatterSeo(contentPath: string) {
+  const file = await fs.readFile(contentPath + MARKDOWN_EXTENSION, 'utf8')
+  const { data } = matter(file)
+
+  return data.seo as PageFrontmatterSeo
```
mirhamasala, line 16: 😳

mirhamasala (approved):
@CharlyMartin - Looking great! 🙌🏼✨

Happy to remove the GovHub stuff!

Regarding the removal of verifyLinks, I’m torn. I don’t love how long it takes to run, and it currently runs on every page regardless of whether that page has changed, which isn’t efficient.

We do have the GitHub Action in place, but it’s only running weekly right now, which feels too infrequent for the pages.

What I’d really like to explore, though, is perhaps using the same link-checking library as part of our CI workflow, but scoped only to the pages that were actually modified. I’m not sure how feasible that is. It would probably require a new ticket and a bit of investigation.

That said, I do really value what `verifyLinks` gives us: confidence that the main links on every page are working. Broken links are such a bad user experience on the main pages. So while I’m okay moving away from the current setup, I think we need a better, more targeted solution in its place.

apps/ff-site/cypress/tasks/getPageFrontmatterSeo.ts
CharlyMartin: Ah yeah 😅 Not sure it's helpful, these are not really type-safe

CharlyMartin:
@mirhamasala That makes sense, thanks for the detailed explanation!

PR #1380, opened 2025-04-24
[UXIT-2533] FFDW Lighthouse “Links rely on color to be distinguishable” issue

barbaraperic:
## 📝 Description

Removing `!font-normal` resolves the Lighthouse issue: _low-contrast text is difficult or impossible for many users to read. Link text that is discernible improves the experience for users with low vision._

- **Type:** Bug fix 

## 📸 Screenshots

Screenshot 2025-04-24 at 07 45 36

### DIFF
--- a/apps/ffdweb-site/src/app/_styles/globals.css
+++ b/apps/ffdweb-site/src/app/_styles/globals.css
@@ -338,7 +338,7 @@
 
   /* TEXT LINKS */
   .text-link {
-    @apply text-brand-primary-300! hover:text-brand-primary-500! focus:brand-outline! font-normal! no-underline! hover:underline!;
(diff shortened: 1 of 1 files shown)


mirhamasala (approved):
Well spotted! 🙌🏼 We can also now run [Unlighthouse](https://unlighthouse.dev/) reports locally in dev mode, by the way. We can talk about it later today, but I think it could be pretty handy! ❤️

barbaraperic:
> Well spotted! 🙌🏼 We can also now run [Unlighthouse](https://unlighthouse.dev/) reports locally in dev mode, by the way. We can talk about it later today, but I think it could be pretty handy! ❤️

Ah nice! Excited to use it!

PR #1422, opened 2025-05-12
[FF][UXIT-2068] Refactor blog post data fetching to use shared cache and bulk markdown read [skip percy]

mirhamasala:
## 📝 Description
Refactors the blog post data utility to memoize the result of `getAllMarkdownData` during runtime, reducing repeated filesystem reads in Vercel's serverless environment. This aligns with Next.js best practices for module-level caching and improves performance in production while ensuring accurate data during development.

* **Type:** Refactor / Performance Optimization

## 🛠️ Key Changes
* Replaced per-post file read with a bulk `getAllMarkdownData` call and `.find()` lookup for individual post access
* Added a module-level cache (`cachedPosts`) to store transformed post data across invocations
* Introduced dev-mode bypass using `process.env.NODE_ENV` to ensure hot reloads reflect content changes
* Updated getBlogPostData and `getBlogPostsData` to support this shared caching strategy

## 📌 To-Do Before Merging
- [ ] Confirm that dev changes to markdown files are picked up without server restart

## ✏️ Note

Normally, using `.find()` over all blog posts for a single post would be inefficient, especially as content grows. However, in this case:
- We’re already loading all posts at build time via `generateStaticParams` and `generateMetadata`
- Calls to `getBlogPostData()` are often preceded by or coexist with a call to `getBlogPostsData()` anyway
- We now memoize the post list during runtime, avoiding repeated filesystem access

As a result, using `.find()` here is a pragmatic and low-cost tradeoff that improves performance under both static generation and serverless function reuse.

We’re not “searching the filesystem for every slug” — we’re searching an in-memory array that we’ve already loaded (and cached) efficiently.

### DIFF
--- a/apps/ff-site/src/app/blog/utils/getBlogPostData.ts
+++ b/apps/ff-site/src/app/blog/utils/getBlogPostData.ts
@@ -7,31 +7,41 @@ import { BlogPostFrontmatterSchema } from '../schemas/BlogPostFrontmatterSchema'
 
 const BLOG_DIRECTORY_PATH = PATHS.BLOG.entriesContentPath as string
 
+type RawBlogPost = Awaited<
+  ReturnType<typeof getMarkdownData<typeof BlogPostFrontmatterSchema.shape>>
+>
+export type BlogPostData = Awaited<ReturnType<typeof transformBlogPostData>>
+
+const isDev = process.env.NODE_ENV === 'development'
+
+let cachedPosts: BlogPostData[] | null = null
+
 export async function getBlogPostData(slug: string) {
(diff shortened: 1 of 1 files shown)


mirhamasala:
Now that we’ve confirmed generateStaticParams() handles all the blog post loading we need at build time, we no longer need the local in-memory caching (cachedPosts) implementation. The data is already read once during the build and used to statically generate each page, so runtime memoization adds no real benefit. 

We will, however, look into explicitly setting `export const revalidate` and export const dynamic` in relevant routes to ensure static behavior is enforced and to prevent accidental fallback to dynamic rendering in the future.

PR #1426, opened 2025-05-14
Use shared createMetadata in FF [skip percy]

CharlyMartin:
This PR implements the new shared `createMetadata` function in FF.

> [!NOTE]
> Deployments are failing because of the allocators page.


apps/ff-site/src/app/_data/graphicsData.ts
```diff
@@ -142,4 +142,4 @@ export const graphicsData: Record<string, StaticImageProps> = {
     data: security5,
     alt: 'A futuristic digital shield with a lock icon in the center, surrounded by binary code and glowing blue circular elements, symbolizing cybersecurity and data protection.',
   },
-}
+} as const satisfies Record<string, StaticImageProps>
```
mirhamasala, line 145: Finally too the time to look into this. Love it. I was looking to see if we could enforce type assertions with const over type annotations but doesn't look like it.

mirhamasala (approved):
This is wonderful! ✨

PR #1458, opened 2025-05-21
Add GitHub workflow to detect apps affected by code changes [skip percy]

CharlyMartin:
## 📝 Description

This pull request adds a GitHub action `apps-detect-changes.yml` to determine whether an app is affected by code changes in a each PR. The output is then used by `apps-cypress-percy-test.yml` To only run the tests in the apps with code changes.

This PR also fixes the deps of the UXIT app as it was preventing the [action](https://github.com/FilecoinFoundationWeb/filecoin-foundation/actions/runs/15164730036/job/42639454612) from running properly. 

## 🧪 How to Test

1. Make a change in one or several apps
2. Commit and push it to this PR
3. Make sure that `apps-detect-changes.yml` picks up changes only in the relevant apps
4. Revert the changes

(Tests previously made on a separate PR https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1456)

## 🔖 Resources

- https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/passing-information-between-jobs
- https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#needs-context

### DIFF
--- a/.github/workflows/apps-cypress-percy-test.yml
+++ b/.github/workflows/apps-cypress-percy-test.yml
@@ -11,12 +11,17 @@ on:
     branches: [main]
 
 jobs:
+  detect-changes:
+    uses: ./.github/workflows/apps-detect-changes.yml
+
   cypress-run:
-    if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
+    needs: detect-changes
+    if: ${{ github.event.pull_request.draft == false && fromJSON(needs.detect-changes.outputs.has_changes) == true }}
     strategy:
       matrix:
-        app: [ff-site, ffdweb-site]
+        app: ${{ fromJSON(needs.detect-changes.outputs.matrix) }}
+      fail-fast: false
     defaults:
       run:
         working-directory: apps/${{ matrix.app }}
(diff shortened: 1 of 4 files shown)


.github/workflows/apps-cypress-percy-test.yml
```diff
@@ -68,8 +73,10 @@ jobs:
 
       - name: Run Cypress Tests without Percy
         if: steps.test-type.outputs.skip_percy == 'true'
-        run: npm run test:cypress
+        run: npm run test:cypress || echo "Script missing, skipping this step..."
+        continue-on-error: true
```
CharlyMartin, line 77: This morning I woke up and the first thing I thought was... Will this also cause the tests to fail silently if Cypress throws an error?? To be tested 😄

CharlyMartin:
https://turborepo.com/docs/guides/skipping-tasks

mirhamasala:
@CharlyMartin ~ I moved this to draft until we figured it out. 🙏🏼

PR #1486, opened 2025-06-20
[SHARED][UXIT-2653] Update monorepo README

CharlyMartin:
## 📝 Description

This PR updates the root README.md with information about our Monorepo setup.

It also removes the npm scripts related to turbo commands, in favour of `npx turbo <command>`

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ This monorepo contains the web projects for [Filecoin Foundation](https://fil.or
 
 ## Overview
 
-This codebase includes multiple [workspaces](https://docs.npmjs.com/cli/v8/using-npm/workspaces) for applications and internal packages. Applications are deployed projects while packages centralise shared code used across applications.
+This codebase includes multiple [workspaces](https://docs.npmjs.com/cli/v8/using-npm/workspaces) for applications and internal packages. Applications are deployed projects while packages centralize shared code used across applications.
 
 ### Applications
 
@@ -12,7 +12,7 @@ You can find the applications in the `apps` directory, which include:
 
 - [ff-site](apps/ff-site): The Filecoin Foundation website, live at [fil.org](https://fil.org)
 - [ffdweb-site](apps/ffdweb-site): The Filecoin Foundation for the Decentralized Web website, live at [ffdweb.org](https://ffdweb.org)
-- [uxit](apps/uxit): The public dashboard of the The Filecoin Foundation User Experience Improvement Team, live at [uxit.fil.org](https://uxit.fil.org)
+- [uxit](apps/uxit): The public dashboard of the Filecoin Foundation User Experience Improvement Team, live at [uxit.fil.org](https://uxit.fil.org)
 
 ### Packages
 
@@ -112,7 +112,7 @@ Then, create a `src` directory in the package directory and add your source file
 
 _packages/ui/package.json_
 
-Finally, to use the newly-created package in an application, add it as a dependency `manually` in the application's `package.json` file and run `npm install` from the root of the monorepo.
+Finally, to use the newly-created package in an application, add it as a dependency manually in the application's `package.json` file and run `npm install` from the root of the monorepo.
 
 For instance, to use the `@filecoin-foundation/ui` package in the `ff-site` application, add the following line to the `dependencies` section of the `apps/ff-site/package.json` file:
 


README.md
```diff
@@ -1,31 +1,148 @@
...
+- [ff-site](apps/ff-site): The Filecoin Foundation website, live at [fil.org](https://fil.org)
+- [ffdweb-site](apps/ffdweb-site): The Filecoin Foundation for the Decentralized Web website, live at [ffdweb.org](https://ffdweb.org)
+- [uxit](apps/uxit): The public dashboard of the The Filecoin Foundation User Experience Improvement Team, live at [uxit.fil.org](https://uxit.fil.org)
+
+### Packages
+
+You can find the packages in the `packages` directory. These packages include components, hooks, types, utility functions and configuration shared accross the applications.
+
+> [!NOTE]
+> They are are [just-in-time](https://turborepo.com/docs/core-concepts/internal-packages#just-in-time-packages) packages, meaning they are built by the application that uses them. They do not produce a compiled JavaScript output and are not published anywhere.
```
mirhamasala, line 22:
```suggestion
> They are [just-in-time](https://turborepo.com/docs/core-concepts/internal-packages#just-in-time-packages) packages, meaning they are built by the application that uses them. They do not produce a compiled JavaScript output and are not published anywhere.
```

README.md
```diff
@@ -1,31 +1,148 @@
...
+
+Or the shorthand syntax:
+
+```bash
+npm install jest -w 'apps/ff-site'
+```
+
+This will update both the `apps/ff-site/package.json` and the root `package-lock.json` file.
+
+We also use [syncpack](https://www.npmjs.com/package/syncpack) to occasionally check and fix version mismatch across the monorepo with the following commands:
```
mirhamasala, line 89: `npm run check:versions` is actually included in our Husky pre-push hook, so version checks run automatically before every push. I do still like for us to mention the commands here explicitly so people know how to run them manually.

README.md
```diff
@@ -1,31 +1,148 @@
...
+
+You can find the applications in the `apps` directory, which include:
+
+- [ff-site](apps/ff-site): The Filecoin Foundation website, live at [fil.org](https://fil.org)
+- [ffdweb-site](apps/ffdweb-site): The Filecoin Foundation for the Decentralized Web website, live at [ffdweb.org](https://ffdweb.org)
+- [uxit](apps/uxit): The public dashboard of the The Filecoin Foundation User Experience Improvement Team, live at [uxit.fil.org](https://uxit.fil.org)
+
+### Packages
+
+You can find the packages in the `packages` directory. These packages include components, hooks, types, utility functions and configuration shared accross the applications.
```
mirhamasala, line 19:
```suggestion
You can find the packages in the `packages` directory. These packages include components, hooks, types, utility functions and configuration shared across the applications.
```

README.md
```diff
@@ -1,31 +1,148 @@
...
+### Adding npm dependencies
+
+The `npm install <dependency>` command will install dependencies for all applications and packages in the monorepo. Generally, you want to add a dependency inside a specific application or package. To do so, run:
+
+```bash
+npm install <dependency> --workspace=<application/package>
+```
+
+> [!NOTE]
+> The install commands should be run from the root of the monorepo. Do not install dependencies inside the application or package directory, otherwise Turborepo will not be not aware of it and will not resolve the dependencies properly.
```
mirhamasala, line 73:
```suggestion
> The install commands should be run from the root of the monorepo. Do not install dependencies inside the application or package directory, otherwise Turborepo will not be aware of it and will not resolve the dependencies properly.
```

README.md
```diff
@@ -1,31 +1,148 @@
...
 
-## Development
+At this point you should be able to run the development server for any of the applications.
+
+### Development
 
-To start the development server, run `npm run dev`, which is an alias for `turbo dev`. This command will start the development server for all applications and packages in the monorepo.
+Running `npx turbo dev` from the root of the monorepo will call the `dev` script for all applications and packages in the monorepo, as defined in each `package.json`. In our case, this will start the development server for all applications in our monorepo, packages do not have a `dev` script.
+
+In 99% of cases, you only want to start the development server for one application at a time. To so start do so, you should run `turbo dev` with a filter flag.
```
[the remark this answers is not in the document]
mirhamasala: We are using both `npx turbo ...` and `turbo` in this explanation. Perhaps we can stay consistent but also explain that if Turbo is installed globally, the shorthand `turbo` can be used.

mirhamasala:
>It also removes the npm scripts related to turbo commands, in favour of npx turbo <command>

@CharlyMartin - I don't know the exact advantage of this, but it would be nice to add the reason to the PR description? Just an example:

```
We use `npx turbo` instead of relying on npm scripts or globally installed Turbo. This ensures the correct version defined in the project is used and avoids script indirection.
```

But also it make me think whether we intended this? Leave it up to you.

mirhamasala (changes requested):
Good first pass, @CharlyMartin. I’ve left a few comments for clarification and cleanup. Once those are addressed, I’ll do a second review. Thanks!

README.md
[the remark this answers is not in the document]
CharlyMartin: I prefer to use `npx turbo` so that it always runs the project's version (if installed) and avoids mismatch.

README.md
```diff
@@ -1,31 +1,148 @@
...
+
+```bash
+npx turbo ffdweb-site#build
+```
 
-## Building
+Once the build is complete, you might want to start the production server by running `npx turbo start --filter=ffdweb-site` or `npx turbo ffdweb-site#start`.
 
-To build the monorepo, run `npm run build`. This command will build all applications and packages in the monorepo. Then, `npm run start` will run the development server using the built files.
+## Remote caching
```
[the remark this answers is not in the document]
CharlyMartin: Yes, I suppose it's the continuity of the previous paragraph

CharlyMartin:
> > It also removes the npm scripts related to turbo commands, in favour of npx turbo
> 
> @CharlyMartin - I don't know the exact advantage of this, but it would be nice to add the reason to the PR description? Just an example:
> 
> We use `npx turbo` instead of relying on npm scripts or globally installed Turbo. This ensures the correct version defined in the project is used and avoids script indirection.

I have a slight preference for `npx turbo dev` over `npm run dev` - even though they do the same thing - because it makes it more obvious that we're using `turbo` and not `next dev` or other dev commands.

I like your note, I will add it!

mirhamasala (approved):
@CharlyMartin - This is good to go for now! Thank you for working on this. ❤️

PR #1494, opened 2025-06-24
[UXIT][UXIT-2652] Update UXIT README (skip percy)

CharlyMartin:
## 📝 Description

This PR modifies the README.md with updated instructions.


mirhamasala (dismissed):
Looks good to me - I do think we should add set up options directly here but please GPT best practices for projects in monorepos.

PR #1504, opened 2025-06-27
[UXIT-2908] Add robots.txt and sitemap

CharlyMartin:
## 📝 Description

This PR adds `robots.ts` and `sitemap.ts` functions to automatically generate` /robots.txt` and `/sitemap.xml` files.


apps/filecoin-site/src/app/robots.ts
```diff
@@ -0,0 +1,13 @@
+import type { MetadataRoute } from 'next'
```
mirhamasala, line 1: @CharlyMartin ~ I'd like to prevent doing things different from the other apps. It will make things more difficult to understand and maintain in the future. So, if you feel like this is the better way to do things, then we need to ensure to adapt the rest as well, and perhaps even abstract where needed. 🙏🏼

mirhamasala (commented):
@CharlyMartin LGTM but I do have one [important comment](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1504/files#r2171882775) that we should address/talk about. 🙏🏼

apps/filecoin-site/src/app/robots.ts
CharlyMartin: Makes sense. This way is a bit more programmatic because we can reuse the domain name. I'll update accordingly!

apps/filecoin-site/src/app/_utils/createPathConfig.ts
```diff
@@ -0,0 +1,11 @@
+import { type StaticPath } from '@/constants/paths'
+
+export function createPathConfig<Path extends StaticPath, Label extends string>(
+  path: Path,
+  label: Label,
+) {
```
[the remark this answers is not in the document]
CharlyMartin: Or am I misunderstanding it?

mirhamasala (approved):
LOVE IT! 😻

PR #1508, opened 2025-06-27
[FILECOIN][UXIT-2923] Add site layout and font

CharlyMartin:
## 📝 Description

This PR implements the `SiteLayout` component and imports the [custom fonts](https://github.com/filecoin-project/filecoin.io/blob/tina/staging/src/sass/globals/fonts.scss) from the existing filecoin.io project: `suisse_intl` and `suisse_mono`

### DIFF
--- a/apps/filecoin-site/src/app/_styles/globals.css
+++ b/apps/filecoin-site/src/app/_styles/globals.css
@@ -1,4 +1,5 @@
 @import 'tailwindcss';
+@source "../../../../../node_modules/@filecoin-foundation/ui";
 
 @theme {
   --color-brand-50: oklch(0.98 0.0157 216.67);
--- a/apps/filecoin-site/package.json
+++ b/apps/filecoin-site/package.json
@@ -18,6 +18,7 @@
     "@filecoin-foundation/eslint-config": "0.0.0",
     "@filecoin-foundation/typescript-config": "0.0.0",
     "@filecoin-foundation/utils": "0.0.0",
+    "@filecoin-foundation/ui": "0.0.0",
     "@tailwindcss/postcss": "^4.1.11",
     "@types/node": "^24.0.3",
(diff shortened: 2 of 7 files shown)


packages/ui/src/SiteLayout.tsx
```diff
@@ -16,7 +16,7 @@ export function SiteLayout({
   children,
 }: SiteLayoutProps) {
   return (
-    <html lang="en" className={font.className}>
+    <html lang="en" className={font}>
```
mirhamasala, line 19: Can you explain this change? Nice to add the motivation for these to the PR description. 🙏🏼

CharlyMartin: This was to be able to pass a class for a custom font, but we don't need one anymore. I'll revert this.

apps/filecoin-site/src/app/_styles/globals.css
```diff
@@ -12,4 +15,8 @@
   --color-brand-800: oklch(0.53 0.1592 252.62);
   --color-brand-900: oklch(0.47 0.1224 248.85);
   --color-brand-950: oklch(0.33 0.0795 249.37);
+
+  --font-sans: 'suisse_intl', Arial, Helvetica, sans-serif;
+  --font-mono:
+    'suisse_mono', Courier, Lucida Sans Typewriter, Lucida Typewriter, monospace;
```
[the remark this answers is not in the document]
CharlyMartin: Figma mentions this [custom font](https://www.figma.com/design/2O2t3VNZPyb41eWznf8qtO/Filecoin.io-v3?node-id=164-1013&t=8FLcJNYaaU1dBilK-4), but I will use `Inter` as discussed on Slack.

mirhamasala (approved):
Yay! 🙌🏼

PR #1523, opened 2025-07-02
[FILECOIN][UXIT-2952] Implement ShareArticle Component

barbaraperic:
## 📝 Description

This PR:

* Reuses `ShareArticle` component in `filecoin-site`
* Makes `sectionTitle` prop in `ShareArticle` component to be explicitly defined, otherwise it is not shown

## 📌 To-Do Before Merging

- [ ] Remove test example in `/blog`
- [ ] Add states once the design is ready

## 📸 Screenshots
Screenshot 2025-07-09 at 11 46 48


apps/ffdweb-site/src/app/_styles/globals.css
```diff
@@ -209,7 +209,7 @@
 
   /* COPY TO CLIPBOARD */
   .copy-to-clipboard-button {
-    @apply hover:text-brand-primary-400;
+    @apply hover:text-brand-primary-400 p-2;
```
barbaraperic, line 212: Had to remove the `TOUCH_TARGET.touchAreaPadding` from the `CopyToClipboard` component as it is different for filecoin-site. Alternatively we can pass a prop to modify the TouchTarget padding?

apps/filecoin-site/src/app/blog/[slug]/page.tsx
```diff
@@ -18,7 +19,13 @@ export default async function BlogPost({ params }: BlogPostProps) {
...
     <>
       <Title backgroundVariant="light">{title}</Title>
-
+      <div className="my-10 flex justify-start">
+        <SharePost
+          articleTitle={title}
+          path={`/blog/${slug}`}
+          baseUrl="https://filecoin.io"
+        />
+      </div>
```
Comment (barbaraperic) [inline apps/filecoin-site/src/app/blog/[slug]/page.tsx]: Test example - remove before merge

apps/filecoin-site/src/app/_styles/globals.css
```diff
@@ -110,4 +110,22 @@
...
+  }
+
+  .social-link {
+    @apply rounded-full border-2 border-zinc-200 p-1 text-zinc-950;
+  }
+
+  /* COPY TO CLIPBOARD */
+  .copy-to-clipboard-button {
+    @apply rounded-full border-2 border-zinc-200 p-1 text-zinc-950;
+  }
```
barbaraperic, line 130: Thought of merging it into 1 class, but we have to keep `copy-to-clipboard-button `naming, which doesn't make sense to put on social link component...

mirhamasala: I don't understand ..

apps/ffdweb-site/src/app/_styles/globals.css
mirhamasala: Can't we keep it the same? It seems like a very small change?

apps/ff-site/src/app/_styles/globals.css
```diff
@@ -188,7 +188,7 @@
 
   /* COPY TO CLIPBOARD */
   .copy-to-clipboard-button {
-    @apply hover:text-brand-400;
+    @apply hover:text-brand-400 p-2;
```
mirhamasala, line 191: Why are we changing FF?

apps/filecoin-site/src/app/_styles/globals.css
mirhamasala: What about something like

```css
@layer components {
  .icon-button {
    @apply rounded-full border-2 border-zinc-200 p-1 text-zinc-950;
  }
}

.social-link {
  @apply icon-button;
  /* Additional social-link specific styles if needed */
}

.copy-to-clipboard-button {
  @apply icon-button;
  /* Additional copy button specific styles if needed */
}
```

apps/filecoin-site/src/app/_components/SharePost.tsx
```diff
@@ -0,0 +1,67 @@
...
+        <li
+          key={label}
+          className={clsx('inline-flex', TOUCH_TARGET.touchAreaPadding)}
+        >
+          <Tooltip description={`Share on ${label}`} side="bottom">
+            <a
+              aria-label={`Share ${articleTitle} on ${label}`}
+              href={href}
+              rel="noopener noreferrer"
+              className="focus:brand-outline social-link"
```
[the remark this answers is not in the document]
mirhamasala: Or are we only moving changeable styles? I don't remember. :-)

mirhamasala (changes requested):
@barbaraperic Left some comments! 🙏🏼

packages/utils/src/generateShareArticleLinks.ts
```diff
@@ -7,44 +7,50 @@ import {
```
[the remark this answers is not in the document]
barbaraperic: AH yes, I remember we discussed, I will look into it 🙏

apps/filecoin-site/src/app/_components/SharePost.tsx
```diff
@@ -0,0 +1,67 @@
...
+  articleTitle: string
+  path: string
+  baseUrl: string
+  sectionTitle?: string
+}
+
+const TOUCH_TARGET: TouchTarget = {
+  touchAreaPadding: 'p-2',
+  touchAreaOffset: '-m-2',
+}
```
CharlyMartin, line 21: The `TOUCH_TARGET` is not doing its job here. It's added to the `li`, not the `button`, so it does not augment the clickable / touchable area.

BUT, if we add it to the button, the border would be further away than what is on Figma. We would need to rework this component to make this work.

Or talk to Filipa to make these buttons transparent, like on FF / FFDW.

PR #1553, opened 2025-07-14
[FILECOIN][UXIT-2978] Update Button Styles

CharlyMartin:
## 📝 Description

This PR updates the button style to be closer to Figma.

## 🛠️ Key Changes

- Moves `focus:brand-outline` outside shared button to add custom focus styles
- Remove one gradient class to keep two variables: `gradient-brand--dark` and `gradient-brand--light`
- Remove `.button--disabled` class in favor of disabled pseudo class
- Remove dark button variants in favor of automatic detection of `.dark-section` parent

## 📌 To-do before Merging

- [ ] Remove button storybook


apps/filecoin-site/src/app/(homepage)/page.tsx
```diff
@@ -108,11 +110,11 @@ export default function Home() {
             <Button
               key="button-1"
```
mirhamasala, line 111: Why do we have these keys?

CharlyMartin: The `cta` prop is an array of buttons, so they each need unique keys

apps/filecoin-site/src/app/_components/Footer/NewsletterForm.tsx
```diff
@@ -46,7 +46,7 @@ export function NewsletterForm() {
                 className="absolute inset-0"
                 style={{
                   content: '',
-                  background: 'var(--gradient-brand)',
+                  background: 'var(--gradient-brand--dark)',
```
mirhamasala, line 49: Should it be:

```suggestion
                  background: 'var(--gradient-brand-dark)',
```

mirhamasala (changes requested):
@CharlyMartin ~ Love how this now works with `dark-section`, etc. So much cleaner.

Can you please have a look at the focus styles? They seem odd to me.. I would expect all focus styles to have the same increase in border; otherwise, it looks quite weird when you tab through. But also, the colors for some don't seem right, and there seems to be a flicker on the primary light version.

Image

PR #1571, opened 2025-07-18
[FILECOIN][UXIT-3037] Create Tertiary Button Variant [skip percy]

mirhamasala:
## 📝 Description

This PR includes the tertiary button styles.


apps/filecoin-site/src/app/_components/Button.tsx
```diff
@@ -6,6 +6,7 @@ import { BASE_DOMAIN } from '@/constants/siteMetadata'
 const variantClasses = {
   primary: 'button--primary',
   ghost: 'button--ghost',
```
mirhamasala, line 8: We may or may not want to change this to secondary if it fits at some point..

CharlyMartin (approved):
Thank you for doing this! 😄

PR #1606, opened 2025-07-30
[UXIT-3064] Add Listbox Category Filter

CharlyMartin:
## 📝 Description

This PR adds `StorageCategoryListbox` to filter storage providers on mobile. It uses `@headlessui/react`, instead of our existing `Listbox`, as the implementation differs.

> [!NOTE]
> Currently, the checkbox and listbox have similar designs but duplicate code. This is because they rely on different HTML markup and pseudo-classes to style the active state. This might be refactored in the future.

## 🛠️ Key Changes

- Adds `isTruthy` util function for type-safe filtering of `undefined`.

## 📸 Screenshots

CleanShot 2025-07-30 at 17 27 51

### DIFF
--- a/apps/filecoin-site/src/app/store-data/components/StorageCategoryListbox.tsx
+++ b/apps/filecoin-site/src/app/store-data/components/StorageCategoryListbox.tsx
@@ -7,13 +7,13 @@ import {
 import { CaretDownIcon } from '@phosphor-icons/react'
 
 import { Icon } from '@filecoin-foundation/ui/Icon'
-import { isTruthy } from '@filecoin-foundation/utils/isTruthy'
 
 import {
   storageCategories,
   type StorageCategory,
 } from '../data/storageCategories'
 import { storageProviders } from '../data/storageProviders'
+import { getActiveCategories } from '../utils/getActiveCategories'
 import { getProvidersForCategory } from '../utils/getProvidersForCategory'
 
 const DEFAULT_BUTTON_TEXT = 'Categories'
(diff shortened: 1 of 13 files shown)


apps/filecoin-site/src/app/store-data/components/StorageCategoryListbox.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+const DEFAULT_BUTTON_TEXT = 'Categories'
+
+export function StorageCategoryListbox({
+  activeFilters,
+  setFilters,
+}: StorageCategoryFilterProps) {
+  const activeCategories = activeFilters
+    .map((id) => storageCategories.find((category) => category.id === id))
+    .filter(isTruthy)
```
mirhamasala, line 26: Seems like a lot. No need to change, but another solution from Claude that might read easier:

```
const activeCategoryIds = new Set(activeFilters);
const activeCategories = storageCategories.filter(category => 
  activeCategoryIds.has(category.id)
);
```

apps/filecoin-site/src/app/store-data/components/StorageCategoryListbox.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  return (
+    <Listbox
+      multiple
+      value={activeCategories}
+      onChange={(newSelection) => {
```
mirhamasala, line 32: Would probably abstract it into `handleCategoryChange` 

```ts
function handleCategoryChange(newSelection, setFilters) {
  const newSelectionIds = newSelection.map(category => category?.id);
  setFilters(newSelectionIds);
}
```

apps/filecoin-site/src/app/store-data/components/StorageCategoryListbox.tsx
```diff
@@ -0,0 +1,97 @@
...
+          <Icon component={CaretDownIcon} size={16} weight="bold" />
+        </div>
+      </ListboxButton>
+
+      <ListboxOptions
+        className="focus:brand-outline mt-2 w-[var(--button-width)] space-y-2 border border-zinc-200 bg-white p-2 shadow-lg"
+        anchor="bottom start"
+      >
+        {storageCategories.map((category) => {
+          const { id, name, description } = category
```
mirhamasala, line 52:
```suggestion
        {storageCategories.map(({id, name, description) => {
```

PR #1683, opened 2025-08-20
[FILECOIN][UXIT-3155] Adapt Cards border styles [skip percy]

barbaraperic:
## 📝 Description

- I've found `SimpleCard` component a bit hard to update and understand styles, so I refactored the component to use styles depending on its borders (all | none | only-top). Let me know thoughts!
- adapted border radius to match the ones in Figma: `SimpleCardWithLogo` (`rounded-sm`) and `SimpleCard` (`rounded-2xl`)

- **Type:**  Refactor

## 📸 Screenshots

Screenshot 2025-08-20 at 13 02 17

Screenshot 2025-08-20 at 13 12 53


CharlyMartin (approved):
LGTM @barbaraperic!

Just one piece of feedback: the focus style doesn't hug the border of the card. The focus style is square while the card is `rounded-2xl`.

Image

I couldn't find a quick way to make it work, but maybe you will. Other than that, good to go!

PR #1711, opened 2025-08-25
[FILECOIN][UXIT-3148] Assets import order [skip percy]

barbaraperic:
## 📝 Description

Updated ESLint configuration to include asset imports under internal grouping (assets before components).

- **Type:** Refactor


barbaraperic:
> Oh, great! Could we do `@/content/** → @/data/** → @/assets/** → @/utils/**`
> 
> _That ordering treats `@/assets/` as more foundational than `@/utils/` because assets are pure static resources with zero dependencies, while utils often contain logic that might reference or process those assets. It follows a dependency hierarchy where the most dependency-free modules (assets) come before modules that might depend on them (utilities that could manipulate images, reference icons, etc.)._

Done 🙂

PR #1717, opened 2025-08-26
[FILECOIN][UXIT-3048] Set up Cypress metadata tests [skip percy]

barbaraperic:
## 📝 Description

- Added Cypress configuration file for end-to-end testing.
- Added Cypress tasks for fetching entry frontmatter and random slugs.
- Created test files for all 12 current pages ensuring metadata and link validation.
- Updated package.json to include Cypress scripts for running tests.
- Enhanced TypeScript configuration to support Cypress paths.

➡️ NOTE 1: 
There's a known hydration error ([open issue here](https://github.com/cypress-io/cypress/issues/27204)) when using Cypress 15 with React 18, Next, and server components. 

This happens because Cypress injects its own `<script>` into the head before hydration, which causes mismatch against our inline `<script>` tags (`StructuredDataScript`). The errors don’t appear in a normal browser session, only when Cypress is running. 

I've added `<Suspense>` around `StructuredDataScript` as a suggested workaround to prevent this hydration mismatches.

➡️ NOTE 2: 
Once we configure Percy, we can update the github action workflow (since it includes running both Cypress and Percy).

## 🧪 How to Test

- run `npm run build && npm run start`
- run npm run test:cypress
- check if all test pass ✅

## 📸 Screenshots

Screenshot 2025-08-26 at 09 22 47


packages/cypress/src/metadata/testPageMetadata.ts
```diff
@@ -5,18 +5,22 @@ export type TestMetaDataOptions = {
   title: string
   description: string
   baseUrl: string
+  excludeMetadata?: {
+    og?: ['image']
+    twitter?: ['image']
+  }
```
barbaraperic, line 11: A workaround until we add images to `createMetadata`

mirhamasala: Could we use the fallback image for now instead of having a workaround?

mirhamasala:
> Once we configure Percy, we can update the github action workflow (since it includes running both Cypress and Percy).

@barbaraperic update how?

apps/filecoin-site/src/app/_components/SiteLayout.tsx
```diff
@@ -43,7 +45,9 @@ export function SiteLayout({ children }: SiteLayoutProps) {
       className={clsx(aspekta.variable, funnelSans.variable, 'antialiased')}
     >
       <head>
-        <StructuredDataScript structuredData={ORGANIZATION_SCHEMA_BASE} />
+        <Suspense>
```
mirhamasala, line 48: I'm unsure about the implications of wrapping everything in `Suspense` - @CharlyMartin ?

mirhamasala (commented):
@barbaraperic - I think this mostly looks good to me but I would like @CharlyMartin to have a second look. I'm unsure about implications of wrapping the entire `SiteLayout` in `Suspense`.

Also, @barbaraperic @CharlyMartin - Is there are reason why we have the what seem to be the exact same helper functions across different apps? `getRandomSlug`, `getEntryFrontmatter`, `getMetaTitleWithSuffix`, `getMateTitleTemplate`? Or we simply haven't had the chance yet to refactor these?

Lastly, I would suggest we use the fallback image instead of changing the logic to include `excludeMetadata`.

apps/filecoin-site/src/app/_components/SiteLayout.tsx
CharlyMartin: The only implication I can think of is on SEO. We would need to test some pages in the [rich results tool](https://search.google.com/test/rich-results) to confirm the data is still detected.

Having said that, it's a bit odd that we're using `Suspense` here, as `StructuredDataScript` does not perform async operations / is not an dynamic boundary. But I understand the limitation, so maybe we add a comment to explain why?

OR, we remove `StructuredDataScript` from the head and place it before the closing body tag. It doesn't have to be in the head, anywhere on the page should be fine.

What do you think?

CharlyMartin (commented):
Looking good @barbaraperic!

As @mirhamasala said, there's an opportunity to centralize / refactor the helper functions as they all seem to do the same. And to answer her question, I think we just haven't had the time to do this yet.

apps/filecoin-site/src/app/_components/SiteLayout.tsx
CharlyMartin: The only impact I can think of is on SEO. We would need to test some pages in the [rich results tool](https://search.google.com/test/rich-results) to confirm the data is still detected.

Having said that, it's a bit odd that we're using `Suspense` here, as `StructuredDataScript` does not perform async operations / is not an dynamic boundary. But I understand the limitation, so maybe we add a comment to explain why?

OR, we remove `StructuredDataScript` from the head and place it before the closing body tag. It doesn't have to be in the head, anywhere on the page should be fine.

What do you think?

mirhamasala: @barbaraperic What happens if we just remove Suspense? Do the cypress tests not run at all? Or just run with an error?

barbaraperic: @mirhamasala  They run, but fail.

barbaraperic: Screenshot 2025-08-27 at 11 36 20

barbaraperic: Another approach would be to add exception handing

```
// apps/filecoin-site/cypress/support/e2e.ts
Cypress.on('uncaught:exception', (err) => {
  // Handle React 19 hydration errors
  if (
    err.message.includes('Hydration failed because the initial UI does not match') ||
    err.message.includes('Minified React error #418') ||
    err.message.includes('Minified React error #423') ||
    err.message.includes('Unknown root exit status')
  ) {
    return false;
  }
  return true;
});
```

mirhamasala: I think I like that more for now - We could add a comment about revisiting it with the URL to the known issue.

PR #1724, opened 2025-08-26
[FILECOIN][UXIT-3185] Update navigation interactive styles

CharlyMartin:
## 📝 Description

This PR updates the light theme design of `DesktopNavigation` to match Figma, along with a few improvements:
- An `isCurrent` prop was added to the shared `ui/NavigationMenu` component to apply a `data-state='current'` attribute, to make styling easier.
- A new `useIsNavigationMenuActive` hook was created to check if the current path matches any of the menu's child links.
- The desktop navigation was refactored, extracting the menu logic into a new `NavigationMenu` component that uses this new hook.
- The `BackgroundVariantContextProvider` provider was created to make it easier to read the `backgroundVariant` of a given section without passing props

## 📸 Screenshots

CleanShot 2025-08-27 at 10 46 02@2x


apps/filecoin-site/src/app/_components/BackgroundVariantContext.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const backgroundVariant = use(BackgroundVariantContext)
+
+  if (!backgroundVariant) {
+    throw new Error(
+      'useBackgroundVariant must be used within a BackgroundVariantContext',
+    )
+  }
+
+  return backgroundVariant
+}
```
mirhamasala, line 19: So we need this? Considering we're providing the `transparent` default?

mirhamasala (approved):
@CharlyMartin this is great! And so much clearer than what we had. ✨

One quick thing: could you please add a note in PR descriptions when they’re based on each other? I didn’t realize that [#1728](https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1728) was based on this one, and I ended up creating a mess. Thanks!

PR #1725, opened 2025-08-27
[FILECOIN][UXIT-3195] Refactor Cypress task and util functions [skip percy]

barbaraperic:
## 📝 Description

- Moved `tasks`, `utils` and `support` folders from local cypress folder to `@filecoin-foundation/cypress` package
- Update import paths across all spec files to use shared package
- Removed local cypress folders in favor of centralized utilities

- **Type:**  Refactor

### ff-site
Screenshot 2025-08-27 at 12 45 20

### ffdw-site

Screenshot 2025-08-27 at 12 52 45

### filecoin-site

Screenshot 2025-08-27 at 12 40 54


CharlyMartin (approved):
Solid refactor Barbara! It's looking great 🙌

One thing I would change is how we use the shared cypress utils, like `getMetaTitleWithSuffix`. Instead of writing the following everytime:

```ts
// apps/ff-site/cypress/e2e/digest_page_spec.cy.ts

        const metaTitleTemplate = getMetaTitleTemplate({
          title: seo.title,
          rootMetadata: ROOT_METADATA,
        })
```

I would keep a local vercion in each apps to centralize this logic, like this:

```
// apps/ff-site/cypress/utils/getMetaTitleWithSuffix.ts

import { getMetaTitleWithSuffix as sharedGetMetaTitleWithSuffix } from '@filecoin-foundation/cypress/utils'

import { ROOT_METADATA } from '@/constants/siteMetadata'

export function getMetaTitleWithSuffix(title: string) {
  return sharedGetMetaTitleWithSuffix({
    title,
    organizationName: ROOT_METADATA.title.default,
  })
}
```

So we can keep doing this:

```ts
// apps/ff-site/cypress/e2e/digest_page_spec.cy.ts

        tests.metadata.fn({
          path,
          title: getMetaTitleTemplate(seo.title),
          description: seo.description,
          baseUrl: BASE_URL,
        })
```

And same logic for `getMetaTitleTemplate`

What do you think?

PR #1729, opened 2025-08-27
[UXIT-3182] Fix Orbit links

CharlyMartin:
## 📝 Description

This commit introduces a `MarkdownLink` wrapper component in `ff-site` to be used where `prose` styles apply, like here on the Orbit page.


CharlyMartin:
> Can we explain better in the PR description when `MarkdownLink` has to be used?
> 
> For example, the `prose` class is also used in `MarkdownContent`, `MarkdownPage, and `CoreFunctions`. How come link styles work there out of the box (do they?) But not in this case? (Without that, it's difficult for me to understand whether this is a necessary evil.)

The `prose` class is mainly used in `MarkdownContent`, which automatically uses `MarkdownLink` internally for links, so nothing to worry about here.

`MarkdownPage` renders `MarkdownContent`, so it's the same thing.

`CoreFunctions` also uses `prose` with links, but these links are not "content". They are used for navigation purposes and have a special style. We couldn't use a SmartTextLink here, which is what `MarkdownLink` replaces in `prose` context.

So `MarkdownLink` has to be used for typical text links in `prose` context. This happens automatically in 99% of the time. Then, once in a while, an edge case happens, like here on the Orbit and we have to get a bit creative.

PR #1770, opened 2025-09-17
[FF][UXIT-2730] Allocators Page · Update schema 

CharlyMartin:
## 📝 Description

This PR updates the `AllocatorSchema` Zod schema according to the new JSON format in the [Allocators repository](https://github.com/filecoin-project/Allocator-Registry/tree/main/Allocators).

Data inconsistencies: https://filecoinfoundation.slack.com/archives/C07R5EGAAMT/p1758116017361409?thread_ts=1747733960.701129&cid=C07R5EGAAMT

## 🛠️ Key Changes

- The zod schemas are split into multiple files for clarity
- The `location` property is no longer available in the `AllocatorSchema`, so the `location` column and its corresponding filter have been removed from the allocators table
- The `type` filter has been updated to use the static options from the zod enum.

## 📸 Screenshots

CleanShot 2025-09-17 at 14 05 00

## 🔖 Resources

- https://zod.dev/api


apps/ff-site/src/app/filecoin-plus/allocators/schemas/AllocatorFileSchema.ts
mirhamasala, line 1: Currently the schema mixes conventions for “no value” — some fields allow '' `(.or(z.literal('')))`, others use `.optional() (undefined)`. This inconsistency forces consumers to handle missing values differently across fields (`val === ''` vs `val === undefined`).

To avoid this friction, Claude recommends standardizing. A good option is to normalize empty strings to null at parse time so consumers always work with string | null:


```ts
const OptionalIsoDateSchema = z
  .string()
  .transform((val) => (val.trim() === '' ? null : val))
  .pipe(z.string().datetime().nullable());
```

This keeps compatibility with upstream data that sends `''`, but presents a consistent `string | null` type downstream.

apps/ff-site/src/app/filecoin-plus/allocators/schemas/AllocatorsSchema.ts
```diff
@@ -0,0 +1,87 @@
...
+const FilecoinAddressSchema = z.string().regex(/^f[a-zA-Z0-9]+$/, {
+  message: 'Invalid Filecoin address',
+})
+
+const AuditSchema = z.object({
+  started: z.iso.datetime(),
+  ended: OptionalIsoDateSchema,
+  dc_allocated: OptionalIsoDateSchema,
+  outcome: AuditOutcomeSchema.or(z.literal('')),
+  datacap_amount: z.coerce.number(),
```
mirhamasala, line 36: `z.coerce.number()` can silently convert unexpected values (e.g., "abc" → NaN). This allows invalid input to slip past validation and can cause runtime issues downstream.

Prefer explicit validation to ensure only valid numeric input is accepted:

// Accept numeric strings but validate
datacap_amount: `z.string().pipe(z.coerce.number().positive())`,

// OR require strict numbers only
`datacap_amount: z.number().nonnegative()`,

Choose `.pipe()` if upstream data may contain strings, or `z.number()` if you can guarantee clean numeric inputs.

apps/ff-site/src/app/filecoin-plus/allocators/schemas/AllocatorFileSchema.ts
```diff
@@ -0,0 +1,31 @@
...
+      html: z.url(),
+    })
+    .optional(),
+})
+
+export const AllocatorFileMetaDataSchema =
+  AllocatorFileMetaDataBaseSchema.extend({
+    content: z.string(),
+    encoding: z.enum(['base64']),
+  })
```
mirhamasala, line 26: Consider making the base/externded schema relatioship more explicit:


```suggestion
export const AllocatorFileMetaDataSchema =
  AllocatorFileMetaDataBaseSchema.merge(
    z.object({
      content: z.string(),
      encoding: z.enum(['base64']),
    }),
  )
```

apps/ff-site/src/app/filecoin-plus/allocators/schemas/AllocatorFileSchema.ts
```diff
@@ -0,0 +1,31 @@
+import { z } from 'zod'
+
+export const AllocatorFileMetaDataBaseSchema = z.object({
+  name: z.string(),
+  path: z.string(),
+  sha: z.string(),
+  size: z.number(),
```
mirhamasala, line 7: Right now, any number will pass

apps/ff-site/src/app/filecoin-plus/allocators/schemas/AllocatorsSchema.ts
CharlyMartin: I like the first approach `z.string().pipe(z.coerce.number().positive())`

In the source data, `datacap_amount` is either a number or a string representing a number, so I imagine we would need something like that:

```ts
z.number().positive().or(.string().pipe(z.coerce.number().positive()))
```

apps/ff-site/src/app/filecoin-plus/allocators/schemas/AllocatorsSchema.ts
```diff
@@ -0,0 +1,87 @@
...
+  'KYC Submitted': z.iso.datetime(),
+  Approved: z.iso.datetime(),
+  Declined: OptionalIsoDateSchema,
+  'DC Allocated': OptionalIsoDateSchema,
+})
+
+// Some "audit" values contain leading/trailing spaces in the Allocator repo: https://github.com/filecoin-project/Allocator-Registry/tree/main/Allocators
+// Once these values are cleaned, we can simplify "audit" to be `z.array(ApplicationAuditSchema)`
+const ApplicationSchema = z.object({
+  audit: z.array(
```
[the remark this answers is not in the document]
CharlyMartin: I like it! It cleans the output as well, unlike my initial solution.

apps/ff-site/src/app/filecoin-plus/allocators/schemas/AllocatorsSchema.ts
CharlyMartin: Actually, `z.coerce.number().positive()` seems to be enough! Zod handles `NaN`, meaning throws an error if the input is not coercible to a number.

CleanShot 2025-09-18 at 14 37 56

apps/ff-site/src/app/filecoin-plus/allocators/schemas/AllocatorsSchema.ts
```diff
@@ -0,0 +1,87 @@
...
+  'Automated',
+  'Automated Allocator',
+  'Other',
+])
+
+export const MetapathwayTypeSchema = z.enum(['MDMA', 'RKH', 'AMA', 'ORMA'])
+
+const OptionalIsoDateSchema = z.iso.datetime().or(z.literal(''))
+
+const FilecoinAddressSchema = z.string().regex(/^f[a-zA-Z0-9]+$/, {
```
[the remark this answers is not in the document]
CharlyMartin: I tighten the address regexp.

Though, I asked Claude to create a bullet-proof regexp based on the [docs](https://docs.filecoin.io/smart-contracts/filecoin-evm-runtime/address-types), and the validation failed. I don't know if the data is invalid or if Claude hallucinated.

apps/ff-site/src/app/filecoin-plus/allocators/schemas/AllocatorsSchema.ts
```diff
@@ -0,0 +1,87 @@
...
+const ApplicationSchema = z.object({
+  audit: z.array(
+    z
+      .string()
+      .trim()
+      .refine((value) => ApplicationAuditSchema.safeParse(value).success, {
+        message: 'Invalid application audit type',
+      }),
+  ),
+  distribution: z.array(z.string()),
```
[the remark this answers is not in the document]
CharlyMartin: "distribution" is, as of now, always an empty array.

Not sure what values it will hold.

One of the things I will need to clarify with the Fil+ team.

PR #1786, opened 2025-09-23
[FF][UXIT-3271] Configure Airtable client timeout and retry logic

CharlyMartin:
## 📝 Description

This PR updates the Airtable API client as follows:
1. `noRetryIfRateLimited` is set to `true` to disable automatic retries. This ensures the application fails fast when rate-limited, providing more predictable execution times.
2. A `requestTimeout` of 3 seconds is introduced to prevent requests from hanging and potentially causing serverless function timeouts.

This solves the current [time-out issue](https://filecoinfoundation.slack.com/archives/C06MTEQ3SP6/p1758626653393969) on the orbit page by erroring out instead of infinitely retrying. I suspect the issue to be the 1k API limit on the Airtable account, [to be confirmed](https://filecoinfoundation.slack.com/archives/C08MDPD487R/p1758705408468849) with the Fil-B team.

## 🛠️ Key Changes

It also moves the zod schema in its own file.

## 📸 Screenshots

CleanShot 2025-09-24 at 10 28 51@2x

## 🔖 Resources

https://airtable.com/pricing

### DIFF
--- a/apps/ff-site/src/app/orbit/schema/AirtableRecordSchema.ts
+++ b/apps/ff-site/src/app/orbit/schema/AirtableRecordSchema.ts
@@ -1,0 +1,13 @@
+import { z } from 'zod'
+
+import { AIRTABLE_ORBIT_EVENTS_CONFIG } from '../constants/airtableOrbitEventsConfig'
+
+const { TITLE, CITY, START_DATE, REGISTRATION_LINK } =
+  AIRTABLE_ORBIT_EVENTS_CONFIG.FIELDS
+
+export const AirtableRecordSchema = z.object({
+  [TITLE]: z.string(),
+  [CITY]: z.string(),
+  [START_DATE]: z.coerce.date(),
+  [REGISTRATION_LINK]: z.url().optional(),
+})
--- a/apps/ff-site/src/app/orbit/services/airtable.ts
+++ b/apps/ff-site/src/app/orbit/services/airtable.ts
@@ -7,6 +7,7 @@ import {
   AIRTABLE_ORBIT_EVENTS_CONFIG,
   APPROVED_STATUS_VALUE,
 } from '../constants/airtableOrbitEventsConfig'
+import { AirtableRecordSchema } from '../schema/AirtableRecordSchema'
 
 const airtable = new Airtable({
   apiKey: process.env.AIRTABLE_READ_ONLY_TOKEN,
@@ -16,13 +17,6 @@ const airtable = new Airtable({
 
 const { BASE_ID, EVENTS_TABLE_ID, FIELDS } = AIRTABLE_ORBIT_EVENTS_CONFIG
 const { TITLE, CITY, START_DATE, REGISTRATION_LINK } = FIELDS
-
-const airtableRecordSchema = z.object({
-  [TITLE]: z.string(),
-  [CITY]: z.string(),
-  [START_DATE]: z.coerce.date(),
-  [REGISTRATION_LINK]: z.url().optional(),
-})
 
 export async function fetchAndParseAirtableEvents() {
   const rawAirtableRecords = await fetchAirtableRecords()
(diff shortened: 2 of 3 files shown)


CharlyMartin:
> @CharlyMartin What happens if, despite these configurations, we still hit a rate limit? Is it logged anywhere (f.e. Sentry)? How is it handled on the frontend?

@mirhamasala We are currently hitting a rate limit, and I suspect the free plan is the cause of it (TBC). If that's the case, these configurations will not solve the rate problem, but they will prevent the page from retrying forever and crashing.

We do capture this error with `Sentry.captureException(error)`, and it's handled like any other Airtable error in the frontend by displaying a link to the original calendar in Airtable.

apps/ff-site/src/app/orbit/schema/AirtableRecordSchema.ts
```diff
@@ -0,0 +1,13 @@
...
+import { AIRTABLE_ORBIT_EVENTS_CONFIG } from '../constants/airtableOrbitEventsConfig'
+
+const { TITLE, CITY, START_DATE, REGISTRATION_LINK } =
+  AIRTABLE_ORBIT_EVENTS_CONFIG.FIELDS
+
+export const AirtableRecordSchema = z.object({
+  [TITLE]: z.string(),
+  [CITY]: z.string(),
+  [START_DATE]: z.coerce.date(),
+  [REGISTRATION_LINK]: z.url().optional(),
```
mirhamasala, line 12: Do we want Airtable to throw if Airtable sends an empty string?

apps/ff-site/src/app/orbit/services/airtable.ts
```diff
@@ -7,21 +7,20 @@ import {
   AIRTABLE_ORBIT_EVENTS_CONFIG,
   APPROVED_STATUS_VALUE,
 } from '../constants/airtableOrbitEventsConfig'
+import { AirtableRecordSchema } from '../schema/AirtableRecordSchema'
 
-const airtable = new Airtable({ apiKey: process.env.AIRTABLE_READ_ONLY_TOKEN })
+const airtable = new Airtable({
+  apiKey: process.env.AIRTABLE_READ_ONLY_TOKEN,
+  requestTimeout: 3_000,
```
mirhamasala, line 14: Nit but best practices dictate that timeouts, limits, URLs, API keys should all live in config/env and not source code.

`requestTimeout: Number(process.env.AIRTABLE_REQUEST_TIMEOUT ?? 3000)`

Which would allow ops/devs to ovverride via `AIRTABLE_REQUEST_TIMEOUT=5000` (or shorter for CI tests)

CharlyMartin:
> > We do capture this error with Sentry.captureException(error), and it's handled like any other Airtable error in the frontend by displaying a link to the original calendar in Airtable.
> 
> @CharlyMartin Got it.
> 
> Should we track them separately?
> 
> ```ts
> if (error.statusCode === 429) {
>   Sentry.captureException(error, { tags: { source: 'airtable', type: 'rate_limit' } });
> }
> ```

@mirhamasala I don't think it's necessary, `captureException` receives the following error object, so it should be enough?

```ts
AirtableError {
  error: 'TOO_MANY_REQUESTS',
  message: 'You have made too many requests in a short period of time. Please retry your request later',
  statusCode: 429,
}
```

apps/ff-site/src/app/orbit/schema/AirtableRecordSchema.ts
CharlyMartin: You mean Zod to throw if `REGISTRATION_LINK` is an empty string?

We made REGISTRATION_LINK optional because, in some cases, an event is approved and announced, but the registration link is not available yet.

So we show the partial event info instead of throwing an error.

mirhamasala: Sorry - I mixed up something here. All good.

CharlyMartin: @mirhamasala so good to merge?

PR #1797, opened 2025-09-25
[SHARED][UXIT-3198] Move ignoredClientErrors from edge to client

CharlyMartin:
## 📝 Description

This PR moves `ignoredClientErrors` configuration from edge runtime to client-side instrumentation files across all apps. I initially confused `edge` for `client`, but `edge` refers to middleware and `const runtime = 'edge'` files, which we don't use.

We received this [error](https://filecoin-foundation-qk.sentry.io/issues/6897220810?project=4507390577999872) after the initial `ignoreErrors` was (incorrectly) set up.`TypeError` should have been ignored; hopefully it will be now!

This PR also removes the types from sentry-config as we _may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders_

Finally, it fixes a Zod error introduced by https://github.com/FilecoinFoundationWeb/filecoin-foundation/pull/1796 to fix the build.

### DIFF
--- a/apps/ff-site/instrumentation-client.ts
+++ b/apps/ff-site/instrumentation-client.ts
@@ -1,5 +1,7 @@
 import * as Sentry from '@sentry/nextjs'
 
+import { ignoredClientErrors } from '@filecoin-foundation/sentry-config'
+
 Sentry.init({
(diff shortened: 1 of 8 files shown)


mirhamasala (approved):
❤️

